PR #60, opened 2016-12-01
adding Fiji ROCm 1.3 benchmark data

guacamoleo:
and user can choose which benchmark data to use


tingxingdong:
What is the difference between Fiji mini and Fiji full?

PR #71, opened 2017-01-26
take *.hpp out of library to avoid multiply defintion

tingxingdong:
Summary of proposed changes:
-  a temporary solution: take rocblas.hpp out of library/include until a graceful solution found
-  add batched trtri, trtri, trsm, gemm, trsm into client and tester
-  transfer BUILD_WITH_TENSILE into client cmakefiles
-  add legal headers in some files


clients/include/rocblas.hpp
```diff
@@ -155,26 +151,29 @@
         rocblas_operation transA, rocblas_diagonal diag,
         rocblas_int m, rocblas_int n,
         const T* alpha,
-        const T* A, rocblas_int lda,
-        T*       B, rocblas_int ldb);
+        T* A, rocblas_int lda,
```
amcamd: should this be     const T* A    in place of   T* A

library/src/include/definitions.h
```diff
@@ -1,3 +1,8 @@
+/* ************************************************************************
+ * Copyright 2016 Advanced Micro Devices, Inc.
```
amcamd: should this be    Copyright 2017       not       Copyright 2016     ?

clients/include/rocblas.hpp
tingxingdong: That is a hcc compiler bug. In the trsm code, I have  to use *float *ptr= A+offset",I believe it will pass in gcc. but in hcc, it reports an error. so I have to remove the const qualifier.

kknox: You can probably keep the const qualifier in the API, but then remove the const with a cast before you call into the kernel.

library/src/include/definitions.h
tingxingdong: we use 2016 everywhere now

PR #73, opened 2017-01-27
error found for gemm with dimensions 3, 3, 3 and 5, 5, 5

amcamd:
resolves #___

Summary of proposed changes:
-  add to gemm unit tests problem sizes m,n,k = 3, 3, 3   and    m,n,k = 5, 5, 5
-  These sizes are giving errors
-

### DIFF
--- a/clients/gtest/gemm_gtest.cpp
+++ b/clients/gtest/gemm_gtest.cpp
@@ -44,6 +44,8 @@ Representative sampling is sufficient, endless brute-force sampling is not neces
 const
 vector<vector<int>> matrix_size_range = {
                                         {-1, -1, -1, -1, 1, 1},
+                                        { 3, 33,  3,  3, 3, 3},
+                                        { 5,  5,  5,  5, 5, 5},
                                         {10, 10, 20, 100, 10, 10},
                                         {600,500, 500, 500, 600, 500},
                                         {1024, 1024, 1024, 1024, 1024, 1024}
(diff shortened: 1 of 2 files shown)


clients/gtest/gemm_gtest.cpp
```diff
@@ -44,6 +44,8 @@ Representative sampling is sufficient, endless brute-force sampling is not neces
 const
 vector<vector<int>> matrix_size_range = {
                                         {-1, -1, -1, -1, 1, 1},
+                                        { 3, 33,  3,  3, 3, 3},
```
kknox: Did you mean to put 33 or 3 in this list?

amcamd: 33 is also giving errors, and Tim asked for this to be added as he uses it for batched gemm used for trsm

kknox: I'm trying to validate the ldb of this line to be 3; are these column major?  If this is right, then is the {600,500, 500, 500, 600, 500} line wrong?

amcamd: You are correct, {600,500, 500, 500, 600, 500} corresponds to 
M=600, N=500, K=500, lda=500, ldb=600, ldc=500
GEMM requires lda and ldc >= M and ldb >= k.  This example violates lda >= M.  GEMM will return rocblas_status_invalid_size. The argument checks are in line 60 of 
rocBLAS-develop/library/src/blas3/Tensile/gemm.cpp

3, 33, 3, 3, 3, 3  corresponds to    lda=ldb=ldc=3,   and m=3, n=33, k=3.  This satisfies the argument checks lda >=m, ldc >= m, ldb >= k

tingxingdong: {600, 500, 500, xxx} are is valid for certain transpose but invalid for other transpose but should pass as the argument checker will filter it out.  We purposely make some artificial ones.

tingxingdong: {3, 33,3 ,xxxx} valid for NN, as K is reduction dim but not valid for NT. but should be filtered by argument checker. This size is required by TRSM

kknox: OK, understood; you probably need to get rid of the comments above matrix_size_range.  "Advance users only: BrainStorm the parameters but do not make artificial one which invalidates the matrix."  That is what confused me.

amcamd: Removed Advanced users comment

PR #77, opened 2017-02-09
generate good conditioning matrix

tingxingdong:
resolves #___

Summary of proposed changes:
-  factorize the matrix A with LU factorization first and then serve to trsm
-  add LU (getrf) interface

### DIFF
--- a/library/src/blas3/trtri_device.h
+++ b/library/src/blas3/trtri_device.h
@@ -79,24 +79,24 @@ trtri_device(rocblas_fill uplo,
             }
         }
     }
-    __syncthreads(); // since NB < 64, this synch can be avoided
+    __syncthreads(); // if NB < 64, this synch can be avoided
 
     //invert the diagonal element
     if (tx < n){
         //compute only diagonal element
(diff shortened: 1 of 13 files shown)


kknox:
Seems to still have tiny precision errors.  I can't tell if this is better than before, but I like that you removed the 8000 testing size.

library/src/blas3/trtri_device.h
```diff
@@ -106,7 +106,7 @@ trtri_device(rocblas_fill uplo,
         //use the diagonal one to update current column
         if(tx > col) reg += sA[tx + col * n] * sA[col + col * n];
 
-        __syncthreads(); // since NB < 64, this synch can be avoided
+        __syncthreads(); // if NB < 64, this synch can be avoided
```
amcamd: should line 120 and line 126 also say "if NB < 64" in place of "since NB < 64" ?

amcamd: should line 117 be 0.0  in place of 0  ?

PR #81, opened 2017-02-22
updating Tensile configs to remove bugs (ThreadTileEdge=8) and benchm…

guacamoleo:
checking in new Tensile benchmark files. These files avoid using solutions which the compiler may not be compiling correctly.


tingxingdong:
Why merge it so quickly? It seems it can not even compile.

PR #90, opened 2017-03-24
add option to compile with -fPIC

amcamd:
resolves #___

Summary of proposed changes:
-  
-  
-

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ endif()
 
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build rocBLAS as a shared library" OFF )
+option( COMPILE_WITH_FPIC "Compile with -fPIC" ON )
 
 set( HIP_ROOT /opt/rocm/hip CACHE PATH "Specify hip installation dir")
 set( BOOST_ROOT /opt/boost  CACHE PATH "Specify boost installation dir")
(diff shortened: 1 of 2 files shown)


kknox:
@amcamd 
I don't see it saving much time to explicitly add an option to our cmake for this

cmake -DCOMPILE_WITH_FPIC=ON
vs
cmake -DCMAKE_CXX_FLAGS="-fPIC"

and a flag like this doesn't make sense for non-gcc compilers.

amcamd:
The aim of the proposed change is for fPIC to be the default for compiling the library code. The user should have to specifically set -DCOMPILE_WITH_FPIC=OFF to get it to *NOT* compile with fPIC. 
I think the way you are suggesting would require the user to specify -DCMAKE_CXX_FLAGS="-fPIC" when running cmake. I guess we need to know what we want as default.
The proposed change passes COMPILE_WITH_FPIC to library and external projects. In the library it runs     SET(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -fPIC), so for non-gcc compilers it will do nothing.

pavanky:
@kknox any reason not to enable `-fPIC` for on gcc / clang ? This is pretty much required anytime you are building shared libraries.

pfultz2:
The proper way to set this is to use the [`POSITION_INDEPENDENT_CODE`](https://cmake.org/cmake/help/v3.0/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE) property in cmake. 

> any reason not to enable -fPIC for on gcc / clang by default? This is pretty much required anytime you are building shared libraries.

Cmake always adds `-fPIC` when building shared libraries, so setting this property is rarely needed.

pavanky:
@pfultz2 I have seen this be an issue when static libraries built without `-fPIC` option are linked into dynamic libraries.

amcamd:
I have not made any measurements comparing the performance of the rocBLAS static library with and without -fPIC, but in the past (about 10 years ago) libraries compiled with -fPIC could be 5% slower. Maybe current compilers do not have this small performance degrade with -fPIC.

PR #95, opened 2017-04-27
add batched gemm in google tester; add a ploting python script

tingxingdong:
Summary of proposed changes:
-  add batched gemm in gtest
-  add a python script to automatically plot the gemm performance
-


kknox:
👍 
The failure in unit testing is expected and should be fixed in tooling in the near future.

PR #98, opened 2017-05-02
protect reading out of boundary in trtri gemm

tingxingdong:
resolves #___

Summary of proposed changes:
-  
-  
-


amcamd:
Does not solve errors in tritri, but is a correction in the code that can prevent future errors

PR #100, opened 2017-05-22
rocblas_set_matrix, rocblas_get_matrix

amcamd:
add new functions rocblas_set_matrix and rocblas_get_matrix.


clients/include/rocblas.hpp
```diff
@@ -39,6 +39,18 @@
...
+         rocblas_int elem_size,
+         const void *a, rocblas_int lda,
+         void *b, rocblas_int ldb);
+
+     rocblas_status
+     rocblas_get_matrix( rocblas_int rows, rocblas_int cols, 
+         rocblas_int elem_size,
+         const void *a, rocblas_int lda, 
+         void *b, rocblas_int ldb);
+
```
kknox: In the C++ header, these could be templated.  You could call the routine like:
`rocblas_set_matrix<double>(rocblas_int rows, ... );`
and this removes the need for `rocblas_int elem_size` parameter.

amcamd: removed rocblas_[set,get]_[matrix,vector] from rocblas.hpp. They are auxiliary functions, and none of the auxiliary functions are templated, and none of the auxiliary functions are in rocblas.hpp.

PR #101, opened 2017-05-30
updating to Tensile v3.0.0, removing order, renaming gemm_batched to gemm_strided_batched

guacamoleo:
resolves #___

Summary of proposed changes:
-  
-  
-


guacamoleo:
The testing crashes on trsm; I think this is a carried over "known compiler bug". All ~400 gemm tests passed.

PR #105, opened 2017-06-05
set CMake defaults for common use

amcamd:
The common use is a user building rocBLAS libraries and sample code. The CMakeList.txt options are set so that the common use can be configured with the command  "cmake ..".


guacamoleo:
Looks good to me.

PR #106, opened 2017-06-07
Set get pointer mode

amcamd:
- Add variable rocblas_pointer_mode to handle
- Add functions rocblas_set_pointer_mode and rocblas_get_pointer_mode
- Change enum name rocblas_mem_location to rocblas_pointer_mode

### DIFF
--- a/library/include/rocblas-auxiliary.h
+++ b/library/include/rocblas-auxiliary.h
@@ -19,7 +19,7 @@
 /* ============================================================================================ */
 /*! \brief  indicates whether the pointer is on the host or device. currently HIP API can only recoginize the input ptr on deive or not
 can not recoginize it is on host or not */
-ROCBLAS_EXPORT rocblas_mem_location
+ROCBLAS_EXPORT rocblas_pointer_mode
 rocblas_get_pointer_location(void *ptr);
 
 
--- a/library/include/rocblas-auxiliary.h
+++ b/library/include/rocblas-auxiliary.h
@@ -67,6 +67,20 @@ rocblas_get_stream( rocblas_handle handle, hipStream_t *stream );
 
 
 /********************************************************************************
+ * \brief set rocblas_pointer_mode
+ *******************************************************************************/
+ROCBLAS_EXPORT rocblas_status
+rocblas_set_pointer_mode( rocblas_handle handle, rocblas_pointer_mode pointer_mode);
+
+
+/********************************************************************************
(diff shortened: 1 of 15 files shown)


library/include/rocblas-auxiliary.h
```diff
@@ -19,7 +19,7 @@
 /* ============================================================================================ */
 /*! \brief  indicates whether the pointer is on the host or device. currently HIP API can only recoginize the input ptr on deive or not
 can not recoginize it is on host or not */
-ROCBLAS_EXPORT rocblas_mem_location
+ROCBLAS_EXPORT rocblas_pointer_mode
 rocblas_get_pointer_location(void *ptr);
```
b-sumner: This is not really the ideal name now.

amcamd: An alternative would be to have two enums 1) rocblas_pointer_mode   2) rocblas_mem_location  . rocblas_pointer_mode would be set by the user to indicate if their pointers are on host or device. This is the same as the use in cuBLAS.  rocblas_mem_location would have its current usage, that is, the function would check where the pointer is, and use this to call the appropriate device kernel.

b-sumner: Those wouldn't really be orthogonal.  I think one enum would be enough.  And the functions to access it should be intuitively named.

amcamd: We have the functions
- rocblas_get_pointer_mode
- rocblas_set_pointer_mode
These set and get the variable rocblas_pointer_mode. 

This is an additional function that finds if a pointer is located on the device or the host. Would it be better to name it rocblas_pointer_location   ?

b-sumner: Right, clearly get pointer mode is already taken.  But this function is returning a pointer mode.  How about rocblas_pointer_to_mode or...?

amcamd: How about  rocblas_get_mode_of_pointer  ?

Then the functions would be
rocblas_get_pointer_mode, rocblas_set_pointer_mode, rocblas_get_mode_of_pointer

b-sumner: I have a small preference for pointer_to_mode, but am not opposed to get_mode_of_pointer

amcamd: new commit changes function name to    rocblas_pointer_to_mode

PR #110, opened 2017-06-16
use unique_ptr for hipMalloc and malloc

amcamd:
use unique_ptr to avoid memory leak. The changes convert existing code to use unique_ptr

### DIFF
--- a/library/src/blas1/rocblas_amax.cpp
+++ b/library/src/blas1/rocblas_amax.cpp
@@ -204,21 +204,19 @@ rocblas_amax_template(rocblas_handle handle,
 
     rocblas_status status;
 
-    auto workspace_managed = rocblas_unique_ptr{rocblas::device_malloc(sizeof(T2) * blocks),rocblas::device_free};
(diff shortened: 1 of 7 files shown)


library/src/blas1/rocblas_amax.cpp
```diff
@@ -206,22 +204,26 @@ rocblas_amax_template(rocblas_handle handle,
 
     rocblas_status status;
 
-    T2 *workspace;
-    rocblas_int *workspace_index;
+    auto workspace_managed = rocblas_unique_ptr{rocblas::device_malloc(sizeof(T2) * blocks),rocblas::device_free};
+    T2 *workspace = (T2 *) workspace_managed.get();
```
kknox: I'm pretty sure you can treat an unique_ptr just like a normal pointer, i.e. there is no need to call the .get() method to retrieve the raw pointer.  You should be able to use the ! operator on the unique_ptr and never touch the raw pointer

amcamd: Changed to use ! operator on unique_ptr

PR #121, opened 2017-07-24
improve trsm tests

amcamd:
resolves # swdev-12710

Summary of proposed changes:
- initialize with L from Cholesky factorization of SPD matrix
-  test error and residual using reference trmm in place of reference trsm
-

### DIFF
--- a/clients/common/cblas_interface.cpp
+++ b/clients/common/cblas_interface.cpp
@@ -27,6 +27,11 @@ extern "C" {
     void    cgetrf_(int* m, int* n, rocblas_float_complex* A, int* lda, int* ipiv, int *info);
     void    zgetrf_(int* m, int* n, rocblas_double_complex* A, int* lda, int* ipiv, int *info);
 
+    void    spotrf_(char* uplo, int* m, float* A, int* lda, int *info);
+    void    dpotrf_(char* uplo, int* m, double* A, int* lda, int *info);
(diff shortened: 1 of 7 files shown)


b-sumner:
Won't these large increases let other problems through?  I'd rather better control the input matrices than raise the tolerance so much.

kknox:
@b-sumner  @tingxingdong 
xTRSM is something of a special precision nightmare because of the integrated matrix inversion operation.  I believe Tim tried to cook the input matrices with base2 numbers, but we are still struggling with what appears to be precision problems.

Any other ideas?  We could just test smaller matrices.

amcamd:
I am rewriting the initialization of the matrix. Below is from the new comment in the code

//  Random lower triangular matrix have condition number
//  that grows exponentially with matrix size. Random full
//  matrices have condition that grows linearly with 
//  matrix size. To generate a lower triangular matrix
//  with condition number that grows with matrix size 
//  start with full random matrix A. Calculate B <- A A^T. 
//  Make B strictly diagonal dominant. Next use Cholesky 
//  factorization to calculate L L^T = B. These L factors 
//  should have condition number approximately equal to
//  the condition number of the original matrix.

I will submit the new code. The new code should produce a starting L with better condition number.

b-sumner:
I think limiting the size to the minimum we're sure will test all of the code along with the proposed new matrix calculation should be sufficient.

PR #128, opened 2017-08-22
add haxpy

amcamd:
Add 16 bit or half precision haxpy function to rocBLAS. 

if (1==incx)&&(1==incy) 128 bit or half8 load-store and 32 bit or half2 arithmetic are used.

if (1!=incx)||(1!=incy) low performance 16 bit or __fp16 load-store and arithmetic are used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,17 @@ Unfortunately, googletest and lapack are not as easy to install.  Many distros d
 #### (optional, one time only)
 1.  `mkdir -p [ROCBLAS_BUILD_DIR]/release/deps`
 2.  `cd [ROCBLAS_BUILD_DIR]/release/deps`
-3.  `cmake [ROCBLAS_SOURCE]/deps`
-4.  `make -j$(nproc) install`
+3.  `cmake -DBUILD_BOOST=OFF [ROCBLAS_SOURCE]/deps`
+4.  `sudo make -j$(nproc) install`
 
 Once dependencies are available on the system, it is possible to configure the clients to build.  This requires a few extra cmake flags to the library cmake configure script:
 * `CXX=/opt/rocm/bin/hcc cmake -DCMAKE_INSTALL_PREFIX=package -DBUILD_CLIENTS=ON -DBUILD_CLIENTS_TESTS=ON -DBUILD_CLIENTS_BENCHMARKS=ON [ROCBLAS_SOURCE]`
 
 If the dependencies are not installed into system defaults (like /usr/local ), you can optionally pass the CMAKE_PREFIX_PATH to cmake to help find them.
 * `-DCMAKE_PREFIX_PATH="<semicolon separated install list>"`
+
+Build with
+'make -j$(nproc)'
 
 ## Migrating libraries to ROCm from OpenCL
 [clBLAS][] demonstrated significant performance benefits of data parallel (GPU) computation when applied to solving dense
--- a/clients/common/cblas_interface.cpp
+++ b/clients/common/cblas_interface.cpp
@@ -5,6 +5,7 @@
 
 
 #include <typeinfo>
+#include <memory>
 #include "rocblas.h"
 #include "cblas_interface.h"
 #include "cblas.h"
@@ -108,6 +109,32 @@ extern "C" {
     }
 
     //axpy
+    template<>
+    void cblas_axpy<rocblas_half>( rocblas_int n,
+                            rocblas_half alpha,
+                            rocblas_half *x, rocblas_int incx,
+                            rocblas_half *y, rocblas_int incy)
+    {
+        rocblas_int abs_incx = incx >=0 ? incx : -incx;
+        rocblas_int abs_incy = incy >=0 ? incy : -incy;
+        float alpha_float = static_cast<float>(alpha);
+        std::unique_ptr<float []> x_float(new float[n*abs_incx]());
+        std::unique_ptr<float []> y_float(new float[n*abs_incy]());
+        for (int i = 0; i < n; i++)
+        {
+            x_float[i*abs_incx] = static_cast<float>(x[i*abs_incx]);
+            y_float[i*abs_incy] = static_cast<float>(y[i*abs_incy]);
+        }
+
+        cblas_saxpy(n, alpha_float, x_float.get(), incx, y_float.get(), incy);
+
+        for (int i = 0; i < n; i++)
+        {
+            x[i*abs_incx] = static_cast<rocblas_half>(x_float[i*abs_incx]);
+            y[i*abs_incy] = static_cast<rocblas_half>(y_float[i*abs_incy]);
+        }
+    }
+
     template<>
     void cblas_axpy<float>( rocblas_int n,
                             float alpha,
--- a/clients/include/testing_axpy.hpp
+++ b/clients/include/testing_axpy.hpp
@@ -17,6 +17,15 @@
 
 using namespace std;
 
+#define CLEANUP()                                                  \
+do {                                                               \
+        if (dx)           CHECK_HIP_ERROR(hipFree(dx));            \
+        if (dy)           CHECK_HIP_ERROR(hipFree(dy));            \
+        if (dy_devptr)    CHECK_HIP_ERROR(hipFree(dy_devptr));     \
+        if (alpha_devptr) CHECK_HIP_ERROR(hipFree(alpha_devptr));  \
+        rocblas_destroy_handle(handle);                            \
+} while (0)
+
 /* ============================================================================================ */
 template<typename T>
 void testing_axpy_bad_arg()
@@ -33,8 +42,9 @@ void testing_axpy_bad_arg()
     status = rocblas_create_handle(&handle);
     verify_rocblas_status_success(status,"ERROR: rocblas_create_handle");
 
-    if(status != rocblas_status_success) {
-        rocblas_destroy_handle(handle);
+    if(status != rocblas_status_success) 
+    {
+        rocblas_destroy_handle(handle);                            \
         return;
     }
 
@@ -67,7 +77,7 @@ void testing_axpy_bad_arg()
                     dx_null, incx,
                     dy, incy);
 
-        verify_rocblas_status_invalid_pointer(status,"Error: x, y, is nullptr");
+        verify_rocblas_status_invalid_pointer(status,"Error: x is nullptr");
     }
     // testing for (nullptr == dy)
     {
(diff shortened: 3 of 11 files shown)


README.md
```diff
@@ -44,15 +44,18 @@ Unfortunately, googletest and lapack are not as easy to install.  Many distros d
 #### (optional, one time only)
 1.  `mkdir -p [ROCBLAS_BUILD_DIR]/release/deps`
 2.  `cd [ROCBLAS_BUILD_DIR]/release/deps`
-3.  `cmake [ROCBLAS_SOURCE]/deps`
-4.  `make -j$(nproc) install`
+3.  `cmake -DBUILD_BOOST=OFF [ROCBLAS_SOURCE]/deps`
```
kknox: I know that our jenkinsfile doesn't build boost because it installs boost from apt, but I don't want to give advice to users NOT to build boost.  I default it on because it should be safe to build boost no matter if they need it or not.  The only penalty is that they are using extra hard drive space.

clients/common/cblas_interface.cpp
```diff
@@ -109,6 +110,32 @@ extern "C" {
...
     template<>
+    void cblas_axpy<rocblas_half>( rocblas_int n,
+                            rocblas_half alpha,
+                            rocblas_half *x, rocblas_int incx,
+                            rocblas_half *y, rocblas_int incy)
+    {
+        rocblas_int abs_incx = incx >=0 ? incx : -incx;
+        rocblas_int abs_incy = incy >=0 ? incy : -incy;
+        float alpha_float = static_cast<float>(alpha);
+        std::unique_ptr<float []> x_float(new float[n*abs_incx]());
```
kknox: Using a std::unique_ptr<> for an array is fine, but I bet it would be easier to declare and you get array notation if you use std::array<>

kknox: it looks like you are not packing your temp buffer.  You copy/cast the sparse half array into a half float array; is there any reason why you wouldn't pack the array while you are copying?

clients/include/testing_axpy.hpp
```diff
@@ -17,6 +17,15 @@
 
 using namespace std;
 
+#define CLEANUP()                                                  \
```
kknox: Code like this would benefit from the application of std::unique_ptr<>.  Then the CLEANUP macro goes away, and you don't have to worry about calling CLEANUP right before every return.

kknox: You can specialize std::unique_ptr<> with a custom deleter, to call hipFree.  You can google that for syntax.

clients/include/testing_axpy.hpp
```diff
@@ -102,21 +112,20 @@ void testing_axpy_bad_arg()
...
+    T *dx = NULL, *dy = NULL, *dy_devptr = NULL, *alpha_devptr = NULL;
 
     rocblas_handle handle;
     rocblas_status status;
     status = rocblas_create_handle(&handle);
     verify_rocblas_status_success(status,"ERROR: rocblas_create_handle");
 
     if(status != rocblas_status_success) {
-        rocblas_destroy_handle(handle);
+        CLEANUP();
```
kknox: Syntax like this could dissappear with std::unique_ptr.  See comments above.

library/include/rocblas-types.h
```diff
@@ -28,8 +28,11 @@ typedef int32_t rocblas_int;
 typedef float2  rocblas_float_complex;
 typedef double2 rocblas_double_complex;
 // half type TODO put name of half here
-typedef float    rocblas_half;
+typedef __fp16   rocblas_half;
 typedef float2   rocblas_half_complex;
+// half vectors
+typedef __fp16 half8 __attribute__((ext_vector_type(8)));
+typedef __fp16 half2 __attribute__((ext_vector_type(2)));
```
kknox: Half2 comes from hip_fp16.h, this is duplicate.

kknox: I'm considering if we need to put compiler guards in our header file, as some compilers might not know what a __fp16 is.  hip_fp16 does this.

library/src/blas1/rocblas_axpy.cpp
```diff
@@ -3,13 +3,11 @@
...
  * ************************************************************************ */
 #include <hip/hip_runtime.h>
-
- 
-
 #include "rocblas.h"
- 
 #include "definitions.h"
 
+extern "C" half2 __v_pk_fma_f16(half2, half2, half2) __asm("llvm.fma.v2f16");
```
kknox: Should this be promoted to a HIP header?

library/src/blas1/rocblas_axpy.cpp
```diff
@@ -135,33 +248,150 @@ rocblas_axpy_template(rocblas_handle handle,
...
     dim3 threads(NB_X, 1, 1);
 
     hipStream_t rocblas_stream;
     RETURN_IF_ROCBLAS_ERROR(rocblas_get_stream(handle, &rocblas_stream));
 
-    if( rocblas_pointer_to_mode((void*)alpha) == rocblas_pointer_mode_device ){
-        hipLaunchKernel(HIP_KERNEL_NAME(axpy_kernel_device_scalar), dim3(blocks), dim3(threads), 0, rocblas_stream, n, alpha, x, incx, y, incy);
+    if (rocblas_pointer_mode_device == rocblas_pointer_to_mode((void*)alpha))
+    {
+        hipLaunchKernel(HIP_KERNEL_NAME(axpy_kernel_device_scalar),
```
kknox: I think we are supposed to start using the newew API: [hipModuleLaunchKernel ](https://github.com/ROCm-Developer-Tools/HIP/blob/9f4ea9255cb44af249d156204b60b6a7fa206143/docs/markdown/hip_porting_driver_api.md#using-hipmodulelaunchkernel)

Lets stop adding more of the deprecated hipLaunchKernel.  This comment applies to the rest of the file

README.md
amcamd: I found that if you build Boost the build will fail. If you run the given command 
sudo apt install libboost-program-options-dev
and then build with -DBUILD_BOOST=OFF it will build correctly. This is what is done in the Jenkinsfile

clients/include/testing_axpy.hpp
amcamd: We could change this throughout the test code. It would require changing many other files, and i would likely be best to do it in a separate pull request.

clients/include/testing_axpy.hpp
amcamd: See reply above, this will require a change throughout the test code, and it should likely be in a separate pull request.

library/src/blas1/rocblas_axpy.cpp
adityaatluri: HIP half2 is a mess. It is better this way.

library/include/rocblas-types.h
adityaatluri: I recommend not including hip_fp16.h at all in rocblas as of now. HIP half2 will not generate vector code.

clients/common/cblas_interface.cpp
amcamd: std::array would allow for array notation for x_float (and y_float), but we are already using i*abs_incx for x, so we use the same indexing on the LHS and RHS of the equation.

We could pack the temp buffer, but it may be simpler to have the same incx and incy parameters in cblas_axpy as in cblas_saxpy.

README.md
kknox: OK, then I need to debug the boost build; hopefully that's easy, it used to work.  I only do this in the jenkinsfile to save build time in our CI

clients/common/cblas_interface.cpp
kknox: I think its worth the effort to pack the temp buffers; you can create terrible performance edge cases with sparse vectors.  The increase in complexity would be worthwhile to me.

clients/include/testing_axpy.hpp
kknox: A different PR could handle cleaning up the rest of the codebase.  For this PR, you could define the unique_ptr< hipmemory* > and not create more code to clean up later.

library/src/blas1/rocblas_axpy.cpp
kknox: ok

clients/common/cblas_interface.cpp
amcamd: In separate discussion agreed that this can stay as it is test code, not library code.

pfultz2: `std::array<>` requires the size to be known at compile-time. For runtime sizes a `std::vector` can be used, such as:

```cpp
std::vector<float> x_float(n*abs_incx);
```

But this will store the size in `std::vector`, so it takes extra space than the `unique_ptr`. Most of the time, its not a problem.

library/src/blas1/rocblas_axpy.cpp
b-sumner: It is not better this way!  Why don't we fix HiP instead of this ugly workaround.  Please do not add this.

library/include/rocblas-types.h
b-sumner: Then we should fix HiP instead of adding this to the library.  We do not want developers looking at this code and copying this in their own code.

library/src/blas1/rocblas_axpy.cpp
b-sumner: Just to add here too:  We do not want developers seeing this ugliness and copying it into their own code.

library/src/include/definitions.h
```diff
@@ -14,6 +14,18 @@
  * thereby it can include top-level definitions included by all
  ******************************************************************************/
 
+// half vectors
+typedef __fp16 half8 __attribute__((ext_vector_type(8)));
+typedef __fp16 half2 __attribute__((ext_vector_type(2)));
+extern "C" half2 __v_pk_fma_f16(half2, half2, half2) __asm("llvm.fma.v2f16");
```
b-sumner: There's no need to call the function __v_pk_fma_f16, although the final result is likely the v_pk_fma_f16 ISA instruction.  I'd suggest calling it llvm_fma_v2f16 since what is really being requested is the LLVM fma intrinsic on type v2f16 == half2.

adityaatluri: Does it matter to change the name? Using __v_pk_fma_f16 is better readable as it says it is an intrinsic and it generates v_pk_fma_f16 isa.

b-sumner: It *may* generate that ISA.  But the specific request is for the LLVM intrinsic.  This change is an improvement.

library/src/blas1/rocblas_axpy.cpp
amcamd: Will refactor in a separate pull request to use new API

PR #129, opened 2017-08-23
correct rocblas_dot for inc<0

amcamd:
rocblas_dot was returning rocblas_status_success when incx<0 || incy<0, and tests were not catching the error.
Remove the return for incx<0 || incy<0. rocblas_dot gives correct behavior for incx<0||incy<0, so no change is needed other than to remove the incorrect return. Also change tests to catch this error.


amcamd:
trsm errors are NaN errors likely from gemm NaN propagation when beta == 0

kknox:
Should we filter out these failures for now?  Do these trsm failures happen randomly?

PR #140, opened 2017-09-29
add cold and hot loops to improve gemm timing repeatability

amcamd:
add variables number_cold_calls and number_hot_calls to testing_gemm.hpp . For rocblas_bench gemm will be called number_cold_calls before timing number_hot_calls. This will give greater repeatability for rocblas_bench timing of gemm.

### DIFF
--- a/clients/include/testing_gemm.hpp
+++ b/clients/include/testing_gemm.hpp
@@ -276,6 +276,8 @@ void testing_gemm_bad_arg()
 template<typename T>
 rocblas_status testing_gemm(Arguments argus)
 {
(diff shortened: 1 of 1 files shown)


guacamoleo:
what did you think about my proposal to report all raw data then report analysis?

amcamd:
I think we need to keep rocblas_bench simple. If a user needs descriptive statistics it will be best if they write code themselves, and they can experiment and make sure the statistics they are collecting are valid descriptors of the data.

amcamd:
After off-line discussion, this change will be checked in, and future changes will be made to provide statistical measures of performance.

PR #144, opened 2017-10-06
geam code and tests

amcamd:
Add code and tests for geam,   C <- alpha A + beta B


library/src/blas3/geam_device.h
```diff
@@ -0,0 +1,124 @@
...
+    const T * __restrict__ B, rocblas_int ldb,
+    T beta,
+          T *              C, rocblas_int ldc)
+{
+    rocblas_int tx = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    rocblas_int ty = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
+
+    if (tx < m && ty < n)
+    {
+        if (transA == rocblas_operation_none && transB == rocblas_operation_none)
```
b-sumner: I haven't checked if this is following the style elsewhere, but I would suggest only calculating the offsets under the conditions, and doing the actual work (loads, compute, store) outside of the conditions. Also, I'd consider using one fma and one multiply for potentially improved accuracy.

amcamd: The following commit contains improved code that addresses the comments 8a54224e8cf3034322b6c4b5ac6605f122078a21

library/src/blas3/geam_device.h
```diff
@@ -19,22 +19,30 @@ geam_device(
...
+            b_index = tx + ty*ldb;
         }
-        else if (transA == rocblas_operation_transpose && transB == rocblas_operation_transpose)
+        else
         {
-            C[tx + ldc*ty] = (alpha) * A[lda*tx + ty] + beta * B[ldb*tx + ty];
+            b_index = tx*ldb + ty;
         }
+
+        C[c_index] = alpha * A[a_index];
```
b-sumner: What is the reason to break this into two statements?

amcamd: I must have misunderstood your previous comment, This was intended to address using one fma and one multiply for potentially improved accuracy.

b-sumner: Thanks.  Unfortunately, the compiler isn't required to form an FMA from the expression a*b +c.  An explicit call to fma is required to ensure an fma is used.

adityaatluri: But will it convert to mad or mac op?

b-sumner: It will in cases where it would produce the same result.

amcamd: change to use fma

library/src/blas3/geam_device.h
```diff
@@ -42,7 +42,7 @@ geam_device(
         }
 
         C[c_index] = alpha * A[a_index];
-        C[c_index] += beta * B[b_index];
+        C[c_index] = fma(beta, B[b_index], C[c_index]);
```
b-sumner: Thanks.  But again, why prefer two statments to a single fma(beta, B[i], alpha*A[i])

amcamd: Changed to use a single statement with fma.

library/src/blas3/geam_device.h
adityaatluri: Isn't the name be gemm? Or using geam is intentional?

adityaatluri: Is FMA faster than MAD?

amcamd: geam is a BLAS like extension for matrix add/transpose (the transpose is out of place, not in place). It is different to gemm.

adityaatluri: @amcamd thanks!

kknox:
You merged this, but I would still like to know if the restrict qualifier is valid in this kernel

amcamd:
You are correct, the API should support in place operations where the pointers will alias, so the restrict qualifier will not be valid. I will create a new pull request to cover the in-place case. These in-place cases do require lda=ldc and transA=N when C=A, and also ldb=ldc and trandB=N when C=B.

kknox:
@amcamd @guacamoleo 
Guac showed me I forgot to click 'submit' on my code-review, so you never got a chance to see my comments.  Sorry about the confusion this was for you!

PR #150, opened 2017-10-16
change api amax -> iamax

amcamd:
Correct API with change
rocblas_samax -> rocblas_isamax
rocblas_damax -> rocblas_idamax


kknox:
Looks good to me (LGTM)

PR #152, opened 2017-10-19
add error message for cmake3.5.2

amcamd:
rocBLAS build fails if cmake 3.5.2 is used. It passes if cmake 3.5.1 is used. 

Add descriptive error message that will help users avoid the build fail with cmake 3.5.2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,19 +28,24 @@ find_package( ROCM CONFIG QUIET PATHS /opt/rocm )
 if( NOT ROCM_FOUND )
   set( rocm_cmake_tag "master" CACHE STRING "rocm-cmake tag to download" )
   file( DOWNLOAD https://github.com/RadeonOpenCompute/rocm-cmake/archive/${rocm_cmake_tag}.zip
-      ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip SHOW_PROGRESS)
+      ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip STATUS status LOG log)
+
+  list(GET status 0 status_code)
+  list(GET status 1 status_string)
+
+  if(NOT status_code EQUAL 0)
+    message(FATAL_ERROR "error: downloading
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -28,11 +28,20 @@ find_package( ROCM CONFIG QUIET PATHS /opt/rocm )
...
   set( rocm_cmake_tag "master" CACHE STRING "rocm-cmake tag to download" )
   file( DOWNLOAD https://github.com/RadeonOpenCompute/rocm-cmake/archive/${rocm_cmake_tag}.zip
-    ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip )
+      ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip SHOW_PROGRESS)
 
-  execute_process( COMMAND ${CMAKE_COMMAND} -E tar xzf ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip
+  execute_process( COMMAND ${CMAKE_COMMAND} -E tar xzvf ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip
     WORKING_DIRECTORY ${PROJECT_EXTERN_DIR} )
 
+  if( NOT EXISTS ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag})
```
kknox: This message is too specific to the one issue that we observe.  The problem is not related to cmake 3.5.2, the problem is related to a custom cmake build, that does not properly have the security options enabled

kknox:
@pfultz2 showed me some file(DOWNLOAD) validation code.  Try to use something along the lines of the following:
```cmake 
file( DOWNLOAD https://github.com/RadeonOpenCompute/rocm-cmake/archive/${rocm_cmake_tag}.zip ${PROJECT_EXTERN_DIR}/rocm-cmake-${rocm_cmake_tag}.zip 
list(GET RESULT_LIST 0 RESULT)
list(GET RESULT_LIST 1 RESULT_MESSAGE)
if(NOT RESULT EQUAL 0)
    message(FATAL_ERROR "Download failed: ${RESULT_MESSAGE}")
endif()
```

amcamd:
Added more general error message similar to that by @pfultz2

PR #157, opened 2017-11-08
targets Tensile v3.3.3 which fixes thread safety and assembly kernels…

guacamoleo:
… on multiple gpus

resolves #___

Summary of proposed changes:
-  
-  
-


tingxingdong:
so is this not a hipGetDevice() bug?  but only the hash ID is wrong?

PR #159, opened 2017-11-09
use unique_ptr and test both pointer modes

amcamd:
refactor tests
* use std::unique_ptr for hipMalloc and hipFree
* use std::unique_ptr for rocblas_create_handle and rocblas_destroy_handle
* all tests for rocblas_pointer_mode_host and rocblas_pointer_mode_device

### DIFF
--- a/clients/gtest/blas1_gtest.cpp
+++ b/clients/gtest/blas1_gtest.cpp
@@ -72,7 +72,8 @@ Representative sampling is sufficient, endless brute-force sampling is not neces
 
 //vector of vector, each pair is a {alpha, beta};
 //add/delete this list in pairs, like {2.0, 4.0}
-vector<vector<double>> alpha_beta_range = { {1.0, 0.0},
+vector<vector<double>> alpha_beta_range = { 
+                                            {1.0, 0.0},
                                             {2.0, -1.0}
                                           };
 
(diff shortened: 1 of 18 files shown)


clients/gtest/blas1_gtest.cpp
```diff
@@ -211,18 +223,7 @@ TEST_P(blas1_gtest, axpy_float)
...
-            EXPECT_EQ(rocblas_status_invalid_size, status);
-        }
-        else if( arg.incx < 0){
-            EXPECT_EQ(rocblas_status_invalid_size, status);
-        }
-        else if( arg.incy < 0){
-            EXPECT_EQ(rocblas_status_invalid_size, status);
-        }
-    }
+    EXPECT_EQ(rocblas_status_success, status);
```
kknox: Does this reduce test precision?  Looks like the old code tried to decipher that the returned error code was a particular value.

clients/gtest/blas1_gtest.cpp
```diff
@@ -174,13 +175,24 @@ TEST_P(blas1_gtest, iamax_float)
...
+    // GetParam return a tuple. Tee setup routine unpack the tuple
+    // and initializes arg(Arguments) which will be passed to testing routine
+    // The Arguments data struture have physical meaning associated.
+    // while the tuple is non-intuitive.
+    Arguments arg = setup_blas1_arguments( GetParam() );
+    rocblas_status status = testing_iamax<double>( arg );
+    // if not success, then the input argument is problematic, so detect the error message
+    if(status != rocblas_status_success)
+    {
+        EXPECT_EQ(rocblas_status_success, status);
```
kknox: Do we need to protect the EXPECT_EQ with a branch?

clients/gtest/blas1_gtest.cpp
kknox: Maybe you meant to use your new macro CHECK_ROCBLAS_ERROR()?

clients/gtest/blas1_gtest.cpp
kknox: Possibly, you meant to use your new macro CHECK_ROCBLAS_ERROR()

amcamd: Thanks, you are correct, will remove branch and use CHECK_ROCBLAS_ERROR()

PR #171, opened 2017-11-21
fix __global__ function rocblas_fmadd_half2 to return void

amcamd:
change function rocblas_fmadd_half2 so that it returns void. The result is returned as a half2* argument, not as the function return.

### DIFF
--- a/library/src/blas1/rocblas_axpy.cpp
+++ b/library/src/blas1/rocblas_axpy.cpp
@@ -88,8 +88,8 @@ axpy_kernel_device_scalar(hipLaunchParm lp,
     }
 }
 
-__global__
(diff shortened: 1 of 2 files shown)


kknox:
I don't see any other __global__ problems.
@AlexVlx , does this fix suffice?

AlexVlx:
@kknox @amcamd thank you gentlemen. I believe that it does, the one other instance was in tensile and I think that Dr. Tanner is working on it:)

guacamoleo:
I corrected Tensile's problem and its checked into the develop branch.

David E. Tanner
_________________________________________________________
MTS Software Engineer  |   Radeon Technologies Group – Open Compute


________________________________

PR #177, opened 2017-11-30
add clang-format

amcamd:
Add clang-format consistently with clang-format in MIOpen.

Jenkins runs clang-format runs against pull request and the pull request will fail if any files do not comply with the specified format in .clang-format. 

At this stage the user can
*  format files individually with: clang-format-3.8 -style=file -i <path-to-source-file> 
* this will change the files in place
* commit the changed files
* the pull request should now pass

Alternatively the user can
* ignore the failure and merge the pull request

The user can also run the following and it will cause files to be automatically formatted before they are commited:  ./.githooks/install . Any pull request containing already formatted commits should pass the format check.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,90 @@
+---
+Language:        Cpp
+AccessModifierOffset: 0
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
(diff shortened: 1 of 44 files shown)


amcamd:
this pull request is stale, transfer changes to new pull request

Issue #143, opened 2017-10-03
roc::rocblas target and C++ compilers


jszuppe:
If you use `roc::rocblas` target (defined in `rocblas-config.cmake`) in your cmake script to link to rocBLAS, you have to use `hcc` as you C++ compiler. I think this happens because `hip::hip_hcc` in [here](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/CMakeLists.txt#L127) is marked `PUBLIC`, which means its dependencies are transitive and compilation options are passed to `roc::rocblas` (I understand that this is so the HIP dirs are included). Unfortunately, some of those options work only with `hcc` compiler.

I just want to know if it's supposed to be that way (`roc::rocblas` works with `hcc`, if you want to use different compiler use `rocblas_INCLUDE_DIRS` , `rocblas_LIB_INSTALL_DIR`,  and `-lrocblas`), or if this is work-in-progress.

kknox:
Hi @jszuppe, good to see you again.  Thanks for the feedback. 

I made the `hip::hip_hcc` dependency public because we expose the hip header files through the [rocblas header files](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/include/rocblas-types.h#L16).  However, I see your point that the public compiler flags is inconvenient.  I don't want to drag a compiler dependency forward.

Ultimately, I think the real fix is going to be to split the hip exported targets, such that I can make the hip header files public without making the link flags public.  That will require refactoring up in the hip layer.  That may take some time.

Until then, I can probably hide the `hip::hip_hcc` dependency privately and export the hip header files publicly from our config targets.

jszuppe:
> I made the hip::hip_hcc dependency public because we expose the hip header files through the rocblas header files. 

Yes, you documented why you did it. I just wasn't sure if the mentioned consequence of that was intended or not. I assumed it might be intended since you dropped support for `nvcc`, and maybe want to push people to use `hcc` as C++ compiler on ROCm.

> Ultimately, I think the real fix is going to be to split the hip exported targets, such that I can make the hip header files public without making the link flags public. That will require refactoring up in the hip layer. That may take some time.

I agree it should be addressed in the hip layer. 

> Until then, I can probably hide the hip::hip_hcc dependency privately and export the hip header files publicly from our config targets.

[...]

PR #178, opened 2017-11-30
Centos

kknox:
Summary of proposed changes:
-  Provides build support for CentOS 7
-  Changes rocblas link dependency on hip to private
- rocblas definition rocblas_half changed, because gcc compilers did not understand __fp16 type
  - this is a potential backward breaking change, if users used half before


clients/include/testing_axpy.hpp
```diff
@@ -84,10 +84,15 @@ rocblas_status testing_axpy(Arguments argus)
     rocblas_int N = argus.N;
     rocblas_int incx = argus.incx;
     rocblas_int incy = argus.incy;
-    T h_alpha = argus.alpha;
     rocblas_int safe_size = 100;   // arbitrarily set to 100
-
-    std::unique_ptr<rocblas_test::handle_struct> test_handle(new rocblas_test::handle_struct);
+    T h_alpha = 0;
```
amcamd: I think this needs to change to     T h_alpha = argus.alpha;

kknox: It's initialized just below, depending on T type

clients/include/testing_axpy.hpp
```diff
@@ -124,6 +129,8 @@ rocblas_status testing_axpy(Arguments argus)
     srand(1);
     rocblas_init<T>(hx, 1, N, abs_incx);
     rocblas_init<T>(hy_1, 1, N, abs_incy);
+    // rocblas_init<T>(hx, 1, N, abs_incx, 1.0);
```
amcamd: remove these commented lines

kknox:
I believe this may fix #143; I can now build rocblas clients with the host gcc compiler.  

Our clients still link to HiP because they use hip for memory allocation, but the rocblas library library does not force the clients to be hcc or hipcc.

PR #181, opened 2017-12-04
adding clang-format style and scripts

amcamd:
Add clang-format consistently with clang-format in MIOpen.

Jenkins runs clang-format runs against pull request and the pull request will fail if any files do not comply with the specified format in .clang-format.

At this stage the user can
* format files individually with: clang-format-3.8 -style=file -i <filename>. This will format the files in place. Commit the changed files. The pull request should now pass
* Run the following and it will cause files to be formatted in place before they are committed: ./.githooks/install . Pull requests containing already formatted commits should pass the format check.
*  ignore the failure and merge the pull request

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,90 @@
+---
+Language:        Cpp
+AccessModifierOffset: 0
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: false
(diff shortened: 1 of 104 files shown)


amcamd:
hcc_rocm tests pass, the bleeding edge hcc_ctu tests fail, and we are looking into the hcc_ctu failure.

PR #186, opened 2018-01-02
CI add fedora

kknox:
Summary of proposed changes:
-  Enable fedora builds of rocblas
-  Update base images to rocm/dev images
-  Compile client apps with host compilers


docker/dockerfile-build-fedora
```diff
@@ -11,40 +9,39 @@ ARG user_uid
...
-    gfortran \
-    libboost-program-options-dev \
-    rpm \
+    make \
+    cmake \
+    clang \
+    gcc-c++ \
+    gcc-gfortran \
+    python34 \
+    PyYAML \
```
amcamd: A question: is the change from python2.7 to python34 for "non Tensile" parts of the code? Does Tensile still use python2.7?

kknox: Yes, I needed it for some other dependency, but unfortunately I can't remember which one anymore.  I believe that Fedora already has python2 installed.

kknox:
The build failure is related to a known hcc-ctu, and is safe to merge

PR #188, opened 2018-01-12
Add logging

amcamd:
log each rocBLAS function call to file ~/rocblas_logfile.csv if environment variable ROCBLAS_LAYER=1.

C++ code for variadic parameter pack is in file library/src/include/handle.h


clients/include/testing_logging.hpp
```diff
@@ -0,0 +1,306 @@
...
+#include <unistd.h>
+#include <pwd.h>
+#include <fstream>
+#include <string>
+#include <algorithm>
+#include <iterator>
+
+using namespace std;
+
+// replaces X in string with s, d, c, z or h depending on typename T
```
kknox: AXPY and AMAX have natural X's in the name.  You may need to find a unique character glyph to replace

clients/include/testing_logging.hpp
```diff
@@ -0,0 +1,306 @@
...
+    }
+    else if(std::is_same<T, rocblas_double_complex>::value)
+    {
+        std::replace(input_string.begin(), input_string.end(), 'X', 'z');
+    }
+    else if(std::is_same<T, rocblas_half>::value)
+    {
+        std::replace(input_string.begin(), input_string.end(), 'X', 'h');
+    }
+    return input_string;
```
kknox: Maybe you could document what characters we should use for complex half types.  Could we instead use of the proposed codes, such as r16?

library/include/rocblas-types.h
```diff
@@ -106,6 +106,13 @@ typedef enum rocblas_pointer_mode_ {
     rocblas_pointer_mode_device = 1
 } rocblas_pointer_mode;
 
+/*! \brief Indicates if layer is active*/
+typedef enum rocblas_layer_mode {
```
kknox: Make the intention of these to be bitmasks with the 0x prefix, i.e.  0x1, 0x2, 0x4

library/src/handle.cpp
```diff
@@ -1,28 +1,78 @@
...
+
+    // set layer_mode from vaule of environment variable ROCBLAS_LAYER
+    char* str_layer_mode;
+    if((str_layer_mode = getenv("ROCBLAS_LAYER")) == NULL)
+    {
+        layer_mode = rocblas_layer_mode_none;
+    }
+    else
+    {
+        int int_layer_mode = atoi(str_layer_mode);
```
kknox: You can read straight from atoi() into a rocblas_layer_mode type.  That could get rid of the temp variable.

library/src/handle.cpp
```diff
@@ -1,28 +1,78 @@
...
+    }
+
+    // open log file
+    if(layer_mode & rocblas_layer_mode_logging)
+    {
+        // construct filepath for log file in home directory
+        const char* homedir_char;
+        std::string homedir_str;
+
+        if((homedir_char = getenv("HOME")) == NULL)
```
kknox: Instead of reading the HOME env variable, we should attempt to write log files into the PWD.  First, you have to check if the PWD is writeable, and if it is not, then the user should be able to specify an alternate logging directory with an overriding ENV, ROCBLAS_LOG_DIR or some such.

library/src/handle.cpp
```diff
@@ -1,28 +1,78 @@
...
+        {
+            homedir_char = getpwuid(getuid())->pw_dir;
+        }
+        if(homedir_char == NULL)
+        {
+            std::cerr << "rocBLAS ERROR: cannot determine home directory for rocBLAS log file"
+                      << std::endl;
+            std::cerr << "rocBLAS ERROR: turn off logging or create a home directory" << std::endl;
+            exit(-1);
```
kknox: Remove the exit() call; libraries should not kill the host process.  If we can't log, output an error msg like you have above and let the app move on.

clients/include/testing_logging.hpp
amcamd: the function names are lower case, so it is rocblas_Xaxpy and rocblas_Xamax. Only the upper case X is replaced, not the lower case x.

kknox:
I think a new rocblas layer/bit should be defined, which helpfully outputs a string that can be copy/pasted to drive the rocblas-bench executable, with the exact parameters logged with the function.

library/src/handle.cpp
amcamd: fixed

library/include/rocblas-types.h
amcamd: fixed: use binary literal from C++ 14 and available in clang and gcc since 2013

library/src/handle.cpp
amcamd: fixed: log file will be in current working directory if it is writable, else in ROCBLAS_LOG_DIR

library/src/handle.cpp
amcamd: fixed: removed exit() call

clients/include/testing_logging.hpp
amcamd: I think  h is commonly used for half real, so we can keep with it. I have not seen demand for half complex. Can we delay deciding what we do for half complex till we need it, rather than specifying now what we will use.

PR #196, opened 2018-02-12
Modify cmake to detect rocm device to compile for

kknox:
Summary of proposed changes:
-  Fixed a bug in recognizing cached copies of Tensile
- CMake now calls out to an external utility to determine the hardware on the clients machine.  This allows tensile to tune appropriately for the developers hardware.  The old 'fat binary' approach is still valid, and a list of devices can be passed through AMDGPU_TARGETS


amcamd:
Added to pull request 198.

PR #203, opened 2018-03-05
Add cmake and install.sh controls for specifying which tensile

bensander:
Add options to control git branch, git fork, use local dir, change test logic.  Improves ability to run rocBLAS tests on local Tensile directories or custom forks&branches.


CMakeLists.txt
```diff
@@ -131,21 +131,44 @@ option( BUILD_SHARED_LIBS "Build rocBLAS as a shared library" ON )
...
+    set( Tensile_ROOT "${PROJECT_EXTERN_DIR}/Tensile" CACHE STRING "Local path of Tensile" )
+    set( Tensile_TensileConfig ${Tensile_ROOT}/Tensile/Source/TensileConfig.cmake )
+  
+    message (STATUS "using local Tensile from ${Tensile_TEST_LOCAL_PATH}, copied to ${Tensile_ROOT}")
+  else()
+    # Use the virtual-env setup and download package from specified repot:
+    include(virtualenv)
+
+    set( tensile_fork "ROCmSoftwarePlatform" CACHE STRING "Tensile fork to use" )
+    set( tensile_branch "develop" CACHE STRING "Tensile tag to download" )
```
amcamd: Can we make the default    master   . The reasoning is that before releasing the develop branch of rocBLAS is merged into master. We do not want to have to change this before the merge. It is dangerous if we forget to change it, as then the master branch of rocBLAS will be pulling the develop branch of Tensile.

CMakeLists.txt
```diff
@@ -131,21 +131,44 @@ option( BUILD_SHARED_LIBS "Build rocBLAS as a shared library" ON )
...
-  virtualenv_install("git+https://github.com/ROCmSoftwarePlatform/Tensile.git@${tensile_tag}")
-  list(APPEND CMAKE_PREFIX_PATH ${VIRTUALENV_HOME_DIR})
+  set( Tensile_TEST_LOCAL_PATH "" CACHE PATH "Use local Tensile directory instead of fetching a GitHub branch" )
+  set( CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories" )
+  set_property( CACHE Tensile_LOGIC PROPERTY STRINGS asm_full asm_lite hip_lite other )
+
+
+  if (Tensile_TEST_LOCAL_PATH)
+    # Use user's local Tensile path:
+    execute_process( COMMAND ${CMAKE_COMMAND} -E copy_directory ${Tensile_TEST_LOCAL_PATH} ${PROJECT_EXTERN_DIR}/Tensile)
```
pfultz2: This should install with `virtualenv_install("-e ${Tensile_TEST_LOCAL_PATH}")` so it will properly install the dependencies for tensile. Also, this links the source files back to tensile, so as you make changes it will be updated in rocblas as well.

CMakeLists.txt
pfultz2: Actually, I think this should use a commit hash or a tag as the default. Otherwise, we wont have reproducible builds for rocblas.

library/src/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ if( BUILD_WITH_TENSILE )
   endif( )
 
   # HACK: We include the config file directly because find_package(Tensile) is broken
-  include(${VIRTUALENV_HOME_DIR}/cmake/TensileConfig.cmake)
+  include( ${Tensile_TensileConfig} )
```
pfultz2: No need to change this as tensile should always be installed in the virtualenv.

bensander: @pfultz2  - I'm not familiar with virtualenv and my primitive attempts to implement the requested change were unsuccessful.    Not sure what the -e switch does but the install location appears different than downloading from Git?  This is the error when I try the changes you suggested:

CMake Error at library/src/CMakeLists.txt:37 (include):
  include could not find load file:

    /root/X/rocBLAS3-bs/build/release/virtualenv/cmake/TensileConfig.cmake


-- AMDGPU_TARGETS=gfx900
CMake Error at library/src/CMakeLists.txt:41 (TensileCreateLibrary):
  Unknown CMake command "TensileCreateLibrary".

pfultz2: > I'm not familiar with virtualenv and my primitive attempts to implement the requested change were unsuccessful.

The virtualenv just allows for pip to install to a local directory instead of to the system. 

> Not sure what the -e switch does but the install location appears different than downloading from Git? This is the error when I try the changes you suggested:

This is a flag for pip that is documented [here](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs). From the docs, this does appear to install tensile with `python setup.py develop`(ie development mode) instead. From the [docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode):

> To do this, use the setup.py develop command. It works very similarly to setup.py install or the EasyInstall tool, except that it doesn’t actually install anything. Instead, it creates a special .egg-link file in the deployment directory, that links to your project’s source code.

So this probably wont install the cmake file by default. Have you tried it without the `-e` flag?

I'll take a look at this in the morning as well.

pfultz2: I checked this and it does work correctly just dropping the `-e` flag. So doing `virtualenv_install("-${Tensile_TEST_LOCAL_PATH}")` will install tensile from a local directory.

PR #212, opened 2018-03-26
default logging stream to std::cerr

amcamd:
Change the default stream for logging to std::cerr. 

Alternatively, if the user sets the environment variable ROCBLAS_LOG_TRACE_PATH to a full path name, open a file in the full path name, and stream trace logging to that file. If the environment variable ROCBLAS_LOG_BENCH_PATH is set to a full path name, open  a file in the full path name, and stream bench logging to that file. 

If opening the files fails, try to open files stream_trace.csv or stream_bench.txt in the current working directory. If opening the files fails, default to streaming to std::cerr.

**Additional Information**

The environment variable ROCBLAS_LAYER is a bit mask as follows:
* if ROCBLAS_LAYER is not set there is no logging
* if ROCBLAS_LAYER is set to 1, then there is trace logging
* if ROCBLAS_LAYER is set to 2, then there is bench logging
* if ROCBLAS_LAYER is set to 3, then there is both trace and bench logging

trace logging outputs a line each time a rocBLAS function is called, the line contains the function name, and the values of arguments

bench logging outputs a line each time a rocBLAS function is called. The line can be used with the executable rocblas-bench to execute the function with the same arguments.


clients/include/testing_logging.hpp
```diff
@@ -71,15 +78,36 @@ void testing_logging()
...
+    char env_rocblas_layer[80] = "ROCBLAS_LAYER=3";
+    verify_equal(
+        putenv(env_rocblas_layer), 0, "failed to set environment variable ROCBLAS_LAYER=3");
+
+    // set environment variable to give pathname of for log_trace file
+    std::string trace_name1             = "stream_trace.csv";
+    char env_rocblas_log_trace_path[80] = "ROCBLAS_LOG_TRACE_PATH=stream_trace.csv";
+    verify_equal(putenv(env_rocblas_log_trace_path),
+                 0,
+                 "failed to set environment variable ROCBLAS_LAYER=3");
```
pauldwo: ROCBLAS_LAYER=3    should be   ROCBLAS_LOG_TRACE_PATH=stream_trace.csv

clients/include/testing_logging.hpp
```diff
@@ -71,15 +78,36 @@ void testing_logging()
...
+
+    // set environment variable to give pathname of for log_bench file
+    std::string bench_name1             = "stream_bench.txt";
+    char env_rocblas_log_bench_path[80] = "ROCBLAS_LOG_BENCH_PATH=stream_bench.txt";
+    verify_equal(putenv(env_rocblas_log_bench_path),
+                 0,
+                 "failed to set environment variable ROCBLAS_LAYER=3");
```
pauldwo: ROCBLAS_LAYER=3 should be ROCBLAS_LOG_BENCH_PATH=stream_bench.txt

clients/include/testing_logging.hpp
amcamd: Thanks, will fix

clients/include/testing_logging.hpp
amcamd: Thanks, good catch, will fix

library/src/include/logging.h
```diff
@@ -11,55 +11,71 @@
 #include <sys/param.h>
 
 /**
- * @brief Logging function
+ *  * @brief Logging function
```
pauldwo: Not sure whether this extra * is interpreted by doxygen.

amcamd: Thanks again,  fixed in next commit

PR #243, opened 2018-05-03
Add const guarantee for input matrix A.

iotamudelta:
Is required for tensorflow bringup.


iotamudelta:
@amcamd now passes all unit tests and format checks. Thanks for the patience!

PR #278, opened 2018-06-05
known_bug rocblas-test for stride_a = 0

amcamd:
Add tests for known bug. Error occurs when stride_a = 0,  and k < 256. Note that tests do pass for k = 32, but not for 31 or 33.


amcamd:
hcc-rocm-ubuntu tests pass. Failure in hcc-ctu not caused by this pull request.

PR #298, opened 2018-06-27
Check rocBlas API result code in benchmark warmup calls

bensander:


bensander:
will rebase and submit new pr

PR #299, opened 2018-06-27
add m chunking so offset fits in 32 bit register

amcamd:
Add chunking for BLAS dimension m to prevent the offset form exceeding the capacity of a 32 bit register. Previous pull requests #281 #289 are for chunking BLAS in the n dimension.

Also add tests for m and n chunking.

### DIFF
--- a/clients/gtest/gemm_gtest.cpp
+++ b/clients/gtest/gemm_gtest.cpp
@@ -117,8 +117,6 @@ const vector<vector<double>> full_alpha_beta_range = {
 // sgemm/dgemm,
 const vector<vector<char>> transA_transB_range = {{'N', 'N'}, {'N', 'T'}, {'C', 'N'}, {'T', 'C'}};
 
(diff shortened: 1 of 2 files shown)


amcamd:
Maybe the rocBLAS ERROR print statements should be replaced by assert statements

clients/gtest/gemm_gtest.cpp
```diff
@@ -327,3 +340,9 @@ INSTANTIATE_TEST_CASE_P(checkin_blas3_tiny,
                         Combine(ValuesIn(tiny_matrix_size_range),
                                 ValuesIn(full_alpha_beta_range),
                                 ValuesIn(transA_transB_range)));
```
zaliu: Did you mean to use transA_transB_NN_range instead?

amcamd: The tiny should use all values NN, NT, TN, TT. Commit below removes the unused transA_transB_NN_range.

PR #311, opened 2018-07-25
copy hip .yaml files to asm_full and asm_lite

amcamd:
Copy .yaml files for hgemm from hip_lite to asm_full and asm_lite.


amcamd:
#310 does the same thing as this commit

PR #337, opened 2018-08-17
gemm_ex: call tensile not gemm, add blocking

amcamd:
Directly call Tensile interface in place of calling rocblas_hgemm, rocblas_sgemm, rocblas_dgemm.

Also add blocking in the m and n dimentions to allow for offset to fit in 32 bit register


amcamd:
replace with #338

PR #400, opened 2018-10-31
Refactoring of rocBLAS testing and benchmarking to be data-driven

leekillough:
Summary of proposed changes:
-  Starting with [SDH]GEMM, refactor the correctness and performance tests to be data-driven based on YAML files.
-  Remove some redundant code, such as the old code which used ranges (start:stop:step).
-  Cleanup white space (convert tabs to spaces and remove trailing whitespace) -- incidental change for when other changes were made to a file.
- Make function-like macros use do {} while(0) idiom so that they are syntactically equivalent to function calls, as in if (test) MACRO(x); else MACRO(y);


.dir-locals.el
```diff
@@ -0,0 +1,28 @@
+;; This causes emacs to abide by formatting rules: No tab characters, except
```
leekillough: This just applies styles similar to clang-format within emacs to files of certain modes in all subdirectories. But mainly, it prevents tabs and trailing whitespace on lines.

.gitignore
```diff
@@ -37,3 +37,8 @@ tags
```
leekillough: emacs leaves certain temporary files around which should be ignored by Git.

clients/benchmarks/CMakeLists.txt
```diff
@@ -91,4 +91,22 @@ endif( )
 set_target_properties( rocblas-bench PROPERTIES CXX_EXTENSIONS NO )
 set_target_properties( rocblas-bench PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
 
+set( ROCBLAS_DGEMM_PERF_YAML "${CMAKE_CURRENT_BINARY_DIR}/dgemm_perf.yaml")
```
leekillough: This builds the dgemm_perf.yaml file from a Python script which is much more compressed than the YAML.
It then builds a binary file from the YAML file, for direct input to rocblas-bench.

clients/benchmarks/client.cpp
```diff
@@ -3,7 +3,8 @@
  * ************************************************************************ */
 
 #include <iostream>
-#include <stdio.h>
+#include <cstring>
```
leekillough: In C++11 and later, #include <cstdio> #include <cmath> #include <cstrings> etc. are preferred to #include <stdio.h> #include <math.h> #include <strings.h> etc.

clients/benchmarks/client.cpp
```diff
@@ -36,272 +37,22 @@
 #include "testing_gemm_strided_batched_ex.hpp"
 #endif
 
-namespace po = boost::program_options;
-
-int main(int argc, char* argv[])
+static int run_bench_test(const char* function, char precision, Arguments argus)
 {
-    Arguments argus;
```
leekillough: These diffs are hard to follow. Basically, the running of a performance test has been put into a function by itself, which can either be called from the command-line parser, or from a file reader which reads the Arguments data. strcmp() is used instead of std::string because we want to be able to remove testing_ prefix simply by moving up a pointer, and because it's generally faster.

clients/benchmarks/client.cpp
```diff
@@ -36,272 +37,22 @@
...
-         "Validate GPU results with CPU? 0 = No, 1 = Yes (default: No)")
-
-        ("iters,i",
-         po::value<rocblas_int>(&argus.iters)->default_value(10),
-         "Iterations to run inside timing loop")
-
-
-
-
-        ("algo",
```
leekillough: These 4 values were added since the refactoring started, so they are merged here.

clients/benchmarks/client.cpp
```diff
@@ -36,272 +37,22 @@
...
-         po::value<uint32_t>(&argus.algo)->default_value(0),
-         "extended precision gemm algorithm")
-        
-        ("solution_index",
-         po::value<int32_t>(&argus.solution_index)->default_value(0),
```
leekillough: solution_index used to be uint32_t but is int32_t now.

clients/benchmarks/client.cpp
```diff
@@ -669,3 +419,286 @@ int main(int argc, char* argv[])
 
     return 0;
 }
+
```
leekillough: This function reads benchmarking binary data from a file, and runs the tests in sequence.

clients/benchmarks/dgemm_genperf.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python
```
leekillough: This generates a representative set of DGEMM tests combinatorially, generating a YAML file. Duplicates are removed. The YAML file is > 200 MB, so it's preferable that it not be checked in but rather generated on demand.

clients/benchmarks/perf.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
leekillough: Eventually when more performance tests are added, they can be added here.

clients/common/rocblas_gentest.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/python
```
leekillough: This is the main engine for the creation of test cases from YAML files. It parses YAML files, handles !include directives from one YAML file to another, and generates binary files corresponding to Arguments structure records. It is fairly complex Python code, but it is hopefully readable.

clients/common/utility.cpp
```diff
@@ -71,8 +71,15 @@ rocblas_int query_device_property()
         printf("Query device success: there are %d devices \n", device_count);
     }
 
-    for(rocblas_int i = 0; i < device_count; i++)
+    for(rocblas_int i = 0;; i++)
```
leekillough: The original code looked odd -------------------------
Like this.
----------------------------------------------

The new format is:

--------------------------------------------------
Device ID 0
...
--------------------------------------------------
Device ID 1
...
--------------------------------------------------

That is the reason for moving the loop exit into the middle of the loop, after the printing of the horizontal lines.

clients/common/utility.cpp
```diff
@@ -189,8 +192,8 @@ rocblas_operation char2rocblas_operation(char value)
     case 'n': return rocblas_operation_none;
     case 't': return rocblas_operation_transpose;
     case 'c': return rocblas_operation_conjugate_transpose;
+    default: return static_cast<rocblas_operation>(-1);
```
leekillough: Rather than silently falling back on defaults, a value of -1 (cast to the appropriate enum) is returned for an unknown value. This is used in other places (e.g. client.cpp) to detect illegal option values, rather than having to test it in two places.

clients/gtest/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ endif( )
 
 set(rocblas_test_source
     rocblas_gtest_main.cpp
+    ${Tensile_TEST_SRC}
```
leekillough: Tensile_TEST_SRC is moved to the front of the linker list, in order that the GEMM tests occur first during Google Tests runs. This allows quicker debugging of GEMM tests without having to specify fancy Google Test filters. It immediately runs GEMM tests, and will fail with error if anything is wrong with them.

clients/gtest/CMakeLists.txt
```diff
@@ -109,6 +109,16 @@ else( )
   target_link_libraries( rocblas-test PRIVATE ${HIPHCC_LOCATION} )
 endif( )
 
+set( ROCBLAS_TEST_DATA "${PROJECT_BINARY_DIR}/staging/rocblas_gtest.data")
```
leekillough: This generates the rocblas_gtest.data file from the gtest.yaml file. Once built, it doesn't need to be rebuilt unless gtest.yaml, one of its included yaml files, or the rocblas_gentest.py script is changed.

clients/gtest/gemm.yaml
```diff
@@ -0,0 +1,2552 @@
+# Definitions is a placeholder for defining YAML anchors to alias later.
```
leekillough: This is a combinatorial way of expressing test cases in a concise format.

clients/gtest/gemm_gtest.cpp
```diff
@@ -3,7135 +3,135 @@
  * ************************************************************************ */
```
leekillough: This file is practically rewritten, to be data-driven instead of having test cases statically compiled into C++.

clients/gtest/gemm_gtest.cpp
```diff
@@ -3,7135 +3,135 @@
...
-gemm_tuple conv_ctest_bwddata_fp32_067 {{121, 512, 1, 121, 512, 121}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_068 {{121, 75, 1, 121, 75, 121}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_069 {{121, 832, 1, 121, 832, 121}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_070 {{12321, 147, 1, 12321, 147, 12321}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_071 {{12321, 27, 1, 12321, 27, 12321}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_072 {{12321, 363, 1, 12321, 363, 12321}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_073 {{12321, 75, 1, 12321, 75, 12321}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_074 {{12544, 147, 1, 12544, 147, 12544}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_075 {{12544, 1600, 1, 12544, 1600, 12544}, {1, 0}, {'N', 'T'}};
-gemm_tuple conv_ctest_bwddata_fp32_076 {{12544, 27, 1, 12544, 27, 12544}, {1, 0}, {'N', 'T'}};
```
leekillough: This is a lambda expression which filters data entries for their relevance to testing_gemm.

leekillough: This formats the trailing portion of the testcase name, by using transA, transB, M, N, K, lda, ldb, and ldc. In case there are duplicates, a final suffix increments, to distinguish test cases.

leekillough: Each category (quick, pre_checkin, nightly, known_bug) is instantiated separately, with a filter to select the current category, so that the test name prefix contains the category name.

clients/gtest/gtest.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
leekillough: This is the master test YAML file, which can be expanded later to include any number of tests.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -12,6 +12,8 @@
 
 int main(int argc, char** argv)
 {
+    // Open data file
```
leekillough: In the rocblas_gtest_main main() program, we initialize the RocBLAS_TestData class instantiation with the data file name.

clients/include/rocblas.yaml
```diff
@@ -0,0 +1,97 @@
+# Data types are defined as either aliases to Python-recognized ctypes,
```
leekillough: See embedded YAML comments.

clients/include/rocblas_test_unique_ptr.hpp
```diff
@@ -1,52 +1,52 @@
-#ifndef GUARD_ROCBLAS_MANAGE_PTR
```
leekillough: The changes in this file are cosmetic: Removal of tab characters and trailing whitespace, and the use of the "do {} while (0)" idiom to make function-like macros syntactically equivalent to function calls.

clients/include/testing_gemm.hpp
```diff
@@ -73,9 +74,9 @@ void testing_gemm_NaN(Arguments argus)
                                          rocblas_test::device_free};
     auto dC_managed = rocblas_unique_ptr{rocblas_test::device_malloc(sizeof(T) * size_C),
                                          rocblas_test::device_free};
-    T* dA = (T*)dA_managed.get();
```
leekillough: static_cast or reinterpret_cast is preferable to C-style casts.

clients/include/testing_gemm.hpp
```diff
@@ -196,15 +197,15 @@ void testing_gemm_bad_arg()
         status = rocblas_gemm<T>(
             handle, transA, transB, M, N, K, alpha_null, dA, lda, dB, ldb, &beta, dC, ldc);
 
-        verify_rocblas_status_invalid_pointer(status, "ERROR: C is nullptr");
+        verify_rocblas_status_invalid_pointer(status, "ERROR: alpha is nullptr");
```
leekillough: Copy-and-paste error

clients/include/testing_gemm.hpp
```diff
@@ -387,7 +388,10 @@ rocblas_status testing_gemm(Arguments argus)
         CHECK_HIP_ERROR(hipMemcpy(hC_2.data(), dC, sizeof(T) * size_C, hipMemcpyDeviceToHost));
 
         // CPU BLAS
-        cpu_time_used = get_time_us();
+        if(argus.timing)
```
leekillough: We do not need to collect timing info unless argus.timing is true. Previously the code unconditionally collected timing info, even if it didn't use it.

clients/include/testing_gemm.hpp
```diff
@@ -427,13 +434,9 @@ rocblas_status testing_gemm(Arguments argus)
         if(argus.norm_check)
         {
             double error_hst_ptr =
-                norm_check_general<T>('F', M, N, ldc, hC_gold.data(), hC_1.data());
+                fabs(norm_check_general<T>('F', M, N, ldc, hC_gold.data(), hC_1.data()));
```
leekillough: In this case, fabs() is clearer than using conditional expressions.

clients/include/utility.h
```diff
@@ -525,63 +530,97 @@ class Arguments
...
+
+enum rocblas_data_class
+{
+    rocblas_test_data,
+    rocblas_perf_data,
+};
+
+// Class used to read Arguments data into the tests
+template <rocblas_data_class>
+struct RocBLAS_Data
```
leekillough: This is the Data file iterator class. It contains a singleton which is initialized with a filename on the first call.

The data file is read an arbitrary number of times, by calling the begin() method and iterating until it reaches end(), then calling begin() again, it resets to the beginning of the file. This allows an arbitrary number of passes through the data file. The begin() method allows an option filter functor which returns true when a datum is to be admissible.

clients/include/utility.h
```diff
@@ -525,63 +530,97 @@ class Arguments
...
+        {
+            cerr << "Cannot open " << file << ": " << strerror(errno) << endl;
+            throw ifstream::failure("Cannot open " + file);
+        }
+    }
+
+    ifstream ifs;
 };
 
+typedef RocBLAS_Data<rocblas_test_data> RocBLAS_TestData;
```
leekillough: Since the RocBLAS_Data class template is static and only consists of single instances, we have to instantiate it once for the correctness tests and once for the performance tests.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -66,10 +66,10 @@ void device_strided_batched_matrix_copy(const void* src,
             size_t matrix_size = n1 * n2 * elem_size;
```
leekillough: This file's changes only consist of removing trailing whitespace and fixing typos

library/src/include/rocblas_unique_ptr.hpp
```diff
@@ -1,22 +1,22 @@
-#ifndef GUARD_ROCBLAS_MANAGE_PTR_HPP
```
leekillough: This file's changes consist only of whitespace (replacing tabs with spaces).

clients/include/testing_gemm_sweep.hpp
```diff
@@ -1,237 +0,0 @@
-/* ************************************************************************
```
leekillough: testing_gemm_sweep was redundant. It was basically the same as testing_gemm except that it removed timing code, and made incorrect error calculations. It should be removed.

.dir-locals.el
amcamd: This is fine if it helps emacs users and does not harm any other user

clients/include/utility.h
```diff
@@ -6,87 +6,101 @@
...
+                    __LINE__);                    \
+            exit(EXIT_FAILURE);                   \
         }                                         \
-        cout << endl;                             \
-    }
+    while(0)
+
+#define CHECK_ROCBLAS_ERROR(error)                                  \
+    do                                                              \
+        if(error != rocblas_status_success)                         \
```
amcamd: I am not sure if you need to add extra {} for the do{} while(0)

clients/include/utility.h
```diff
@@ -6,87 +6,101 @@
...
+                fprintf(stderr, "rocblas_status error");            \
+            }                                                       \
+            fprintf(stderr, "\n");                                  \
+            return error;                                           \
+        }                                                           \
+    while(0)
+
+#define BLAS_1_RESULT_PRINT                           \
+    do                                                \
+        if(argus.timing)                              \
```
amcamd: also here, maybe extra {} for the do {}  while(0)

.dir-locals.el
leekillough: Yes, it only affects emacs users. Any .dir-locals.el in a directory is loaded whenever emacs edits a file in or below that directory.

clients/benchmarks/client.cpp
zaliu: But you're replacing <stdio.h> with <cstring> as opposed to <cstdio>.  Is that what you intended?

leekillough: I will add cstdio, since printf is used below, even though it gets pulled in from other headers.

clients/include/utility.h
leekillough: I added { } around it for consistency.

PR #403, opened 2018-11-06
Threaded test for Tensile locking code

sdquiring:
Right now this seems to expose a bug in rocm/hip.  Adding this PR to see what happens on the CI machines.


sdquiring:
This should be ready to merge assuming it passes the CI tests.

sdquiring:
I should mention that I still haven't reproduced a failure in the Tensile Map code, but that I'm moving on to other priorities.

PR #406, opened 2018-11-07
Adding Int8 GEMM support

yoichiyoshida:
resolves #___

Summary of proposed changes:
-  Added int8 support to API
-  Added small int8 tests
-  Tests that are not all 1s currently do not pass.


clients/common/cblas_interface.cpp
```diff
@@ -724,6 +724,70 @@ void cblas_gemm<rocblas_double_complex>(rocblas_operation transA,
...
+
+    int const sizeA = ((transA == rocblas_operation_none) ? k * lda : m * lda);
+    int const sizeB = ((transB == rocblas_operation_none) ? n * ldb : k * ldb);
+    int const sizeC = n * ldc;
+
+    std::unique_ptr<double[]> A_double(new double[sizeA]());
+    std::unique_ptr<double[]> B_double(new double[sizeB]());
+    std::unique_ptr<double[]> C_double(new double[sizeC]());
+
+    for(int i = 0; i < sizeA; i++)
```
amcamd: I think this will only work if transA == T. If transA == N you need tripple loop, one for m, two for k. The two for k account for a unit stride between 4 consecutive values, and a lda stride between non-consecutive values, something like

for(i = 1; i < m; i++){
for(j1 = 0; j1 < 4; j1++){
for(j2 = 0; j2 < k; j2+=4){
A_double[i+(j2+j1)*lda) = static_cast<double>(A[i+j1+j2*lda]);
}}}

Please check the above, I am writing it down quickly.

amcamd: The multiplication asterisk before lda in above equation seems to have been lost in formatting, it should be

i+(j2+j1) multiplied by lda

clients/common/cblas_interface.cpp
```diff
@@ -724,6 +724,70 @@ void cblas_gemm<rocblas_double_complex>(rocblas_operation transA,
...
+    {
+        A_double[i] = static_cast<double>(A[i]);
+    }
+    for(int i = 0; i < sizeB; i++)
```
amcamd: I think this is same as above, but it will work for transB = N (where k is unit-stride)  not for transB = T (where k has stride lda)

clients/common/cblas_interface.cpp
amcamd: also missing   j2 multiplied by lda

PR #413, opened 2018-11-09
Search for rocblas_gtest.data in the same directory as rocblas-test

leekillough:
Search for rocblas_gtest.data in the same directory as rocblas-test

### DIFF
--- a/clients/gtest/rocblas_gtest_main.cpp
+++ b/clients/gtest/rocblas_gtest_main.cpp
@@ -6,6 +6,26 @@
 #include <stdexcept>
 #include "utility.h"
 
+#define GTEST_DATA "rocblas_gtest.data"
+
+// Return path of this executable
+static string exepath()
+{
+    string pathstr;
+    char* path = realpath("/proc/self/exe", 0);
+    if(path)
+    {
+        char* p = rindex(path, '/');
+        if(p)
+        {
+            p[1]    = 0;
+            pathstr = path;
+        }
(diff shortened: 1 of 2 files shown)


leekillough:
The error message seems fairly clear, that it cannot open rocblas_gtest.data.

This is not a common user error, but more like an internal error which should never happen.

The rocblas_gtest.data file and rocblas-test executable are closely tied together. rocblas-test cannot work without rocbas_gtest.data.

The rocblas-test user does not need to specify the rocblas_gtest.data file, because they didn't have to specify it before refactoring, and the normal Google Tests workflow does not involve supplying data files on the command-line or elsewhere, but simply running an executable.

The -h message is handled by Google Tests.

The opening of the data file needs to occur before the first Google Test is created, which happens at the same time as Google Tests parses the command-line. 

We could defer the opening of the data file until the first time it's actually needed, with a little more logic.

leekillough:
Another change will be posted soon, which defers the opening of the file until it's absolutely required, which allows the Google Tests CLI parsing and other stuff to finish first (so that rocblas-test -h etc. can be printed).

gilbertlee-amd:
Looks good

leekillough:
I verified, by putting a fake filename in the rocblas_gtest_main.cpp file, that rocblas-test -h prints the help message before aborting with the bad filename. (I don't know why Google Tests doesn't immediately exit after printing -h help output.)

PR #415, opened 2018-11-12
Refactoring of rocBLAS tests

leekillough:
- Refactor host and device memory allocation in terms of `host_vector<>` and `device_vector<>` classes
  - Greatly simplified syntax -- much simpler than overly-complicated `unique_ptr` usage.
  - Automatic allocation and deallocation of memory using RAII principles
  - No need to use `.get()` or `.data()` to access address of data; automatic decay of object into pointer to data
  - `device_vector<>` has memory guards to detect out of bounds memory corruption during testing
- Add `rocblas_local_handle` class as an automatically created and destroyed `rocblas_handle`
- Explicitly specify both input and output data types in `cblas_gemm<>()` calls, to avoid ambiguity
- Use `rocblas_int` type instead of `int` type in some size calculations and loop variables, for consistency in case `rocblas_int` >` int`
- Use C++11 random number generator
- Use static_cast<>() instead of C-style casts in some places
- Remove unnecessary casts of pointer arguments in calls, e.g. `(const T*)&alpha`
- Remove spurious `return` left in `testing_scal.hpp`
- Add `alpha == nullptr` tests to GEMV


clients/common/norm.cpp
```diff
@@ -66,16 +66,15 @@ double norm_check_general<rocblas_half>(char norm_type,
                                         rocblas_half* hCPU,
                                         rocblas_half* hGPU)
 {
-    // norm type can be O', 'I', 'F', 'o', 'i', 'f' for one, infinity or Frobenius norm
+    // norm type can be 'O', 'I', 'F', 'o', 'i', 'f' for one, infinity or Frobenius norm
     // one norm is max column sum
     // infinity norm is max row sum
     // Frobenius is l2 norm of matrix entries
 
     double error_double = std::numeric_limits<double>::quiet_NaN();
```
leekillough: host_vector<> is a much simpler mechanism than unique_ptr. There is no need to use unique_ptr simple to free data when a variable goes out of scope. Any class with a destructor is sufficient.

clients/common/norm.cpp
```diff
@@ -86,10 +85,10 @@ double norm_check_general<rocblas_half>(char norm_type,
     float alpha      = -1.0f;
     rocblas_int size = lda * N;
```
leekillough: With `host_vector<>` we don't have to use `.get()` or `.data().` to get the address. `host_vector<T>` automatically decays into `T*`

clients/common/norm.cpp
```diff
@@ -226,13 +224,13 @@ double norm_check_general<rocblas_half>(char norm_type,
     // use triangle inequality ||a+b|| <= ||a|| + ||b|| to calculate upper limit for Frobenius norm
     // of strided batched matrix
 
-    std::unique_ptr<float[]> hCPU_float(new float[N * lda + (batch_count - 1) * stride_a]());
-    std::unique_ptr<float[]> hGPU_float(new float[N * lda + (batch_count - 1) * stride_a]());
-    for(int i_batch = 0; i_batch < batch_count; i_batch++)
```
leekillough: `rocblas_int` should be used instead of `int` in case `rocblas_int` is larger than `int`.

clients/common/unit.cpp
```diff
@@ -8,19 +8,6 @@
 #include "unit.h"
 #include "utility.h"
```
leekillough: This macro has been moved to `utility.h`

clients/gtest/gemm_ex_gtest.cpp
```diff
@@ -291,8 +291,7 @@ class parameterized_gemm_ex : public ::TestWithParam<gemm_ex_tuple>
 TEST(known_bugs_int8, IdentityMatrix)
 {
     // Simple test cases running on matrices full of ones
```
leekillough: `rocblas_local_handle` is like `rocblas_handle` except that it is automatically created and destroyed.

clients/gtest/gemm_ex_gtest.cpp
```diff
@@ -428,16 +416,22 @@ TEST(known_bugs_int8, IdentityMatrix)
...
 
             // Copy results back to host
-            hipMemcpy(hD.get(), dD, L * L * sizeof(int32_t), hipMemcpyDeviceToHost);
+            hipMemcpy(hD, dD, L * L * sizeof(int32_t), hipMemcpyDeviceToHost);
 
             bool isMatch = true;
             for(int i = 0; i < L; i++)
             {
                 for(int j = 0; j < L; j++)
                 {
```
leekillough: Added short-circuiting

clients/include/testing_asum.hpp
```diff
@@ -103,35 +88,31 @@ rocblas_status testing_asum(Arguments argus)
         CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_device));
         status = rocblas_asum<T1, T2>(handle, N, dx, incx, d_rocblas_result_2);
 
-        asum_arg_check(status, d_rocblas_result_2);
```
leekillough: We have to add <T2> explicit template argument, to force the second argument to T2* type.

clients/include/testing_axpy.hpp
```diff
@@ -139,18 +126,10 @@ rocblas_status testing_axpy(Arguments argus)
     hy_gold = hy_1;
 
     // allocate memory on device
```
leekillough: Another example of simplification.

clients/include/testing_geam.hpp
```diff
@@ -241,30 +226,20 @@ rocblas_status testing_geam(Arguments argus)
     hB_copy = hB;
 
     // allocate memory on device
```
leekillough: Simplify!!!

clients/include/testing_gemm.hpp
```diff
@@ -297,33 +275,23 @@ rocblas_status testing_gemm(Arguments const& argus)
     const auto size_C = static_cast<size_t>(ldc) * static_cast<size_t>(N);
 
     // allocate memory on device
```
leekillough: Simplify!!!

clients/include/testing_gemm_ex.hpp
```diff
@@ -444,37 +425,25 @@ rocblas_status testing_gemm_ex_template(rocblas_operation transA,
     const size_t size_D = static_cast<size_t>(ldd) * static_cast<size_t>(N);
 
     // allocate memory on device
```
leekillough: Super Simplify!!!

clients/include/testing_gemm_kernel_name.hpp
```diff
@@ -54,7 +51,7 @@ rocblas_status testing_gemm_kernel_name(Arguments argus)
     rocblas_int bsb = ldb * B_col * 2;
     rocblas_int bsc = ldc * N * 2;
```
leekillough: These pointers aren't used, but I still thought they should be initialized.

clients/include/testing_gemm_strided_batched.hpp
```diff
@@ -132,21 +124,11 @@ rocblas_status testing_gemm_strided_batched(Arguments argus)
         size_one_c + static_cast<size_t>(stride_c) * static_cast<size_t>(batch_count - 1);
 
     // allocate memory on device
```
leekillough: Simplify!!!

clients/include/testing_gemm_strided_batched.hpp
```diff
@@ -229,25 +211,25 @@ rocblas_status testing_gemm_strided_batched(Arguments argus)
                                                             stride_c,
                                                             batch_count));
 
-        CHECK_HIP_ERROR(hipMemcpy(hC_2.data(), dC, sizeof(T) * size_c, hipMemcpyDeviceToHost));
+        CHECK_HIP_ERROR(hipMemcpy(hC_2, dC, sizeof(T) * size_c, hipMemcpyDeviceToHost));
 
         // CPU BLAS
         cpu_time_used = get_time_us();
         for(rocblas_int i = 0; i < batch_count; i++)
         {
```
leekillough: No need for `.data()`. But must specify both template arguments `cblas_gemm<T,T>` so that all function arguments are coerced correctly.

clients/include/testing_gemm_strided_batched_ex.hpp
```diff
@@ -517,38 +493,26 @@ rocblas_status testing_gemm_strided_batched_ex_template(rocblas_operation transA
     }
 
     // allocate memory on device
```
leekillough: Super-Simplify

clients/include/testing_gemm_strided_batched_ex.hpp
```diff
@@ -756,23 +723,24 @@ rocblas_status testing_gemm_strided_batched_ex_template(rocblas_operation transA
 
         for(rocblas_int i = 0; i < batch_count; i++)
         {
```
leekillough: No need for `.data()`. But both template arguments specified for `cblas_gemm<Td,Td>` so that all function arguments are cast correctly.

clients/include/testing_gemm_strided_batched_ex.hpp
```diff
@@ -756,23 +723,24 @@ rocblas_status testing_gemm_strided_batched_ex_template(rocblas_operation transA
...
+                               hA + stride_a * i,
+                               lda,
+                               hB + stride_b * i,
+                               ldb,
+                               h_beta_Td,
+                               hD_gold + stride_d * i,
+                               ldd);
         }
 
         cpu_time_used = get_time_us() - cpu_time_used;
```
leekillough: Someone forgot `batch_count` in calculation.

clients/include/testing_gemv.hpp
```diff
@@ -180,31 +163,18 @@ rocblas_status testing_gemv(Arguments argus)
     size_y = dim_y * abs_incy;
 
     // Naming: dK is in GPU (device) memory. hK is in CPU (host) memory
```
leekillough: Super-Simplify.

clients/include/testing_iamax.hpp
```diff
@@ -82,19 +77,13 @@ rocblas_status testing_iamax(Arguments argus)
...
 
     // check to prevent undefined memory allocation error
     if(N <= 0 || incx <= 0)
     {
-        auto dx_managed = rocblas_unique_ptr{rocblas_test::device_malloc(sizeof(T) * safe_size),
-                                             rocblas_test::device_free};
-        auto d_rocblas_result_managed = rocblas_unique_ptr{
-            rocblas_test::device_malloc(sizeof(rocblas_int)), rocblas_test::device_free};
-        T* dx                         = (T*)dx_managed.get();
-        rocblas_int* d_rocblas_result = (rocblas_int*)d_rocblas_result_managed.get();
```
leekillough: `d_rocblas_result` is never used in this block, and hence is removed.

clients/include/testing_iamin.hpp
```diff
@@ -81,20 +76,13 @@ rocblas_status testing_iamin(Arguments argus)
...
+    rocblas_local_handle handle;
 
     // check to prevent undefined memory allocation error
     if(N <= 0 || incx <= 0)
     {
-        auto dx_managed = rocblas_unique_ptr{rocblas_test::device_malloc(sizeof(T) * safe_size),
-                                             rocblas_test::device_free};
-        auto d_rocblas_result_managed = rocblas_unique_ptr{
-            rocblas_test::device_malloc(sizeof(rocblas_int)), rocblas_test::device_free};
-        T* dx                         = (T*)dx_managed.get();
```
leekillough: `d_rocblas_result` is never used in this block, and hence is removed.

clients/include/testing_logging.hpp
```diff
@@ -140,24 +140,12 @@ void testing_logging()
     rocblas_int size_d   = (ldd > stride_d ? ldd : stride_d) * safe_dim * batch_count;
 
     // allocate memory on device
```
leekillough: Super-simplify

clients/include/utility.h
```diff
@@ -48,6 +56,20 @@ typedef rocblas_half half;
         }                                         \
     } while(0)
```
leekillough: This macro was moved from `rocblas_test_uniq_ptr.hpp`

clients/include/testing_scal.hpp
```diff
@@ -28,17 +28,12 @@ void testing_scal_bad_arg()
...
-    rocblas_handle handle = unique_ptr_handle->handle;
+    rocblas_local_handle handle;
 
     rocblas_int size_x = N * incx;
 
     // allocate memory on device
-    auto dx_managed = rocblas_unique_ptr{rocblas_test::device_malloc(sizeof(T) * size_x),
-                                         rocblas_test::device_free};
-    return;
-
```
leekillough: I'm not sure why this return is here. I'm removing it assuming it was accidentally checked in after someone was hand-debugging code and put it in there as an early return to test what was causing a failure.

clients/include/testing_gemm_kernel_name.hpp
```diff
@@ -54,7 +51,7 @@ rocblas_status testing_gemm_kernel_name(Arguments argus)
...
 
-    T *dA, *dB, *dC;
+    T *dA = nullptr, *dB = nullptr, *dC = nullptr;
 
     return rocblas_gemm_kernel_name<T>(handle,
```
leekillough: These pointers aren't used, but I still thought they should be initialized.

clients/include/testing_gemm_strided_batched_ex.hpp
```diff
@@ -756,23 +723,24 @@ rocblas_status testing_gemm_strided_batched_ex_template(rocblas_operation transA
...
-        cblas_gflops  = gemm_gflop_count<Td>(M, N, K) / cpu_time_used * 1e6;
+        cblas_gflops  = gemm_gflop_count<Td>(M, N, K) * batch_count / cpu_time_used * 1e6;
+
```
leekillough: Someone forgot `batch_count` in calculation.

clients/include/testing_gemv.hpp
```diff
@@ -31,82 +30,81 @@ void testing_gemv_bad_arg()
...
 
         verify_rocblas_status_invalid_pointer(status, "rocBLAS TEST ERROR: y is null pointer");
     }
+    {
+        T* alpha_null = nullptr;
+        status =
+            rocblas_gemv<T>(handle, transA, M, N, alpha_null, dA, lda, dx, incx, &beta, dy, incy);
+
+        verify_rocblas_status_invalid_pointer(status, "rocBLAS TEST ERROR: alpha is null pointer");
+    }
```
leekillough: alpha == NULL test was omitted from the GEMV tests, and from the GEMV code itself. This adds it to the tests (an earlier PR added it to the library).

clients/gtest/gemm_ex_gtest.cpp
```diff
@@ -428,16 +416,22 @@ TEST(known_bugs_int8, IdentityMatrix)
...
-                    isMatch &= (hD[j * L + i] == j * L + i);
+                    if(hD[j * L + i] != j * L + i)
+                    {
+                        isMatch = false;
+                        goto end;
+                    }
```
leekillough: Added short-circuiting

clients/common/norm.cpp
```diff
@@ -66,16 +66,15 @@ double norm_check_general<rocblas_half>(char norm_type,
...
     // Frobenius is l2 norm of matrix entries
 
     double error_double = std::numeric_limits<double>::quiet_NaN();
 
-    std::unique_ptr<float[]> hCPU_float(new float[N * lda]());
-    std::unique_ptr<float[]> hGPU_float(new float[N * lda]());
-    for(int i = 0; i < N * lda; i++)
+    host_vector<float> hCPU_float(N * lda), hGPU_float(N * lda);
+    for(rocblas_int i = 0; i < N * lda; i++)
     {
```
leekillough: `rocblas_int` should be used instead of `int` in case `rocblas_int` is larger than `int`.

clients/include/utility.h
```diff
@@ -123,54 +145,262 @@ inline float half_to_float(rocblas_half val)
...
+    {
+        static_assert(sizeof(UINT_T) == sizeof(U), "Type sizes do not match");
+        union
+        {
+            UINT_T u;
+            U fp;
+        } x;
+        do
+            x.u = uniform_int_distribution<UINT_T>()(rocblas_rng);
+        while(!(x.u & (((UINT_T)1 << SIG) - 1))); // Reject Inf (mantissa == 0)
```
amcamd: If this is slowing down the code could you create a     pad_mask  in random_data . There would be a separate pad_mask for half, float, double, and you could  replace the two lines with
x.u  |=  pad_mask;
The pad_mask would set the exponent to all ones and set the lowest mantissa bit to 1

leekillough: It shouldn't be slowing down the code; it just does one `test` instruction (the compiler optimizes the constant mask expression) and a branch if the `(value & mask) == 0`. There's a `1/2^SIG` probability that it repeats. Repeating it ensures that all legal values are equally likely. Setting the lowest mantissa `== 1` all the time, would throw away half of the possible random values. There is no need for a separate `pad_mask` because the compiler optimizes constant expressions, and `SIG` and `EXP` are constants. This is the assembly code for the test and mask above:

```
        movabs  rbp, 4503599627370495        ; 2^SIG-1

.L31:
...     
     ;  rax = uniform_int_distribution<UINT_T>()(rocblas_rng)

        test    rax, rbp   ; Tests whether mantissa == 0
        je      .L31       ; Only repeats if mantissa == 0; 1/2^SIG probability
        movabs  rbx, 9218868437227405312   ; (2^EXP-1)<<SIG
        or      rax, rbx   ; Sets all exponent bits to 1
...
 ```

It tests whether `%rax` ANDed with `2^SIG-1` is zero, and if it is, gets another random number (only happens with `1/2^SIG` chance). It then ORs to set the exponent to all 1's.

The expected number of iterations is `1/(1 - 2^-SIG)`, which is:
```
double: 4503599627370496 / 4503599627370495 = 1.0000000000000002...
float:  8388608 / 8388607                   = 1.000000112...
half:   1024 / 1023                         = 1.001...
```

In practice, the `TEST` and `JE` instructions take 0 cycles, because they overlap other instructions which execute out-of-order.

PR #416, opened 2018-11-12
Adding larger int8 test cases and removing custom int8 tests

yoichiyoshida:
resolves #___

Summary of proposed changes:
-  Removes ones and identity tests for int8
-  Add medium and large cases for int8
-  Add deepbench and resnet50 cases for int8


leekillough:
PR415's changes to gemm_ex_gtest.cpp can be ignored if you're deleting the functions.

You can squash the commits into one when merging into github, so that your merge of my PR doesn't show up in the history.

I'll wait for your code to be checked in before refactoring gemm_ex_gtest.cpp. I'll convert the test cases to YAML and then rewrite the Google Test code.

bragadeesh:
looks like this is passing jenkins, can we merge ?

clients/gtest/gemm_ex_gtest.cpp
```diff
@@ -146,6 +146,203 @@ const vector<vector<int>> NaN_matrix_size_range = {
...
+    { 3136,    64,   576,  3136,   576,  3136,  3136},
+    {   49,   512,  4608,    49,  4608,    49,    49},
+    {50176,   128,   256, 50176,   256, 50176, 50176},
+    {50176,   512,   256, 50176,   256, 50176, 50176},
+    {  784,   128,  1152,   784,  1152,   784,   784}
+};
+
+// vector of vector, each vector is a {M, N, K, lda, ldb, ldc, ldd};
+const vector<vector<int>> resnet50_bwdwrw_size_range = {
+    { 1024,  2048,    49,    49,    49,  1024,  1024},
```
amcamd: I forget, does this K need to be a multiple of 4? Is the above value of 49 skipped, or is K allowed to not be a multiple of 4?

gilbertlee-amd: Currently the test will return invalid (which passes silently).  This was something I meant to ask about but forgot to.

bragadeesh: @yoichiyoshida  does the test check for error code for non-multiple 4? can you point me to line number

bragadeesh: are all the *int8* test cases passing on your side?

gilbertlee-amd: The multiple of 4 check occurs here:

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/blas_ex/rocblas_gemm_ex.cpp#L441

amcamd: OK, I assume it is returning    rocblas_status_not_implemented

gilbertlee-amd: It returns rocblas_status_invalid_size.  However I think these silently pass testing.  Was that intentional or a bug?

yoichiyoshida: The call to invalid k sizes currently returns rocblas_status_invalid_size. There's a switch statement in the generic test class that checks various cases, but not this one (because it is int8 specific). However the test currently just passes for cases not checked, so right now it currently passes even though we don't check the return code for this case. For the refactor, we can either remove the invalid cases, or explicitly check for the return code.

amcamd: Thanks, for a refactor, it will be be best to leave the test case and explicitly check the return code.

bragadeesh: yes, infact there should be category of tests that main purpose checks for error code returning sizes and make sure they return the appropriate codes; all normal cases should fail in gtest run if returning error code

PR #422, opened 2018-11-21
Fix packInt8 index overflow from PR420

yoichiyoshida:
Fix packInt8 index overflow introduced by changes made during review of PR420.

### DIFF
--- a/clients/include/utility.h
+++ b/clients/include/utility.h
@@ -685,12 +685,11 @@ inline void rocblas_packInt8(host_vector<T>& A, rocblas_int M, rocblas_int N, ro
     }
 
     host_vector<T> temp(A);
-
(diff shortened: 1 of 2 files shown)


clients/include/utility.h
```diff
@@ -685,12 +685,11 @@ inline void rocblas_packInt8(host_vector<T>& A, rocblas_int M, rocblas_int N, ro
     }
 
     host_vector<T> temp(A);
-
```
gilbertlee-amd: These changes won't help because inputs are still only rocblas_int?

gilbertlee-amd: Nevermind - the index gets promoted to size_t as a result.

gilbertlee-amd: just switch back to the original version if we aren't multi-threading this anyways.

PR #423, opened 2018-11-24
Remerging trtri speed improvement with 2-stage recursion

yoichiyoshida:
Summary of proposed changes:
-  remerging lost performance change from 8b44a2d86da5ed16bcae8547272d659eb84c5b92
-  NOTE: trsm tests are currently not included in the CMake file for gtest, and are currently not run. Should they be?


bragadeesh:
I am merging this PR, but we need to get the trsm tests scrubbed

PR #425, opened 2018-11-27
Addition of L2 BLAS function TRSV 

mahmoodw:
#172729

Summary of proposed changes:
-  Added TRSV functionality by calling TRSM
-  Added tests for  TRSV


leekillough:
I will resolve merge conflicts and refactor the test code to use the new method.

bragadeesh:
need to squash all commits into 1

library/src/blas2/rocblas_trsv.cpp
```diff
@@ -0,0 +1,277 @@
...
+            pointer_mode == rocblas_pointer_mode_host ? &alpha_h : (T*)alpha_d.get(),
+            A,
+            lda,
+            x,
+            m);
+    }
+    else
+    {
+        auto dx_mod =
+            rocblas_unique_ptr{rocblas::device_malloc(sizeof(T) * m), rocblas::device_free};
```
bragadeesh: sorry, its the one here, need to see if it can be moved to program scope

PR #429, opened 2018-12-05
Complete refactoring of rocBLAS test framework

leekillough:
1. Total refactoring and cleanup of rocBLAS test framework
- Make Google Tests data-driven instead of hard-coded in C++ files
- Convert all existing test data into YAML files
- Make Google Test names depend on test arguments
- Make test code for rocBLAS functions as small as possible, abstracting out common features and distilling test code down to only the minimum required to define the test
- Make test code more easily expandable to new data types, without copying-and-pasting
- Handle dispatching functions by type in one place (`type_dispatch.hpp`), rather than in every individual test file
- Write documentation on how to add rocBLAS tests (`clients/gtest/README.md`)
- Remove unnecessary or duplicated code, using templates or macros to factor copy-and-paste code
- Replace simple functions with inlined code in headers (e.g. `flops.cpp`, `unit.cpp`, `near.cpp`, `arg_check.cpp` and `rocblas_template_specialization.cpp` are eliminated and replaced with simpler and smaller code in their respective headers)
- Split up large `utility.h` header into smaller functional headers, to reduce dependencies
- Simplify code, using C++14 where possible (e.g. `auto` return types, non-trivial `constexpr` functions, variable templates)
- Add `Arguments` pretty-printer, showing all test arguments when a test fails (previously only `rocblas_gemm` arguments were printed)
- Rename test filenames to be more consistent (e.g. `trsm_gtest.cpp` and `trsm_gtest.yaml` are 1-to-1; `rocblas_gtest.yaml` is main YAML test file, while `rocblas_gtest_main.cpp` is main test program)

2. Enhance `rocblas_gentest.py` script
- Allow YAML files to `include:` other YAML files (a YAML extension), taking combined text as one YAML stream
- Allow `function: precision` pairs to be listed within tests, all other parameter combinations being equal, to minimize duplication, e..g.:
```
    function:
      testing_gemm: *half_single_double_precisions
      testing_gemm_ex: *real_precisions
```
- Allow multiple YAML "documents", usually one for each test (allows tests to use YAML anchors/aliases private to their own YAML document declared between `---` and `...`)
- Print source file, line number and source snippet of YAML errors, for easier diagnosis of YAML syntax errors
- Allow known bugs to use wildcards for functions (e.g., `testing_gemm*` for `testing_gemm`, `testing_gemm_ex`, `testing_gemm_strided_batched`, `testing_gemm_strided_batched_ex`)

3. `iamax`/`iamin`
- Combine `iamax` and `iamin` tests, which differ only by max/min (previously only `iamax` test existed)
- Write CBLAS reference implementation for `iamin` (`iamin` is not a standard part of BLAS/CBLAS)
- Simplify `fetch_template.h` / `fetch_template.cpp` code only used in `nrm2`, `asum`, `iamax` and `iamin`, overloading functions for `|Re(x)|+|Im(x)|` and `Re(x)^2+Im(x)^2` for both real and complex types (very minimal risk)
- Start to fix `iamin` implementation (it was totally broken and just a copy-and-paste of `iamax` before, and wasn't tested at all)

4. Miscellaneous
- Enable C++14 for rocBLAS builds
- Use the same name everywhere for `Arguments` variables (`arg` instead of `argus`).
- Tweak `.dir-locals.el` to make emacs remove tabs and trailing whitespace for more text files


clients/benchmarks/client.cpp
```diff
@@ -28,19 +29,58 @@
...
+struct perf_gemm_ex<Ti, To, Tc, std::false_type> : rocblas_test_invalid
+{
+};
+
+// Template to dispatch testing_gemm_strided_batched_ex for performance tests
+// When Ti == void or complex, the test is marked invalid
+template <typename Ti,
+          typename To = Ti,
+          typename Tc = To,
+          typename    = typename std::conditional<!std::is_same<Ti, void>::value && !is_complex<Ti>,
```
pfultz2: If this is C++14, you can write this as `std::is_same<Ti, void>{}` instead of `std::is_same<Ti, void>::value`.

leekillough: **Actually, that's allowed in C++11 as well.**

The syntactic sugar of `std::is_same<Ti, void>{}` or `std::is_same<Ti, void>()` (C++11) or `std::is_same_v<Ti, void>` (C++17) versus `std::is_same<Ti, void>::value` is not overly compelling.

The use of `<type_traits>` (`std::integral_constant`) classes were not introduced by this PR, and they were used with `::value` in the old C++11 code.

clients/gtest/README.md
```diff
@@ -0,0 +1,295 @@
...
+```
+I. Call the `INSTANTIATE_TEST_CATEGORIES` macro which instantiates the Google Tests across all test categories (`quick`, `pre_checkin`, `nightly`, `known_bug`), passing it the same test name as in steps G and H. For example:
+```c++
+INSTANTIATE_TEST_CATEGORIES(gemm);
+```
+J. Don't forget to close the anonymous namespace:
+```c++
+} // namespace
+```
+**II.** Create a `<function>.yaml` file with the same name as the C++ file, just with
```
amcamd: Should this be    **III**     not    **II**      ?

PR #444, opened 2019-01-10
Add profile logging, refactor rocBLAS, and fix TRSV, IAMIN (net -1164 LoC)

leekillough:
-------------------------------------------------------------------------------

# rocBLAS functionality changes:

## Implement Profile Logging

Profile logging keeps count of the number of times each rocBLAS routine is called with specific arguments.

Profile logging is enabled with `ROCBLAS_LAYER=4`. Profile output file is controlled with `ROCBLAS_LOG_PROFILE_PATH`, and like trace and bench logging, defaults to `stderr`.

At the end of program execution, a YAML log file is output which lists each rocBLAS function call, its performance-important arguments, and the number of times it was called with those arguments:
```
{ rocblas_function: "rocblas_dgemv", transA: "C", M: 1, N: 1, lda: 0, incx: 1, incy: 0, call_count: 4 }
{ rocblas_function: "rocblas_dgemv", transA: "C", M: 1, N: 0, lda: 0, incx: 2, incy: 1, call_count: 4 }
{ rocblas_function: "rocblas_dgemv", transA: "N", M: 10, N: 10, lda: 0, incx: 1, incy: 0, call_count: 8 }
{ rocblas_function: "rocblas_dgemv", transA: "C", M: 1, N: 0, lda: 0, incx: 3, incy: -1, call_count: 4 }
{ rocblas_function: "rocblas_dgemv", transA: "T", M: 10, N: 10, lda: 0, incx: 1, incy: 0, call_count: 8 }
{ rocblas_function: "rocblas_dgemv", transA: "T", M: 0, N: 1, lda: 0, incx: 0, incy: -1, call_count: 4 }
{ rocblas_function: "rocblas_sgemv", transA: "C", M: 1, N: -1, lda: 0, incx: -1, incy: 2, call_count: 4 }
{ rocblas_function: "rocblas_dgemv", transA: "C", M: 10, N: 10, lda: 0, incx: 1, incy: 0, call_count: 8 }
```

**Note:** Performance-important arguments are arguments whose values can significantly change the performance of an algorithm, such as matrix and vector sizes, leading dimensions and increments, transposes, etc. Things such as the raw memory address of a matrix, are not considered performance-important, and are not logged by profile logging.

### Implementation details:

When profile logging is enabled, every rocBLAS function passes the names and values of its performance-important arguments to a logging function. The arguments are of the form `"name1", value1, "name2", value2, "name3", value3, ...,` and correspond to a YAML mapping / dictionary printed on the output.

The logging function uses a hash table to memoize the arguments, keeping count of how many times it has seen the same arguments.

The logging function uses C++11 tuples, and every distinct argument list type combination has a separate hash table indexed only by the argument values, not the argument names.

The logging function is thread-safe, and uses atomic counters to count the number of times a function is called with specific arguments. Shared read-write locking is used to minimize contention. To use STL containers, some form of locking is required, since modifying the map can invalidate iterators, and the map cannot be safely read while another thread is modifying it.

When the program terminates, the logger writes all of the distinct rocBLAS function calls, their arguments, and their call counts, to a file in YAML format.

When all types of logging (trace, bench, profile) are disabled, the new implementation has as little overhead as possible, by not generating things such as string representations of kernel arguments, unless at least one form of logging is enabled. If logging is disabled, then branch prediction should make
the overhead of logging insignificant.

-------------------------------------------------------------------------------

# rocBLAS quality of code changes (in no particular order of preference):

## Refactor HIP kernel templates slightly to avoid duplicated code for host scalars and device scalars, by using an extra template parameter and a special inlined function called `load_scalar()`.

**Details:** The scalar arguments (alpha, beta, etc.) in a kernel can have their type templatized, such that either a pointer to a scalar in device memory, or an actual scalar value loaded from host memory, can be passed to the same HIP kernel function in the same way. Then, an inlined function called `load_scalar()` an either load the scalar value if it's a pointer to device memory, or it can simply copy the scalar argument if it's not a pointer.
```
// load_scalar function:
template <typename T>
__attribute__((always_inline)) __host__ __device__ T load_scalar(T x)
{
    return x; // copy scalar passed to function
}

template <typename T>
__attribute__((always_inline)) __host__ __device__ T load_scalar(const T* x)
{
    return *x; // load scalar from device memory
}

// rocBLAS kernel function:
template <typename T, typename U>
__global__ void my_kernel(const T* x, U alpha_device_host, T* y)
{
    // U can either be a pointer or scalar type
    // alpha can come from either device or host memory
    //
    auto alpha = load_scalar(alpha_device_host);
    // ...
}
```
**Also:** See Appendix.

The `load_scalar()` function is marked `always_inline`, and will produce an error if the compiler cannot inline it. It works on both host and device, but since it is a template, it won't generate any code or take up any space until it is actually used somewhere.

**Rationale:** Avoid duplicating HIP kernels, or having to write special wrappers for device and host scalars. The same kernel template can be used for both device and host scalars, and it avoids code duplication. There is no performance penalty. There is just an extra template parameter for the type of the scalar(s), and an inlined function call to copy or load the scalar(s). The extra scalar template parameter is deduced automatically based on the kernel scalar arguments, so it does not have to be specified explicitly.

**Note:** For functions which use `rocblas_half2`, the `load_scalar()` function can be overloaded to automatically load two half-precision values into a single 32-bit scalar, without requiring changes in the rest of the kernel code. This simplifies AXPY. See Appendix.

**Affects:** axpy, scal, gemv, ger, syr

-------------------------------------------------------------------------------

## Move incx / incy adjustments outside of the GPU kernel.

**Details:** Currently, the code checks whether incx >= 0 and incy >= 0 inside of the computational kernel, and uses different address calculations based on whether each of incx, incy is positive or negative. For example, a dot product kernel is written like this:
```
    if (incx >= 0 && incy >= 0)
    {
        dot += x[tid * incx] * y[tid * incy];
    }
    else if (incx < 0 && incy >= 0)
    {
        dot += x[tid * incx - (n-1) * incx] * y[tid * incy];
    }
    else if (incx < 0)
    {
        dot += x[tid * incx - (n-1) * incx] * y[tid * incy - (n-1) * incy];
    }
    else
    {
        dot += x[tid * incx] * y[tid * incy - (n-1) * incy];
    }
```
When `incx` and `incy` are BLAS parameters, they represent strides of vectors. When the strides are negative, the calculations are supposed start at the ends of the vectors, and move backwards. This can be done by adjusting the starting addresses of the vectors to the end of the vectors when the increments
are negative. Then, the same code is used for the negative increments as is used for the positive increments, just with different starting addresses. The adjustments of the x and y vectors' starting addresses is done on the host, before the GPU kernel is started. Even though the host cannot directly access GPU device memory, it can perform adjustments on the device array addresses.
```
// Adjustment in host before the kernel starts:

if (incx < 0)
    x -= ssize_t(incx) * (n - 1);
if (incy < 0)
    y -= ssize_t(incy) * (n - 1);
```
Simplified kernel without conditions:
```
    dot += x[tid * incx] * y[tid * incy];          // No conditional code
```
**Also:** See Appendix.

**Rationale:** Reduce the kernel code size by 50-75%, and avoid conditional code inside of kernels, improving the chances of optimization.

**Affects:** axpy, dot, copy, swap, gemv

-------------------------------------------------------------------------------

## Combine `iamax` and `iamin` code. They differ only by their comparison of maximum and minimum values. Make `iamin` work correctly for the first time, and enable its unit testing.

**Rationale:** Avoid duplicate code, by factoring out the only differences between `iamax` and `iamin`. The same algorithm can be used for both, since they are fundamentally the same type of reduction, just with a different comparison function.

**Note:** `iamin` is not a part of standard BLAS/CBLAS. It is a common extension, though. **It has, until now, been implemented incorrectly in rocBLAS.**

**Affects:** iamax, iamin

-------------------------------------------------------------------------------

## Change the logging control from non-static `rocblas_handle` member variables to static `rocblas_handle` member variables.

**Rationale:** This allows logging to be done separately from the lifetime of handle variables, such as at the end of program execution.

The logging is controlled by environment variables which generally don't change during the lifetime of the program, so logging only needs to be initialized once at program startup, rather than during the creation of every handle.

Some programs might use temporary handles which are created and destroyed locally, while logging is more interested in capturing the entirety of the program. So the logging lifetime should be independent of `rocblas_handle` lifetimes, and controlled at program startup and program exit, or at other predefined points.

This is best done by changing the logging variables from non-static to static members of the handle class.

For profiling of BLAS function calls, this change is necessary to capture total function call counts independent of the lifetime of rocBLAS handles.

**Affects:** All rocBLAS kernels indirectly through logging implementation; no changes to the kernel codes which call logging are needed. `handle->layer_mode` and `handle->log_trace_os` are still valid even if `layer_mode` and `log_trace_os` are `static` members of `rocblas_handle`.

-------------------------------------------------------------------------------

## Check the rocBLAS scalar arguments for invalid pointers before logging.

**Details:** Currently, some logging functions may crash if the rocBLAS function is called with `nullptr` scalar pointer arguments and logging is enabled.

Checking of `nullptr` scalar arguments is now done before logging, and the rocBLAS function returns immediately with `rocblas_status_invalid_pointer` if a `nullptr` scalar argument pointer is passed.

**Alternative:** Always perform logging, independent of scalar argument pointers' validity, but test for, and indicate, `nullptr` scalar argument pointers, instead of unconditionally dereferencing them during logging.

**Affects:** All rocBLAS kernels

-------------------------------------------------------------------------------

## Replace the `replaceX<T>()` function with variable templates.

**Details:** The `replaceX` template function takes a template argument T, and based on `std::is_same<T, float>::value`, `std::is_same<T, double>::value`, etc., replaces the character `'X'` in a `std::string` with the precision letter of the type. This is a runtime operation which requires dynamically allocating a `std::string`, and then scanning and replacing characters in it. It is called 1-3 times per BLAS function call, once for each logging level which is enabled. For example, `replaceX<T>("rocblas_Xgemv")`. This is significant overhead.
```
// replaces X in string with s, d, c, z or h depending on typename T
template <typename T>
std::string replaceX(std::string input_string)
{
    if(std::is_same<T, float>::value)
    {
        std::replace(input_string.begin(), input_string.end(), 'X', 's');
    }
    else if(std::is_same<T, double>::value)
    {
        std::replace(input_string.begin(), input_string.end(), 'X', 'd');
    }
    else if(std::is_same<T, rocblas_float_complex>::value)
    {
        std::replace(input_string.begin(), input_string.end(), 'X', 'c');
    }
    else if(std::is_same<T, rocblas_double_complex>::value)
    {
        std::replace(input_string.begin(), input_string.end(), 'X', 'z');
    }
    else if(std::is_same<T, rocblas_half>::value)
    {
        std::replace(input_string.begin(), input_string.end(), 'X', 'h');
    }
    return input_string;
}
```
Usage:
```
        log_trace(handle,
                  replaceX<T>("rocblas_Xgemv"),
                  transA,
                  m,
                  n,
                  *alpha,
                  (const void*&)A,
                  lda,
                  (const void*&)x,
                  incx,
                  *beta,
                  (const void*&)y,
                  incy);
```
With C++14, variable templates can be used to concisely define C-style strings which depend on a template parameter. For example,
```
template <typename>
constexpr char rocblas_gemv_name[] = "unknown";
template <>
constexpr char rocblas_gemv_name<float>[] = "rocblas_sgemv";
template <>
constexpr char rocblas_gemv_name<double>[] = "rocblas_dgemv";
```
(In C++11, this can be done with `static const[expr]` data members of class templates, or with templatized `constexpr` functions; this change would still be justified even if C++14 variable templates were not used, because of the performance and code size savings.)

**Rationale:** The use of C++14 variable templates is a concise way of mapping types to values. The use of `constexpr char` arrays requires no dynamic memory allocation, unlike `std::string`, and the resolution of the name based on type is done at compile-time. The templatized `constexpr` arrays take up no space except where they are used -- they are "inlined" where they are used, but do not otherwise take up any space. Essentially,
```
        log_trace(handle,
                  rocblas_gemv_name<T>,
                  transA,
                  m,
                  n,
                  *alpha,
                  (const void*&)A,
                  lda,
                  (const void*&)x,
                  incx,
                  *beta,
                  (const void*&)y,
                  incy);
```
is the same as:
```
        log_trace(handle,
                  "rocblas_sgemv",       // rocblas_gemv_name<float>
                  transA,
                  m,
                  n,
                  *alpha,
                  (const void*&)A,
                  lda,
                  (const void*&)x,
                  incx,
                  *beta,
                  (const void*&)y,
                  incy);
```
or:
```
        log_trace(handle,
                  "rocblas_dgemv",       // rocblas_gemv_name<double>
                  transA,
                  m,
                  n,
                  *alpha,
                  (const void*&)A,
                  lda,
                  (const void*&)x,
                  incx,
                  *beta,
                  (const void*&)y,
                  incy);
```
... depending on `T`.

**Affects:** All rocBLAS kernels

-------------------------------------------------------------------------------

## Combine some kernel header files and the kernel .cpp files which use them.

**Details:** Some rocblas .cpp files #include some .h files, such as rocblas_gemv.cpp including gemv_device.h. The .h header files only contain template definitions specifically related to the .cpp files, and are not #included elsewhere.

**Rationale:** The kernel-specifc header files are only `#include`'d in one place, their associated .cpp file. They are not used elsewhere. They are very specific to one kernel, and are unlikely to be used elsewhere.

Putting the kernel templates in the same .cpp file makes the code simpler to read and manage, since you can see the device kernel template followed by the host function which calls it. It also allows making the template private to the kernel (see below), which is more cumbersome and harder to understand if a
header file is used.

The majority of the rocBLAS routines already have their kernel templates inside of their .cpp file, so this is not unusual. This is merely making a few routines conform to the majority of routines, and cutting down on the number of files.

Also, the merging of the device scalar and host scalar templates (see above), cuts the template code size in half, making it less desirable for the device kernels to be in separate .h files.

**Affects:** gemv, ger, syr, trtri_batched

-------------------------------------------------------------------------------

## Use `ssize_t`, `ptrdiff_t`, or `size_t` for kernel subscripts which are used in offset calculations.

**Details:** Some rocBLAS kernels perform address arithmetic using multiplication of indices by stride, and the addition of indices. The resulting calculations are used to index arrays.

**Rationale:** `ssize_t` or `ptrdiff_t` (signed) or `size_t` (unsigned) types should be used for address offset calculations which involve multiplication by strides and/or addition of indices, because while the array dimensions may all fit inside 32-bit rocblas_int values, the offsets within the arrays may exceed the capacity of 32-bit rocblas_int offsets. Since the native addressing mode is 64-bit, 64-bit offsets should be used in offset calculations.

If no subscript arithmetic is done except for trivially indexing into an array with a `rocblas_int` subscript, such as `A[tid]`, then it is not absolutely necessary to make the array subscripts 64-bit, because they will be implicitly converted to 64-bit `ptrdiff_t` when `A[tid]` is processed as `*(A + tid)`. But if a calculation such as `A[tid * incx]` or `A[i * lda + j]` is used, then the indices should be 64-bit, to avoid overflow during the offset calculations.

**Affects:** All rocBLAS kernels

-------------------------------------------------------------------------------

### If block sizes are used as template parameters, place them at the beginning rather than the end of the template parameter list, so that other template parameters at the end of the list can be automatically deduced, and only the block sizes have to be explicitly specified.

**Details:** Block sizes such as `NB`, `NB_X`, `GEMVN_DIM_X`, `GEMVN_DIM_Y`, `GEAM_DIM_X`, `GEAM_DIM_Y`, `IBD`, `IB`, and `JB` are currently `#define`'d as macros. They are sometimes `#undef`'ined and redefined in the same file. They are passed as template parameters sometimes, but almost always, the constant is the same throughout the entire file, and does not need to be a template parameter.

This change would replace block size macro definitions with `constexpr`definitions, which can be scoped to either apply to the whole .cpp file, or locally to a section of code.
```
    #define NB_X 256

    template <typename T, rocblas_int NB>
    void func(T* x, T*y, ...)
    {
      // Use NB for block size
    }

    func<T, NB_X>(x, y, ...);  // Must explicitly specify T and NB arguments
```
Becomes:
```
    constexpr rocblas_int NB = 256;

    template <typename T>
    void func(T* x, T* y, ...)
    {
      // Use NB for block size
    }

    func(x, y, ...);   // T can be automatically deduced from x, y arguments
```
**Rationale:** `constexpr` block size parameters are preferred to macro block size parameters, because of the dangers of inadvertent macro substitution, and the inability to localize macro definitions without repeatedly `#define`'ing and `#undef`'ing them. With `constexpr` block size definitions, they are the same as compile-time constants and they can be localized or globalized as with any other variable.

When a template is always passed the same block size for a given function, it is sometimes better to simply define a `constexpr` value which applies to the whole file, rather than passing a template argument which never changes. `constexpr` variables declared in namespace scope (outside functions) have internal linkage, and won't interfere with variables of the same name in other translation units.

Also, when a block size template parameter is at the end of a template parameter list, it requires the explicit specification of all template arguments for every call. When a block size template parameter is not used, or it is at the beginning of the template parameter list, then the other template type arguments at the end of the list can be automatically deduced in the call, and do not have to be explicitly specified. This reduces the chances of error by mistakenly listing the wrong type.

**Affects:** Almost all rocBLAS kernels

-------------------------------------------------------------------------------

## Put the C++ implementation of rocBLAS functions inside of unnamed namespaces, to avoid namespace conflicts with other compilation units. Only the C APIs are exposed (with very few exceptions as needed).

For a `rocblas_<function>.cpp:`
```
namespace {
// rocBLAS implementation functions, usually C++ templated code
}

extern "C" {
// C wrappers, using functions hidden in the unnamed namespace
}
```
**Rationale:** Avoid name conflicts with other parts of a program which links with rocBLAS. If a rocBLAS implementation uses an identifier name which happens to be used in another part of the user's program, then this avoids a conflict.

This is similar to declaring functions and constants "static".

**Alternative:** Put all internal rocBLAS identifiers in the "rocblas" namespace, only exposing C interfaces and type declarations in the global namespace. If access to a rocBLAS identifier is needed in C++, it can use the rocblas namespace to access it.

**Longer-term:** Place all C++ rocBLAS types and functions in the "rocblas" namespace, not just the .cpp implementations of rocBLAS kernels. This involves more work, and is not as simple as using unnamed namespaces for the C++ implementation functions, since it requires coordinated changes across files. Most production libraries written in C++ use their own namespaces to avoid name conflicts, even if they have C APIS.

**Affects:** All rocBLAS kernels

-------------------------------------------------------------------------------

## Pass `workspace` as the `result` pointer when a scalar result is on host, avoiding the need for a flag and conditional code in the kernel.

**Details:** Right now, the code for returning scalar results distinguishes between device pointers and host pointers by passing a flag and using different code depending on the flag. Since the device code cannot write to host memory, it writes to a temporary workspace instead. For example:
```
    template<..., rocblas_int flag>
    void function(..., T* workspace, T* result)
    {
      // ... compute scalar value

      if (flag)
      {
         *result = value;
      }
      else
      {
         workspace[0] = value;
      }
    }

    if (handle->pointer_mode) == rocblas_pointer_mode_device)
    {
       hipLaunchKernelGGL(  func<..., 1>, ..., workspace, result)
      //                           ^^^
      // all func template arguments must be specified because last one
      // is a non-type parameter which cannot be automatically deduced
    }
    else
    {
       hipLaunchKernelGGL(  func<..., 0>, ..., workspace, result)

      // Copy scalar from workspace[0] to *result
       RETURN_IF_HIP_ERROR(
           hipMemcpy(result, workspace, sizeof(*result), hipMemcpyDeviceToHost));
    }
```
With this change, the code above gets changed to:
```
    template<...>
    void function(..., T* workspace, T* result)
    {
      // ... compute scalar value

      *result = value;      // unconditionally write to *result without using flag
    }

    if (handle->pointer_mode) == rocblas_pointer_mode_device)
    {
       hipLaunchKernelGGL(  func, ..., workspace, result)
      //                       ^^^
      // func doesn't require <...> arguments anymore, because by removing flag,
      // all other template arguments can be deduced automatically
    }
    else
    {
       hipLaunchKernelGGL(  func, ..., workspace, workspace)
       //                                            ^^^    result = &workspace[0]
       RETURN_IF_HIP_ERROR(
           hipMemcpy(result, workspace, sizeof(*result), hipMemcpyDeviceToHost));
    }
```
We pass `workspace` as the `result` pointer when the result is on the host, and so writing to `*result` in the function writes to `workspace[0]`. We then copy from `workspace[0]` device memory to *result host memory as we did before.

For all kernels except amax, amin, the code above can be shortened even more, by merging the two calls into one, and passing the last argument depending on pointer mode:
```
   hipLaunchKernelGGL(  func, ..., workspace,
        handle->pointer_mode != rocblas_pointer_mode_device ? workspace : result)

   if ( handle->pointer_mode != rocblas_pointer_mode_device )
       RETURN_IF_HIP_ERROR(
           hipMemcpy(result, workspace, sizeof(*result), hipMemcpyDeviceToHost));
```
**Rationale:** The code is much simpler. Also, by not using the flag at the end of the template argument list, we avoid having to specify template arguments which can be automatically deduced, while as long as the flag is placed at the end, we always have to specify all template arguments.

**Affects:** asum, amax, amin, dot, nrm2

-------------------------------------------------------------------------------

## Reimplement reduction functions using template metaprogramming instead of hard-coded loops, and put common code for reductions in one place, to avoid code duplication across several routines.

**Details:** The original code for reductions, in `blas1/device_template.h`, was hard-coded with specific limits, and the code was hand-unrolled. For example:
```
template <rocblas_int n, typename T>
__device__ void rocblas_sum_reduce(rocblas_int tx, T* x)
{
    // clang-format off
    __syncthreads();
    if(n > 512) { if(tx < 512 && tx + 512 < n) { x[tx] += x[tx + 512]; } __syncthreads(); }
    if(n > 256) { if(tx < 256 && tx + 256 < n) { x[tx] += x[tx + 256]; } __syncthreads(); }
    if(n > 128) { if(tx < 128 && tx + 128 < n) { x[tx] += x[tx + 128]; } __syncthreads(); }
    if(n >  64) { if(tx <  64 && tx +  64 < n) { x[tx] += x[tx +  64]; } __syncthreads(); }
    if(n >  32) { if(tx <  32 && tx +  32 < n) { x[tx] += x[tx +  32]; } __syncthreads(); }
    if(n >  16) { if(tx <  16 && tx +  16 < n) { x[tx] += x[tx +  16]; } __syncthreads(); }
    if(n >   8) { if(tx <   8 && tx +   8 < n) { x[tx] += x[tx +   8]; } __syncthreads(); }
    if(n >   4) { if(tx <   4 && tx +   4 < n) { x[tx] += x[tx +   4]; } __syncthreads(); }
    if(n >   2) { if(tx <   2 && tx +   2 < n) { x[tx] += x[tx +   2]; } __syncthreads(); }
    if(n >   1) { if(tx <   1 && tx +   1 < n) { x[tx] += x[tx +   1]; } __syncthreads(); }
    // clang-format on
}
// end sum_reduce
```
This can be written with template metaprogramming to automatically generate the code above, without imposing an arbitrary limit of 1024. The limit can be adjusted at compile-time and will automatically generate the above. The `__attribute__((flatten))` can be specified for functions so that any recursive calls are inlined and the result is a single function.

```
// Recursively compute reduction
template <rocblas_int k, typename REDUCE, typename T>
struct rocblas_reduction_s
{
    __forceinline__ __device__ void operator()(rocblas_int tx, T* x)
    {
        // Reduce the lower half with the upper half
        if(tx < k)
            REDUCE{}(x[tx], x[tx + k]);
        __syncthreads();

        // Recurse down with k / 2
        rocblas_reduction_s<k / 2, REDUCE, T>{}(tx, x);
    }
};

// leaf node for terminating recursion
template <typename REDUCE, typename T>
struct rocblas_reduction_s<0, REDUCE, T>
{
    __forceinline__ __device__ void operator()(rocblas_int tx, T* x) {}
};

/*! \brief general parallel reduction

    \details

    @param[in]
    n         rocblas_int. assume a power of 2
    @param[in]
    T         element type of vector x
    @param[in]
    REDUCE    reduction functor
    @param[in]
    tx        rocblas_int. thread id
    @param[inout]
    x         pointer storing vector x on the GPU.
              usually x is stored in shared memory;
              x[0] store the final result.
    ********************************************************************/
template <rocblas_int NB, typename REDUCE, typename T>
__attribute__((flatten)) __device__ void rocblas_reduction(rocblas_int tx, T* x)
{
    static_assert(NB > 1 && !(NB & (NB - 1)), "NB must be a power of 2");
    __syncthreads();
    rocblas_reduction_s<NB / 2, REDUCE, T>{}(tx, x);
}
```

**NOTE:** Performance tests were run on the new BLAS-1 reduction code, versus the old code. The performance is very close (almost equal, within run-to-run variance). Also, it has been verified, by inspecting the assembly code output, that the compiler inlines all of the code in the refactored implementation, eliminating the recursion in the source code.  So the refactored code is much more readable and maintainable, without any performance loss.

Reductions have a "part1" and a "part2" kernel, as explained in detail by the original comments. These two functions can be templatized to avoid duplication. For example:
```
// kennel 1 writes partial results per thread block in workspace; number of partial results is blocks
template <rocblas_int NB, typename FETCH, typename REDUCE, typename Ti, typename To>
__global__ void
rocblas_reduction_kernel_part1(rocblas_int n, const Ti* x, rocblas_int incx, To* workspace)
{
    ssize_t tx  = hipThreadIdx_x;
    ssize_t tid = hipBlockIdx_x * hipBlockDim_x + tx;
    __shared__ To tmp[NB];

    // bound
    if(tid < n)
        tmp[tx] = FETCH{}(x[tid * incx], tid);
    else
        tmp[tx] = To{}; // pad with default value

    rocblas_reduction<NB, REDUCE>(tx, tmp);

    if(tx == 0)
        workspace[hipBlockIdx_x] = tmp[0];
}

// kernel 2 gathers all the partial results in workspace and finishes the final reduction;
// number of threads (NB) loop blocks
template <rocblas_int NB, typename FINALIZE, typename REDUCE, typename To, typename Td>
__global__ void rocblas_reduction_kernel_part2(rocblas_int nblocks, To* workspace, Td* result)
{
    ssize_t tx = hipThreadIdx_x;
    __shared__ To tmp[NB];

    if(tx < nblocks)
    {
        tmp[tx] = workspace[tx];

        // bound, loop
        for(ssize_t i = tx + NB; i < nblocks; i += NB)
            REDUCE{}(tmp[tx], workspace[i]);
    }
    else
    {   // pad with default value
        tmp[tx] = To{};
    }

    if(nblocks < 32)
    {
        // no need parallel reduction
        __syncthreads();

        if(tx == 0)
            for(ssize_t i = 1; i < nblocks; i++)
                REDUCE{}(tmp[0], tmp[i]);
    }
    else
    {
        // parallel reduction
        rocblas_reduction<NB, REDUCE>(tx, tmp);
    }

    // Store result on device or in workspace
    if(tx == 0)
        *result = FINALIZE{}(tmp[0]);
}
```
The `FETCH`, `REDUCE`, and `FINALIZE` functors passed to these templates, abstract out all of the differences between reduction functions, so that all reduction BLAS functions can use the code above. `FETCH` tells the code what to do when loading a value, such as taking its absolute value (asum, amax, amin) or taking the square of its absolute value (nrm2); `REDUCE{}(a, b)` tells how to update `a` with `b` to perform the reduction; and `FINALIZE` tells what to do with the final value, such as to extract the index and add 1 (amax, amin), take the square root (nrm2), or nothing (asum). All of these `FETCH`, `REDUCE`, and `FINALIZE` functors are inlined, and abstract out the differences between the BLAS reduction kernels.

The `To{}` syntax generates a default value for the reduction, which is zero for most functions, but generates a special sentinel value for amax, amin. When `To` is an arithmetic type, `To{}` generates zero. When `To` is a class type like a (maxvalue, maxindex) pair, `To{}` calls the default constructor if it's defined.

Dot product follows a similar pattern, but since it has two vector arguments `(x,y)` and may need more hardware-specific optimization, being more important than the above functions, it is not included with them. However, it shares some of the same `sum` reduction code, as does gemv.

**Rationale:** Reduce duplicated code, and make generation of unrolled loops automatic rather than hard-coded, allowing limits to be easily changed if necessary.

**Affects:** asum, amax, amin, nrm2 (partially: dot, gemv)

-------------------------------------------------------------------------------

## Fix and simplify TRSV.

**Details:** The original implementation of TRSV has a bug: it compares the problem size to the preallocated buffer size, but it uses the wrong sense of comparison:
```
        if(WORKBUF_TRSV_X_SZ <= m)
        {
        // (use preallocated buffer)
        }
        else
        {
        // (allocate a larger temporary buffer of size m)
        }
```
The sense of comparison is backwards, and the implementation can segfault if `m` is larger than the preallocated buffer size. Also, `m` needs to be multiplied by `sizeof(T)`, since `WORKBUF_TRSV_X_SZ` is measured in bytes.

Also, the implementation is unnecessarily complicated with respect to allocating the memory for the scalar `alpha` variable, which is always `1.0`. The new implementation simply uses `static const` variables on host or device to store the value of alpha, eliminating the need to dynamically allocate and deallocate memory for `alpha`:
```
// alpha is a constant value of 1.0 either on host or device depending on pointer mode
static constexpr T alpha_h{1};
static const __device__ T alpha_d{1};
const T* alpha = pointer_mode == rocblas_pointer_mode_device ? &alpha_d : &alpha_h;
```
Finally, there are other simplifications, **reducing the code size by around 50 lines.**

**Affects:** trsv
-------------------------------------------------------------------------------

## Refactor `rocblas_auxiliary.cpp` with respect to pointer arithmetic, `size_t` casts, and `try`-`catch`  blocks.

**Details:**

The original implementation of `rocblas_auxiliary.cpp` casts pointers to `size_t` to perform address arithmetic calculations.

It also has `try`-`catch` blocks around the bodies of many functions.

The new implementation uses `(char *)` pointers to perform address arithmetic, and uses `try`-`catch` function blocks.

The new implementation also reduces the number of casts to `(size_t)` required, by declaring variables `size_t` or by limiting the number of `(size_t)` casts in a given expression to 1, since if `size_t` is larger than the other operand types, the other operands will automatically be promoted to `size_t`.

**Rationale:** While usually okay on most platforms, since `size_t` is almost always the same size as a pointer, it is better to cast pointers to type `(char *)` to perform byte address arithmetic on them, rather than performing integer arithmetic on pointers converted to `size_t`, and then converting them back to pointers. `sizeof(char)` == 1 by definition, and so char pointer arithmetic is always done in bytes.

If one cast in an expression is enough to bring the operands up to `size_t`, then one cast should be sufficient. It hurts readability to have every subexpression prefixed with a `(size_t)` cast, when one is sufficient. Also, simply declaring one of the variables in an expression as size_t can promote the rest of the variables and expression to `size_t`, which is cleaner than having to use explicit `(size_t)` casts.

`try`-`catch` blocks which surround the entire body of a function can be replaced with function `try`-`catch` blocks, which are declared outside of the function. They are only absolutely required for constructors, but they can reduce the indentation level of the code by one level, and reduce distraction from the body of the code, by putting the catch block outside of the function.

-------------------------------------------------------------------------------

## Simplify dim3 usage

**Details:** `dim3` is the HIP datatype used to express 3D topology.

The old code used `dim3` more than necessary, such as saying `dim3(threads)` when `threads` is already a `dim3` variable, so `dim3(threads)` is simply a copy (a no-op).

`dim3`'s constructor accepts 0-3 dimensions, with any omitted trailing dimensions defaulting to 1. Thus `dim3(NB, 1, 1)` can be more concisely written as `dim3(NB)`.

This change removes redundant copies of `dim3`, such as `dim3(var)`, and it removes trailing dimensions of 1, which are the default for `dim3` anyway.

**Rationale:** More concise code, without unnecessary verbosity about dimensions.

-------------------------------------------------------------------------------
# Appendix 

## The first two changes above are able to reduce this AXPY code (106 LoC sans comment):

```
template <typename T>
__global__ void axpy_kernel_host_scalar(
    rocblas_int n, const T alpha, const T* x, rocblas_int incx, T* y, rocblas_int incy)
{
    int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;

    if(incx >= 0 && incy >= 0)
    {
        if(tid < n)
        {
            y[tid * incy] += (alpha)*x[tid * incx];
        }
    }
    else if(incx < 0 && incy < 0)
    {
        if(tid < n)
        {
            y[(1 - n + tid) * incy] += (alpha)*x[(1 - n + tid) * incx];
        }
    }
    else if(incx >= 0)
    {
        if(tid < n)
        {
            y[(1 - n + tid) * incy] += (alpha)*x[tid * incx];
        }
    }
    else
    {
        if(tid < n)
        {
            y[tid * incy] += (alpha)*x[(1 - n + tid) * incx];
        }
    }
}

template <typename T>
__global__ void axpy_kernel_device_scalar(
    rocblas_int n, const T* alpha, const T* x, rocblas_int incx, T* y, rocblas_int incy)
{
    int tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
    // bound
    if(incx >= 0 && incy >= 0)
    {
        if(tid < n)
        {
            y[tid * incy] += (*alpha) * x[tid * incx];
        }
    }
    else if(incx < 0 && incy < 0)
    {
        if(tid < n)
        {
            y[(1 - n + tid) * incy] += (*alpha) * x[(1 - n + tid) * incx];
        }
    }
    else if(incx >= 0)
    {
        if(tid < n)
        {
            y[(1 - n + tid) * incy] += (*alpha) * x[tid * incx];
        }
    }
    else
    {
        if(tid < n)
        {
            y[tid * incy] += (*alpha) * x[(1 - n + tid) * incx];
        }
    }
}

// (In host code)

    if(rocblas_pointer_mode_device == handle->pointer_mode)
    {
        hipLaunchKernelGGL(axpy_kernel_device_scalar, // different kernel than \/
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           alpha,                   // alpha passed as pointer
                           x,
                           incx,
                           y,
                           incy);
    }
    else // alpha is on host
    {
        T scalar = *alpha;
        if(0.0 == scalar)
        {
        }

        hipLaunchKernelGGL(axpy_kernel_host_scalar, // different kernel than /\
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           scalar,                  // alpha passed as FP value
                           x,
                           incx,
                           y,
                           incy);
    }
```

## To this AXPY code (39 LoC sans comment):

```
template <typename T, typename U>
__global__ void axpy_kernel(
    rocblas_int n, U alpha_device_host, const T* x, rocblas_int incx, T* y, rocblas_int incy)
{
    auto alpha = load_scalar(alpha_device_host);
    ssize_t tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
    if(tid < n)
        y[tid * incy] += alpha * x[tid * incx];
}

// (In host code)

    if(incx < 0)
        x -= ssize_t(incx) * (n - 1);
    if(incy < 0)
        y -= ssize_t(incy) * (n - 1);

    if(handle->pointer_mode == rocblas_pointer_mode_device)
        hipLaunchKernelGGL(axpy_kernel,     // same kernel for device and host scalar
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           alpha,           // alpha passed as pointer
                           x,
                           incx,
                           y,
                           incy);
    else if(*alpha)                         // alpha is on host
        hipLaunchKernelGGL(axpy_kernel,     // same kernel for device and host scalar
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           *alpha,          // alpha passed as FP value
                           x,
                           incx,
                           y,
                           incy);
```

## Commented explanation of the simplified AXPY kernel (other kernels are similar):

```
// Type T is the computational datatype of this rocBLAS kernel.
// Type T is automatically deduced based on the types of x and y.
// Type U is automatically deduced based on the type of alpha_device_host.
// Type U can either be a scalar FP value or a pointer to a scalar FP value.
template <typename T, typename U>
__global__ void axpy_kernel(
    rocblas_int n, U alpha_device_host, const T* x, rocblas_int incx, T* y, rocblas_int incy)
{
    // Load the scalar if it on device and alpha_device_host is a pointer;
    // copy the scalar if it is on host and alpha_device_host is a FP value.
    // load_scalar is always inlined; there is no performance degradation.
    //
    auto alpha = load_scalar(alpha_device_host);

    // Use ssize_t instead of rocblas_int so that overflow won't happen when tid
    // is multiplied by incx. This allows memory offsets > 4 GB even though the
    // rocblas_int parameter's maximum is < 4 GB. (See #8 above.)
    //
    ssize_t tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;

    // x and y are pre-adjusted on the host based on the signs of incx, incy.
    // No checking of incx, incy in the kernel is necessary anymore. They are
    // simply used unconditionally.
    //
    if(tid < n)
        y[tid * incy] += alpha * x[tid * incx];
}

// (In host code)
//
// When incx and incy are negative, they are multiplied by (n - 1) and the
// negative result is subtracted from x or y, making x or y point towards the
// end of the array, and then x[tid * incx] and y[tid * incx] work backwards
// from the end of the array instead of forwards.
//
// We cast incx and incy to ssize_t to avoid overflow when multiplying by (n-1),
// since the pointer being adjusted is comparable to size ssize_t / size_t.
// (See #8 above.)
//
    if(incx < 0)
        x -= ssize_t(incx) * (n - 1);
    if(incy < 0)
        y -= ssize_t(incy) * (n - 1);

// We call the same kernel for device scalars and host scalars.
// The template above infers the type U based on the value of alpha_device_host
// If alpha_device_host is a pointer, the kernel above loads the scalar on the
// device. If alpha is a FP value, the kernel above copies the scalar passed in.
// (For simplicity, the extra variable "scalar" in the code above is removed.)
//
    if(handle->pointer_mode == rocblas_pointer_mode_device)
        hipLaunchKernelGGL(axpy_kernel,     // same kernel for device and host scalar
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           alpha,           // alpha passed as pointer
                           x,
                           incx,
                           y,
                           incy);
    else if(*alpha)                         // alpha is on host
        hipLaunchKernelGGL(axpy_kernel,     // same kernel for device and host scalar
                           dim3(blocks),
                           dim3(threads),
                           0,
                           rocblas_stream,
                           n,
                           *alpha,          // alpha passed as FP value
                           x,
                           incx,
                           y,
                           incy);
```

## `load_scalar()` implementation

```
// Load a scalar. If the argument is a pointer, dereference it; otherwise copy
// it. Allows the same kernels to be used for host and device scalars.

// For host scalars
template <typename T>
__attribute__((always_inline)) __device__ __host__ inline T load_scalar(T x)
{
    return x;
}

// For device scalars
template <typename T>
__attribute__((always_inline)) __device__ __host__ inline T load_scalar(const T* xp)
{
    return *xp;
}

// For rocblas_half2, we broadcast a fp16 across two halves
template <>
__attribute__((always_inline)) __device__ __host__ inline rocblas_half2 load_scalar(const rocblas_half2* xp)
{
    auto x = *reinterpret_cast<const _Float16*>(xp);
    return {x, x};
}
```


amcamd:
Once this PR  is merged can you update the rocBLAS Wiki here https://github.com/ROCmSoftwarePlatform/rocBLAS/wiki/x0.Profile-rocBLAS-kernels  with information on profile logging. Add information on  ROCBLAS_LAYER=4 and  ROCBLAS_LOG_PROFILE_PATH.

PR #453, opened 2019-01-17
use string in place of char to extend rocblas-bench datatypes

amcamd:
- For function gemm_ex in rocblas-bench use std::string in place of char for
  - a_type
  - b_type
  - c_type
  - d_type
  - compute_type
- This allows for strings like i8_r, i32_r, f16_r, f32_4 f64_r in place of characters h,s,d. The characters do not cover new types like int8 and int32.
- The multi-character strings correspond to those in rocblas_datatype enum values rocblas_datatype_i8_r, rocblas_datatype_i32_r, ....


leekillough:
I recommend using `char` instead of `std::string`. `std::string` is performance- and memory-expensive by comparison.

When a string needs to be stored (not merely referenced), maybe 8 characters, as in `char a_type[8]`, etc. Otherwise, such as when returning a string from a function, or accepting a string argument, use `const char *`.

Also, since rocBLAS is a C-API, using `std::string` requires allocation, copying, and de-allocating temporaries for each `std::string`, while passing `char*` costs nothing.

For `gemm_ex` arguments, I recommend using `const char*`

leekillough:
In general, don't use std::string in high-performance software, unless it is clear that the string will always need to be dynamically allocated because it has arbitrary unknown length.

leekillough:
I would use `rocblas_datatype` enums instead of strings where possible. When interfacing with string input, say from YAML files or `rocblas-bench` arguments, another routine can be called to convert a string to the `rocblas_datatype` `enum`, and then that is used for the argument.

leekillough:
Headline is misleading: We are not changing `gemm_ex` to have `std::string` arguments.

But the comment still stands: `const char*` should be used instead of `std::string` where possible.

leekillough:
Also, I wouldn't create `rocblas_datatype2char.cpp`. I would put a `constexpr` function in `rocblas_datatype2char.h`, like the existing functions (for it to be `constexpr`, it must use `const char*` instead of `std::string`, because `std::string` is not a literal type).

One of the refactoring changes I did recently, was to remove unnecessary separation of headers and `.cpp` files. This would be reversing that.

In modern C++, most stuff is done in headers. You don't see `.h` or `.hpp` and `.cpp` files separated like you used to.

PR #469, opened 2019-01-29
TRSM Optimization: Custom Trtri Kernel

mahmoodw:
contributes towards #172044

- New custom trtri kernel used for TRSM runs 256 threads in a block to work on two 16x16 diagonal tiles at the same time.
- After completing the inverses the results are used within the kernel to preform 2 gemms. This was previously done outside of the kernel.

### DIFF
--- a/library/src/blas3/trtri_device.h
+++ b/library/src/blas3/trtri_device.h
@@ -165,4 +165,212 @@ __device__ void trtri_device(rocblas_fill uplo,
     }
 }
 
+template <typename T, rocblas_int IB>
+__device__ void custom_trtri_device(rocblas_fill uplo,
+                                    rocblas_diagonal diag,
+                                    rocblas_int n,
+                                    const T* A,
+                                    rocblas_int lda,
+                                    T* invA,
+                                    rocblas_int ldinvA)
+{
+
+    // quick return
+    if(n <= 0)
+        return;
+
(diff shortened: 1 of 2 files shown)


library/src/blas3/trtri_device.h
```diff
@@ -165,4 +165,212 @@ __device__ void trtri_device(rocblas_fill uplo,
...
+        {
+
+            T reg = 0;
+            // use the diagonal one to update current column
+            if(index > col)
+            {
+                reg += diagP[index + col * n] * diagP[col + col * n];
+            }
+
+            // __syncthreads(); // if IB < 64, this synch can be avoided on AMD Fiji
```
mahmoodw: Please advise if these syncthreads can be removed safely for required architectures.

bragadeesh: Best to leave the __syncthreads(), unless you are seeing a big perf impact.
Its presence depends on the algorithm; are only 64-threads in the block collaborating and need to finish before proceeding, or is all 256 threads? If it is 64 only, it is possible to remove

yoichiyoshida: That part of the code is in a branch, only 32 of the 256 threads are collaborating; if anything half the wavefront won't hit the barrier and will introduce more issues if left in. There is a sync after the branch which should keep everything consistent afterwards.

PR #472, opened 2019-01-31
Fix compilation of IAMAX, IAMIN

### LINKED ISSUE #457: regression due to __shared__ variable for hip-clang
The following line recently caused regression in hip-clang:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/ed0d2b161eb6c28ed36732d10843e433d359b1bc/library/src/blas1/rocblas_amax_amin.h#L30

errror msg:
[08:53:58][Step 2/2] clang: warning: argument unused during compilation: '--hip-device-lib-path=/opt/rocm/hip/lib/bitcode' [-Wunused-command-line-argument]
[08:53:59][Step 2/2] In file included from /work/rocblas/library/src/blas1/rocblas_amin.cpp:6:
[08:53:59][Step 2/2] In file included from /work/rocblas/library/src/blas1/rocblas_amax_amin.h:11:
[08:53:59][Step 2/2] /work/rocblas/library/src/blas1/reduction.h:154:19: error: initialization is not supported for __shared__ variables.
[08:53:59][Step 2/2]     __shared__ To tmp[NB];
[08:53:59][Step 2/2]                   ^~~
[08:53:59][Step 2/2] /work/rocblas/library/src/blas1/reduction.h:231:25: note: in instantiation of function template specialization 'rocblas_reduction_kernel_part1<1024, rocblas_fetch_amax_amin<float>, rocblas_reduce_amin, float, index_value_t<float> >' requested here
[08:53:59][Step 2/2]     hipLaunchKernelGGL((rocblas_reduction_kernel_part1<NB, FETCH, REDUCE>),
[08:53:59][Step 2/2]                         ^
[08:53:59][Step 2/2] /work/rocblas/library/src/blas1/rocblas_amax_amin.h:173:19: note: in instantiation of function template specialization 'rocblas_reduction_kernel<1024, rocblas_fetch_amax_amin<float>, rocblas_reduce_amin, rocblas_finalize_amax_amin, float, index_value_t<float>, int>' requested here
[08:53:59][Step 2/2]     auto status = rocblas_reduction_kernel<NB,
[08:53:59][Step 2/2]                   ^
[08:53:59][Step 2/2] /work/rocblas/library/src/blas1/rocblas_amax_amin.h:193:12: note: in instantiation of function template specialization 'rocblas_iamaxmin<float, float>' requested here
[08:53:59][Step 2/2]     return rocblas_iamaxmin<float>(handle, n, x, incx, result); 

This is because hip-clang does not allow default constructor of `__shared__` variables to initialize data members. This is because `__shared__` variables cannot be initialized in either amdgpu or nvptx target. CUDA has the same restriction.

The fix is to remove initialization of data members from the default constructor and add an init() function to do the initialization, and call that function separately.

leekillough:
resolves #457 ___

Summary of proposed changes:
Instead of using `T{}` to generate default values for a class, use variable templates specialized for classes. This avoids needing to define non-trivial default constructors, which are not allowed in `__shared__` variables. No further initialization of `__shared__` variables is required, because the original code did not depend on it.

### DIFF
--- a/library/src/blas1/reduction.h
+++ b/library/src/blas1/reduction.h
@@ -139,6 +139,13 @@ struct rocblas_finalize_identity
     }
 };
 
+// Emulates value initialization T{}. Allows specialization for certain types.
+template <typename T>
+struct default_value
+{
+    __forceinline__ __host__ __device__ constexpr T operator()() const { return {}; }
+};
+
 // kennel 1 writes partial results per thread block in workspace; number of partial results is
 // blocks
 template <rocblas_int NB,
@@ -157,7 +164,7 @@ rocblas_reduction_kernel_part1(rocblas_int n, const Ti* x, rocblas_int incx, To*
     if(tid < n)
         tmp[tx] = FETCH{}(x[tid * incx], tid);
     else
-        tmp[tx] = To{}; // pad with default value
+        tmp[tx] = default_value<To>{}(); // pad with default value
 
     rocblas_reduction<NB, REDUCE>(tx, tmp);
 
@@ -187,7 +194,7 @@ __global__ void rocblas_reduction_kernel_part2(rocblas_int nblocks, To* workspac
     }
     else
     { // pad with default value
-        tmp[tx] = To{};
+        tmp[tx] = default_value<To>{}();
     }
 
     if(nblocks < 32)
--- a/library/src/blas1/rocblas_amax_amin.h
+++ b/library/src/blas1/rocblas_amax_amin.h
@@ -27,9 +27,21 @@ struct index_value_t
     // important: index must come first, so that index_value_t* can be cast to rocblas_int*
     rocblas_int index;
     T value;
-    __forceinline__ __host__ __device__ index_value_t() : index(-1) {}
 };
 
+// Specialization of default_value for index_value_t<T>
+template <typename T>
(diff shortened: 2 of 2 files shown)


yxsamliu:
unfortunately this does not work with hip-clang. I got error like this:
In file included from /home/yaxunl/git/rocblas/rocBLAS/library/src/blas2/rocblas_gemv.cpp:11:
/home/yaxunl/git/rocblas/rocBLAS/library/src/blas2/../blas1/reduction.h:144:33: error: dynamic initialization is not supported for __device__, __constant__, and __shared__ variables.
__host__ __device__ constexpr T default_value{};
                                ^            ~~

leekillough:
> unfortunately this does not work with hip-clang. I got error like this:
> In file included from /home/yaxunl/git/rocblas/rocBLAS/library/src/blas2/rocblas_gemv.cpp:11:
> /home/yaxunl/git/rocblas/rocBLAS/library/src/blas2/../blas1/reduction.h:144:33: error: dynamic initialization is not supported for **device**, **constant**, and **shared** variables.
> **host** **device** constexpr T default_value{};
> ^ ~~

I replaced variable templates with class templates containing functor. More verbose, but it avoids initialization issues.

PR #477, opened 2019-02-04
fix for 9 known bugs

wbgilmartin:
this fixes the issue with the 9 known bugs.

The change was to change the range in the random number initialization to prevent overflow.

All the rocblas-test's were verified to pass on vega10.

-  
-  
-

### DIFF
--- a/clients/gtest/known_bugs.yaml
+++ b/clients/gtest/known_bugs.yaml
@@ -3,12 +3,3 @@
 # Wildcards can be used for the function
 
 Known bugs:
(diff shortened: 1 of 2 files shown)


leekillough:
Has anyone done a study on the statistical distribution of matrix multiplication based on distributions of its arguments' elements, independent or not? 

Because reducing the number of states from 3 to 2 makes the matrices all (0,1)-matrices essentially, which can catch less cases.

Would introducing negative values even things out, like using values between [-2,2]?  (We already conditionally negate values somewhere else -- see `rocblas_init.hpp` -- this would be changing the distribution to include negative random values -- the former is more controlled but the latter might average out more.)

Don't forget to change the comment:
```
/*! \brief  generate a random number in range [1,2,3] */
```

wbgilmartin:
I used initialization range [-2,2].

all tests pass on both vega10 and vega20

PR #483, opened 2019-02-15
try splitting tests

amcamd:
- I am not sure if this will work
- it aims to split tests into gemm and non-gemm


amcamd:
filters are not correct.

PR #488, opened 2019-02-21
env variable GTEST_LISTENER for NO_PASS_LINE_IN_LOG

amcamd:
- Jenkinsfile sets environment variable GTEST_LISTNER to "NO_PASS_LINE_IN_LOG" to omit pass line from log. This reduces the size of the log file. 
- If envirnoment variable GTEST_LISTNER is not set to "NO_PASS_LINE_IN_LOG" then gtest will output pass and fail lines.


leekillough:
Is it supposed to be `GTEST_LISTNER` instead of `GTEST_LISTENER`?

amcamd:
Thanks for this correction, I changed it from LISTNER to LISTENER.

PR #489, opened 2019-02-23
fix for precision specifiers

leekillough:
Allow single-precision letters or longer type names to be used with `rocblas-bench` anywhere types are expected on the command-line.

Remove `precision` as being a main variable in `rocblas-bench`, and makes it simply a default value for `a_type`, `b_type`, `c_type`, `d_type`, `compute_type`, so that after those are set, `precision` is forgotten downstream. That is how the YAML-driven tests work already, since the `Arguments` data structure doesn't use `precision`.

Simplify error-handling in `rocblas-bench` by using exceptions.

Refactor `rocblas-bench` to be much simpler using templates, getting rid of all the precision testing in the body of the code, and relying instead of the type dispatch mechanism.

Fix a couple of typos in type string conversion.

### DIFF
--- a/clients/include/rocblas_arguments.hpp
+++ b/clients/include/rocblas_arguments.hpp
@@ -15,11 +15,6 @@
 #include "rocblas.h"
 #include "rocblas_datatype2string.hpp"
 
-typedef enum rocblas_initialization_ {
-    rocblas_initialization_random_int = 111,
-    rocblas_initialization_trig_float = 222,
-} rocblas_initialization;
-
 /* ============================================================================================ */
 /*! \brief Class used to parse command arguments in both client & gtest   */
 /* WARNING: If this data is changed, then rocblas_common.yaml must also be changed. */
@@ -80,9 +75,10 @@ struct Arguments
 
     // Validate input format.
     // rocblas_gentest.py is expected to conform to this format.
+    // rocblas_gentest.py uses rocblas_common.yaml to generate this format.
     static void validate(std::istream& ifs)
     {
-        auto error = []() {
+        auto error = [] {
             std::cerr << "Fatal error: Binary test data does match input format.\n"
                          "Ensure that rocblas_arguments.hpp and rocblas_common.yaml\n"
                          "define exactly the same Arguments, that rocblas_gentest.py\n"
@@ -151,13 +147,6 @@ struct Arguments
     }
 
     private:
-    /* =============================================================================================
-     */
-    /* All Arguments data members are above. Below are support functions for reading and formatting
-     */
-    /* =============================================================================================
-     */
-
     // Function to read Structures data from stream
     friend std::istream& operator>>(std::istream& str, Arguments& arg)
     {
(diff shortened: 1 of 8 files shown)


clients/include/rocblas_test.hpp
```diff
@@ -113,11 +113,6 @@ class RocBLAS_TestName
     {
         // Placed inside function to avoid dependency on initialization order
         static table_t* table = test_cleanup::allocate<table_t>(&table);
-        if(!table)
-        {
-            fputs("Internal error: nullptr\n", stderr);
-            exit(EXIT_FAILURE);
-        }
```
leekillough: There's no need to test `table` for `nullptr` -- the `test_cleanup::allocate` already throws a `std::bad_alloc` exception if memory cannot be allocated. This is also a rare occurrence which should not need to be singled out. Removing this makes the code simpler.

clients/include/rocblas_datatype2string.hpp
```diff
@@ -129,42 +144,34 @@ constexpr rocblas_side char2rocblas_side(char value)
...
 }
 
 inline rocblas_datatype string2rocblas_datatype(const std::string& value)
 {
     // clang-format off
     return
-        value == "f16_r" ? rocblas_datatype_f16_r :
-        value == "f32_r" ? rocblas_datatype_f32_r :
-        value == "f64_r" ? rocblas_datatype_f64_r :
-        value == "f16_c" ? rocblas_datatype_f32_c :
```
leekillough: Typo

library/src/include/utility.h
```diff
@@ -96,7 +96,7 @@ constexpr auto rocblas_datatype_string(rocblas_datatype type)
     case rocblas_datatype_f16_r: return "f16_r";
     case rocblas_datatype_f32_r: return "f32_r";
     case rocblas_datatype_f64_r: return "f64_r";
-    case rocblas_datatype_f16_c: return "f16_k";
```
leekillough: Typo

clients/include/test_cleanup.hpp
```diff
@@ -18,9 +18,6 @@ class test_cleanup
     }
 
     public:
-    // Push a cleanup handler on the stack
```
leekillough: This function is only ever used once, so just inline it there.

clients/include/rocblas_datatype2string.hpp
```diff
@@ -129,42 +144,34 @@ constexpr rocblas_side char2rocblas_side(char value)
     }
 }
 
-constexpr rocblas_datatype char2rocblas_datatype(char value)
```
leekillough: We no longer need to convert `char` to `rocblas_datatype` anywhere -- it's all done with `string` now.

clients/include/rocblas_arguments.hpp
```diff
@@ -15,11 +15,6 @@
 #include "rocblas.h"
 #include "rocblas_datatype2string.hpp"
 
-typedef enum rocblas_initialization_ {
-    rocblas_initialization_random_int = 111,
-    rocblas_initialization_trig_float = 222,
-} rocblas_initialization;
-
```
leekillough: The `rocblas_initialization` type needs to go into `rocblas_datatype2string.hpp` since that file is included by `rocblas_arguments.hpp` and it needs `rocblas_initialization` to do the conversions.

PR #492, opened 2019-02-26
adding performance scripts for dtrsm hpl

yoichiyoshida:
missing dtrsm list 2 does not have padded lda for 384 as not provided


mahmoodw:
LGTM

PR #497, opened 2019-02-28
improve Doxygen documentation

amcamd:
- remove duplicate Doxygen information from files other than:
  - library/include/rocblas.h
  - library/include/rocblas-functions.h
  - library/include/rocblas-types.h
  - library/include/rocblas-auxiliary.h
- add missing Doxygen information in the above files


library/include/rocblas-auxiliary.h
```diff
@@ -14,80 +13,64 @@
...
-ROCBLAS_EXPORT rocblas_pointer_mode rocblas_pointer_to_mode(void* ptr);
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-/********************************************************************************
- * \brief rocblas_handle is a structure holding the rocblas library context.
- * It must be initialized using rocblas_create_handle()
- * and the returned handle must be passed
```
saadrahim: Are we loosing some detail here?

amcamd: This information is in the file   library/include/rocblas-types.h  .  I think it is better to have the information in library/include/rocblas-types.h ,  and not have it duplicated in both files.

PR #508, opened 2019-03-11
fix for bug swdev-180337. can not build with ROCm2.2 RC1.

wbgilmartin:
resolves swdev-180337

added call to Tensile initializer in the rocblas-create-handle function.


library/src/rocblas_auxiliary.cpp
```diff
@@ -71,6 +74,10 @@ extern "C" rocblas_status rocblas_create_handle(rocblas_handle* handle)
     try
     {
         *handle = new _rocblas_handle();
+
+        static std::once_flag tensileIntiFlag; 
+        std::call_once(tensileIntiFlag, tensileInitialize);
+
```
leekillough: Instead of using `std::call_once()`,  a simpler and faster method is to do:
```
static int dummy = (tensileInitialize(), 0);
```
You do not need `std::call_once`, `#include <thread>`, or `#include <mutex>` anymore.

In C++11 this is guaranteed to be thread-safe.

library/src/rocblas_auxiliary.cpp
```diff
@@ -71,6 +72,9 @@ extern "C" rocblas_status rocblas_create_handle(rocblas_handle* handle)
     try
     {
         *handle = new _rocblas_handle();
+
+        static int dummy = (tensileInitialize(), 0);
+
```
bragadeesh: Please move the Tensile init before rocblas handle; Tensile being independent component need to be initialized first

PR #514, opened 2019-03-13
Updating to rocm2.2

saadrahim:
Jenkinsfile needs to run on rocm2.2


amcamd:
The changes look good. Can you confirm that it runs CI on both gfx900 and gfx906. The CI build or test may fail due to running on ROCm 2.2, and we will need to fix it.

PR #519, opened 2019-03-15
Refactor of Jenkinsfile; switching to use shared library

saadrahim:
Summary of proposed changes:
-  Only project specific Jenkins configuration is stored in the project repository
-  Jenkins shared library at https://github.com/ROCmSoftwarePlatform/rocJenkins stores the common configuration including rocm version and docker image names. We will be able to centrally update the versioning. Some flexibility is lost but we gain easier management. 
- Unused features in the current Jenkinsfile was not implemented in the shared library. New features will be added in the upcoming weeks.
 - Shared library was designed to support Jenkins Blue Ocean UI in the future.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,12 @@
 #!/usr/bin/env groovy
+// This shared library is available at https://github.com/ROCmSoftwarePlatform/rocJENKINS/
+@Library('rocJenkins') _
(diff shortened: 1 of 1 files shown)


saadrahim:
I will leave this open overnight to test the nightly runs and merge later this weekend.

PR #524, opened 2019-03-25
Env variable

mahmoodw:
resolves #172044

Summary of proposed changes:
-  Introduces WORKBUF_TRSM_B_CHNK environment variable to let user decide amount of pre-allocated handle memory
-  Replace hipMemsetAsync with rocblas_tritri_batched_fill


leekillough:
We need to get a handle on this device memory allocation problem, and not simply supply ad hoc one-off solutions for a fire.

The solution should be BLAS-wide, not specific to TRSM, and thus not use TRSM in the environment variable name. It should be general and flexible. It should not affect the API when practical, since it's an out-of-band issue which should use out-of-band solutions.

Adding this environment variable locks us down further into a particular solution, making it harder to backtrack later when designing the correct solution BLAS-wide.

See SWDEV-180330.

PR #538, opened 2019-04-11
avoid NaN propagation in gemv when beta==0

amcamd:
resolves swdev-178779

- make gemv comply with reference implementation by not allowing NaN in y vector to propagate when beta = 0


bragadeesh:
@amcamd , can you also add a testcase for this, thanks

leekillough:
> 
> 
> @amcamd , can you also add a testcase for this, thanks

See https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/539

amcamd:
Confirmed that gemv tests in #539 pass, so this PR can be merged

PR #539, opened 2019-04-12
Add NaN tests to GEMV and GEMM*

leekillough:
Adds NaN tests for gemv, gemm, and gemm_ex.


leekillough:
> 
> 
> install.sh changes LGTM.

Moved to #540 since it's really separate issue and we're still seeing NaN failures with this PR.

PR #547, opened 2019-05-03
add logic for mlperf transform sizes

wbgilmartin:
resolves SWDEV-182072

This updates the logic files which includes tuning for the transformation sizes in swdev-182072.

We see performance gains in most of the cases ranging from a modest 4% gains to 20% gain for the non strided cases and over 200% for the strided cases.

most of the lager sizes a showing efficiency above 85% with peek performance 93.2%


mahmoodw:
LGTM 
* I need to fix the massage script and some files after the merge to avoid duplicate LdcEqualsLdd flags

PR #549, opened 2019-05-06
Fix use of variable templates in header

mhbliao:
- Add necessary `static` specifier to avoid multiple definition.

resolves SWDEV-188303

Summary of proposed changes:
- when variable template is used in header, it's necessary to add `static` specifier to avoid multiple definition errors during link if that header is included/used in multiple sources.
-  
-  
-

### DIFF
--- a/library/src/blas1/rocblas_amax_amin.h
+++ b/library/src/blas1/rocblas_amax_amin.h
@@ -98,15 +98,16 @@ struct rocblas_finalize_amax_amin
 };
 
 template <typename>
-constexpr char rocblas_iamaxmin_name[] = "unknown";
+static constexpr char rocblas_iamaxmin_name[] = "unknown";
 template <>
-constexpr char rocblas_iamaxmin_name<float>[] = "rocblas_isa" QUOTE(MAX_MIN);
+static constexpr char rocblas_iamaxmin_name<float>[] = "rocblas_isa" QUOTE(MAX_MIN);
 template <>
-constexpr char rocblas_iamaxmin_name<double>[] = "rocblas_ida" QUOTE(MAX_MIN);
+static constexpr char rocblas_iamaxmin_name<double>[] = "rocblas_ida" QUOTE(MAX_MIN);
 template <>
-constexpr char rocblas_iamaxmin_name<rocblas_float_complex>[] = "rocblas_ica" QUOTE(MAX_MIN);
+static constexpr char rocblas_iamaxmin_name<rocblas_float_complex>[] = "rocblas_ica" QUOTE(MAX_MIN);
 template <>
-constexpr char rocblas_iamaxmin_name<rocblas_double_complex>[] = "rocblas_iza" QUOTE(MAX_MIN);
+static constexpr char
+    rocblas_iamaxmin_name<rocblas_double_complex>[] = "rocblas_iza" QUOTE(MAX_MIN);
 
 // allocate workspace inside this API
 template <typename To, typename Ti>
--- a/library/src/include/utility.h
+++ b/library/src/include/utility.h
@@ -112,22 +112,22 @@ constexpr auto rocblas_datatype_string(rocblas_datatype type)
 }
 
 // return precision string for data type
-template <typename> constexpr char rocblas_precision_string                [] = "invalid";
-template <> constexpr char rocblas_precision_string<rocblas_half          >[] = "f16_r";
-template <> constexpr char rocblas_precision_string<float                 >[] = "f32_r";
-template <> constexpr char rocblas_precision_string<double                >[] = "f64_r";
-template <> constexpr char rocblas_precision_string<int8_t                >[] = "i8_r";
-template <> constexpr char rocblas_precision_string<uint8_t               >[] = "u8_r";
-template <> constexpr char rocblas_precision_string<int32_t               >[] = "i32_r";
-template <> constexpr char rocblas_precision_string<uint32_t              >[] = "u32_r";
-template <> constexpr char rocblas_precision_string<rocblas_float_complex >[] = "f32_c";
-template <> constexpr char rocblas_precision_string<rocblas_double_complex>[] = "f64_c";
+template <typename> static constexpr char rocblas_precision_string                [] = "invalid";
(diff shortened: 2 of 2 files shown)


mhbliao:
updated after re-formatting

leekillough:
3 A name having namespace scope (3.3.6) has internal linkage if it is the name of
(3.1) — a variable, function or function template that is explicitly declared static; or,
(3.2) — a non-volatile variable that is explicitly declared `const` or `constexpr` and neither explicitly declared `extern` nor previously declared to have external linkage; or
(3.3) — a data member of an anonymous union.

leekillough:
From C++11:

3 A name having namespace scope (3.3.6) has internal linkage if it is the name of
(3.1) — a variable, function or function template that is explicitly declared static; or,
(3.2) — a non-volatile variable that is explicitly declared `const` or `constexpr` and neither explicitly declared `extern` nor previously declared to have external linkage; or
(3.3) — a data member of an anonymous union.

In C++11 there was a defect that `constexpr` namespace variables were not implicitly internal linkage just like const-qualified variables were, which was retroactively fixed in C++11:

http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1112

C++14 added variable templates, but, by leaving the wording above alone, left it ambiguous whether const-qualified variable templates were internal linkage or not. **Most implementations made const-qualified variable templates internal linkage.**

C++17 adds `inline` variables. `constexpr` static member variables, but not `constexpr` namespace variables, are implicitly `inline`. The reason namespace `constexpr` variables are not `inline` by default is that `inline` on a variable without `static` specifies external linkage, even for const-qualified variables, and `inline` overrides the implicit internal linkage of a const-qualified namespace variable, making it `extern` unless it's also declared `static`.

    int x = 1;                                         // external linkage
    static int x =1;                                   // internal linkage
    extern int x =1;                                   // not allowed
    const int x = 1;                                   // internal linkage
    static const int x = 1;                            // internal linkage
    extern const int x = 1;                            // external linkage

    constexpr int x = 1;                               // internal linkage
    static constexpr int x = 1;                        // internal linkage   
    extern constexpr int x = 1;                        // external linkage

    inline int x = 1;                                  // external linkage (C++17 and later)
    inline extern int x = 1;                           // not allowed
    inline static int x = 1;                           // internal linkage 
    inline const int x  =1;                            // external linkage
    inline static const int x = 1;                     // internal linkage
    inline extern const int x = 1;                     // external linkage

    inline constexpr int x = 1;                        // external linkage
    inline static constexpr int x = 1;                 // internal linkage
    inline extern constexpr int x = 1;                 // external linkage

    template<class> int x = 1;                         // external linkage
    template<class> static int x = 1;                  // internal linkage
    template<class> extern int x = 1;                  // not allowed
    template<class> const int x = 1;                   // ambiguous linkage until C++20
    template<class> static const int x =1;             // internal linkage
    template<class> extern const int x = 1;            // external linkage

    template<class> constexpr int x = 1;               // ambiguous linkage until C++20
    template<class> static constexpr int x = 1;        // internal linkage
    template<class> extern constexpr int x = 1;        // external linkage

    template<class> inline int x = 1;                  // external linkage
    template<class> inline extern int x = 1;           // not allowed
    template<class> inline static int x = 1;           // internal linkage
    template<class> inline const int x =1;             // external linkage
    template<class> inline static const int x = 1;     // internal linkage
    template<class> inline extern const int x = 1;     // external linkage
   
    template<class> inline constexpr int x = 1;        // external linkage
    template<class> inline static constexpr int x=1;   // internal linkage
    template<class> inline extern constexpr int x = 1; // external linkage

A very recent (Feburary 2019) issue was raised on this:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1359r0.html#2387

**2387. Linkage of const-qualified variable template**

**Section:** 6.5  [basic.link]    **Status: tentatively ready**     **Submitter:** John Spicer     **Date:** 2018-09-28

The list in 6.5 [basic.link] paragraph 3 specifying which entities receive internal linkage does not mention variable templates, so presumably a variable template has external linkage. 12 [temp] paragraph 6 gives the impression that a specialization of a template with external linkage also has external linkage. **However, current implementations appear to give internal linkage to specializations of const-qualified variable templates. Should const-qualified variable templates have internal linkage?** *[emphasis mine]*

**Notes from the December, 2018 teleconference:**

CWG felt that a `const` type should not affect the linkage of a variable template or its instances.

**Proposed resolution (February, 2019):**

Change 6.5 [basic.link] paragraph 3 as follows:

A name having namespace scope (6.3.6 [basic.scope.namespace]) has internal linkage if it is the name of

 *   a variable, **variable template,** function or function template that is explicitly declared static; or,

 *   a non-inline **non-template** variable of non-volatile const-qualified type that is neither explicitly d eclared extern nor previously declared to have external linkage; or

 *  a data member of an anonymous union.

**[Note: An instantiated variable template that has const-qualified type can have external linkage, even if not declared extern. —end note]**

So, to summarize:

1. In C++17 the committee decided that `inline` overrides `const`/`constexpr` when determining default linkage. `inline` without `static` implies `extern`, even on const-qualified variables.
2. A C++14 oversight left ambiguous whether `const`/`constexpr` variable templates have internal linkage by default, like non-template `const`/`constexpr` variables do. Most compilers made `const`/`constexpr` variable templates have internal linkage, unless declared `inline` or `extern`.
3. Rather than fix variable template linkage behavior to be consistent with non-template variables, and thus make `const`/`constexpr` variable templates have internal linkage by default (unless overridden by `inline` or `extern`), the committee has decided to make them external linkage by default. (This seems like the wrong decision, IMO.)

Since we are not enabling C++17 yet, we cannot use `inline`. That basically leaves `static` .

However, there are probably more changes needed in the library than just these. I will need to make a full sweep.

leekillough:
clang changes which correspond to this change:

https://github.com/llvm/llvm-project/search?q=DR2387&type=Commits

leekillough:
Please refer to #551

yxsamliu:
can we merge this fix into master branch? This issue is blocking HIP-Clang CI. Thanks.

leekillough:
See #551.

After it passes and merges, we can cherry-pick it into master.

Tonight is the 2.5 release, so getting it in develop is higher priority.

PR #550, opened 2019-05-06
add tuning logic for small sizes

wbgilmartin:
resolves SWDEV-187235 and SWDEV-184150

Add tuning logic related to the small sizes in SWDEV-187235 as well as the sizes in SWDEV-184150

overall these was significant performance improvement across most of the sizes described in the task however due to the fact that these are mostly small sizes that are difficult to tune for the over all efficiency for most of these sizes is > 50%.

-  
-  
-


mahmoodw:
LGTM
* I still need to fix the massage script and some files after the merge to avoid duplicate LdcEqualsLdd flags

mahmoodw:
Will need to redo logic fille massage process since the develop archive files have changed since the last commit.

PR #559, opened 2019-05-15
Switch to Python 3 for virtual environment.

sdquiring:
Required for the switch away from Python 2 in Tensile.


sdquiring:
Good point.  Virtualenv installs the TensileCreateLibrary script inside the root bin directory, we can use that instead.

I don't believe we tried Tensile with any python 3 earlier than 3.5.

PR #562, opened 2019-05-21
trsv implementation

TonyYHsieh:
1. use gemv to implement trsv
2. add trsv_ex function

resolves #___

Summary of proposed changes:
-  some of our users didn't accept to use trsm to implement trsv function.
-  We implement trsv function by using gemv function. We are reference to trsm function and replace gemm by gemv to achieve trsv function which has better performance.
-  We also implement trsv_ex and trsv_ex testcase, all implementation are reference to trsm implementation.

### DIFF
--- a/library/src/blas2/gemv.hpp
+++ b/library/src/blas2/gemv.hpp
@@ -1,0 +1,61 @@
+/* ************************************************************************
+ * Copyright 2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#pragma once
+#ifndef _GEMV_HPP_
+#define _GEMV_HPP_
+
+#include <hip/hip_runtime.h>
+
+template <typename T>
+rocblas_status rocblas_gemv_template(rocblas_handle handle,
+                                     rocblas_operation transA,
+                                     rocblas_int m,
+                                     rocblas_int n,
+                                     const T* alpha,
+                                     const T* A,
+                                     rocblas_int lda,
+                                     const T* x,
+                                     rocblas_int incx,
+                                     const T* beta,
+                                     T* y,
+                                     rocblas_int incy);
+
+template <>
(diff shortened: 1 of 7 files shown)


clients/benchmarks/client.cpp
```diff
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright 2016 Advanced Micro Devices, Inc.
+ * Copyright 2019 Advanced Micro Devices, Inc.
```
saadrahim: I don't know our policy but I thought copyright is supposed to 2016-2019

TonyYHsieh: update PR for copywrite format

TonyYHsieh:
> The is no need to add TRSV_EX.
> 
> A 3-month-memory allocation scheme for all of rocBLAS has been designed and it will replace TRSM_EX and TRSV_EX

imcarsonliao:
> > The is no need to add TRSV_EX.
> > A 3-month-memory allocation scheme for all of rocBLAS has been designed and it will replace TRSM_EX and TRSV_EX
> 
> So what should I do for this change??
> Can we keep both TRSV_EX and TRSM_EX before new scheme??

Tony, please remove trsv_ex export API, while internally calling to "ex" function is fine.

TonyYHsieh:
> > > The is no need to add TRSV_EX.
> > > A 3-month-memory allocation scheme for all of rocBLAS has been designed and it will replace TRSM_EX and TRSV_EX
> > 
> > 
> > So what should I do for this change??
> > Can we keep both TRSV_EX and TRSM_EX before new scheme??
> 
> Tony, please remove trsv_ex export API, while internally calling to "ex" function is fine.

Hi imcarsonliao,

I have updated commit to remove trsv_ex API.
Please help to review my update.

Thanks,
Tony.

PR #569, opened 2019-05-30
fix pointer declaration in bfloat16 sample

amcamd:
- correct pointer declaration changing from float* to rocblas_bfloat16*


TejashShah:
LGTM. Thanks Andrew for taking care of it.

PR #570, opened 2019-05-30
formatting using clang-format-9

akilaMD:
Summary of proposed changes:
-  Changed clang formatter to /opt/rocm/hcc/bin/clang-format, which is clang-format-9
-  Passes CI
- This fixes the issue that Michael mentioned when running the format check on Ubuntu-18.04
- Rearranges the order of include statements
- Uses .clang-format file that is in line with Google C++ style guide standards
- Jenkinsfile now uses the clang9 branch of rocJenkins instead of the master branch


leekillough:
If and when this is merged, "Squash and merge" should be used. We don't want 25+ commits in the develop branch.

PR #573, opened 2019-06-04
Trsm: use template matrix copy to replace void matrix copy

TonyYHsieh:
resolves #___

Summary of proposed changes:
-  optimize performance of one of trsm kernel copy_void_ptr_matrix_trsm.
-  copy matrix use template to decide type is faster than void copy with size of type
   In our test case:  WORKBUF_TRSM_B_CHNK=32768 ./clients/staging/rocblas-bench -r d -f trsm -m 256 --lda 256 -n 32768 --ldb 256 --side L --uplo L --transposeA N --diag U -v 1 -i 10, origin copy_void_ptr_matrix_trsm takes 739us, new copy_matrix_trsm takes 384us.
   
original trace:
image

new trace:
image

### DIFF
--- a/library/src/blas3/rocblas_trsm.cpp
+++ b/library/src/blas3/rocblas_trsm.cpp
@@ -640,30 +640,29 @@ rocblas_status rocblas_trsm_right(rocblas_handle handle,
     return rocblas_status_success;
 }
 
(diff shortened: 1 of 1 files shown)


mahmoodw:
Good find! Looks good to me.

PR #574, opened 2019-06-04
add bfloat16 test that will fail for ieee16 and pass for ieee32

amcamd:
- modify bfloat16 test so that it will fail if IEEE16 bit arithmetic is used, and it will pass if IEEE32 bit arithmetic is used

- Full rocBLAS tests pass on gfx900 and gfx906

### DIFF
--- a/clients/include/testing_gemm_ex.hpp
+++ b/clients/include/testing_gemm_ex.hpp
@@ -396,7 +396,7 @@ void testing_gemm_ex(const Arguments& arg)
     if(std::is_same<To, rocblas_half>{} && std::is_same<Tc, float>{})
     {
         // half precision IEEE has max and lowest values 65504 and -65504,
-        // foat precision IEEE has max and lowest values 3.403e+38 and -3.403e+38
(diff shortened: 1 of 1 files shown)


amcamd:
- The check is to verify that IEEE32 bit arithmetic not IEEE16 bit arithmetic is used. 
- Another check will need to be added to verify that IEEE32 bit arithmetic not bfloat16 arithmetic is used.

clients/include/testing_gemm_ex.hpp
```diff
@@ -421,6 +421,34 @@ void testing_gemm_ex(const Arguments& arg)
             hB[ldb + 1] = static_cast<Ti>(positive_two);
         }
     }
+    else if(std::is_same<Ti, rocblas_bfloat16>{} && std::is_same<Tc, float>{})
+    {
+        // half precision IEEE has max and lowest values 65504 and -65504,
+        // foat precision IEEE has max and lowest values 3.403e+38 and -3.403e+38
```
FelixSchwarz: typo ("foat")

amcamd:
Thanks @FelixSchwarz ,  fixed typo.

PR #581, opened 2019-06-13
fixed sorted headers in client.cpp

akilaMD:
Summary of proposed changes:
-  Re-arranged includes to ensure the formatter doesn't reorder them
- Passed on local Ubuntu16 Jenkins instance
- Ubuntu18 Jenkins job in-progress

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -70,7 +70,7 @@ template <typename Ti, typename To, typename Tc>
 struct perf_gemm_ex<Ti,
                     To,
                     Tc,
-                    typename std::enable_if<!std::is_same<Ti, void> {} && !is_complex<Ti>>::type>
+                    typename std::enable_if<!std::is_same<Ti, void>{} && !is_complex<Ti>>::type>
 {
     explicit operator bool()
     {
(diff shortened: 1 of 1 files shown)


akilaMD:
@zaliu Give this a try when you get the chance and let me know if compilation still fails.

clients/benchmarks/client.cpp
```diff
@@ -57,7 +70,7 @@ template <typename Ti, typename To, typename Tc>
 struct perf_gemm_ex<Ti,
                     To,
                     Tc,
-                    typename std::enable_if<!std::is_same<Ti, void>{} && !is_complex<Ti>>::type>
+                    typename std::enable_if<!std::is_same<Ti, void> {} && !is_complex<Ti>>::type>
```
leekillough: The extra spaces should not have been added back.
Are the git hooks using clang 9?

akilaMD: The git hooks are using clang 9...not sure what to make of that. Not sure what happened. I've fixed the change in my 2nd commit.

akilaMD:
> Two nightly tests are failing on both gfx900 and gfx906, and the same tests are passing in the develop branch for gfx900. Can you please find out why the tests are failing, and resolve the failure before merging.

Sure, I'll look into that right away

PR #583, opened 2019-06-18
complex gemm implementation

TonyYHsieh:
resolves #___

Summary of proposed changes:
-  cgemm
-  zgemm

This is first version for complex gemm.
We extract element in complex and do float/double
gemm seperately due to lack of complex kernel
in Tensile. We will replace it after Tensile
complex kernel is ready.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -67,10 +67,7 @@ struct perf_gemm_ex : rocblas_test_invalid
 };
 
 template <typename Ti, typename To, typename Tc>
-struct perf_gemm_ex<Ti,
-                    To,
-                    Tc,
-                    typename std::enable_if<!std::is_same<Ti, void>{} && !is_complex<Ti>>::type>
+struct perf_gemm_ex<Ti, To, Tc, typename std::enable_if<!std::is_same<Ti, void>{}>::type>
 {
     explicit operator bool()
     {
@@ -90,11 +87,10 @@ struct perf_gemm_strided_batched_ex : rocblas_test_invalid
 };
 
 template <typename Ti, typename To, typename Tc>
-struct perf_gemm_strided_batched_ex<
-    Ti,
-    To,
-    Tc,
-    typename std::enable_if<!std::is_same<Ti, void>{} && !is_complex<Ti>>::type>
+struct perf_gemm_strided_batched_ex<Ti,
+                                    To,
+                                    Tc,
+                                    typename std::enable_if<!std::is_same<Ti, void>{}>::type>
 {
     explicit operator bool()
     {
@@ -111,6 +107,25 @@ struct perf_gemm_strided_batched_ex<
 template <typename T, typename = void>
 struct perf_blas : rocblas_test_invalid
 {
+};
+
+template <typename T>
+struct perf_blas<T, typename std::enable_if<is_complex<T>>::type>
+{
+    explicit operator bool()
+    {
+        return true;
+    }
+    void operator()(const Arguments& arg)
+    {
+        if(!strcmp(arg.function, "gemm"))
+            testing_gemm<T>(arg);
+        else if(!strcmp(arg.function, "gemm_strided_batched"))
+            testing_gemm_strided_batched<T>(arg);
+        else
+            throw std::invalid_argument("Invalid combination --function "s + arg.function
+                                        + " --a_type "s + rocblas_datatype2string(arg.a_type));
+    }
 };
 
 template <typename T>
--- a/clients/common/norm.cpp
+++ b/clients/common/norm.cpp
@@ -53,13 +53,9 @@ void zaxpy_(int*                    n,
  * speciliazation ======================================= */
 
 /*! \brief compare the norm error of two matrices hCPU & hGPU */
-template <>
-double norm_check_general<rocblas_bfloat16>(char              norm_type,
-                                            rocblas_int       M,
-                                            rocblas_int       N,
-                                            rocblas_int       lda,
-                                            rocblas_bfloat16* hCPU,
-                                            rocblas_bfloat16* hGPU)
+template <typename T>
+double norm_check_general(
+    char norm_type, rocblas_int M, rocblas_int N, rocblas_int lda, T* hCPU, T* hGPU)
 {
     // norm type can be 'O', 'I', 'F', 'o', 'i', 'f' for one, infinity or Frobenius norm
     // one norm is max column sum
@@ -68,25 +64,26 @@ double norm_check_general<rocblas_bfloat16>(char              norm_type,
 
     double error_double = std::numeric_limits<double>::quiet_NaN();
 
-    host_vector<float> hCPU_float(N * lda), hGPU_float(N * lda);
+    host_vector<double> hCPU_double(N * lda);
+    host_vector<double> hGPU_double(N * lda);
+
     for(rocblas_int i = 0; i < N * lda; i++)
     {
-        hCPU_float[i] = static_cast<float>(hCPU[i]);
-        hGPU_float[i] = static_cast<float>(hGPU[i]);
+        hCPU_double[i] = static_cast<double>(hCPU[i]);
+        hGPU_double[i] = static_cast<double>(hGPU[i]);
     }
 
-    float       work;
+    double      work[1];
     rocblas_int incx  = 1;
-    float       alpha = -1.0f;
+    double      alpha = -1.0;
     rocblas_int size  = lda * N;
 
-    float cpu_norm = slange_(&norm_type, &M, &N, hCPU_float, &lda, &work);
-    saxpy_(&size, &alpha, hCPU_float, &incx, hGPU_float, &incx);
+    double cpu_norm = dlange_(&norm_type, &M, &N, hCPU_double.data(), &lda, work);
+    daxpy_(&size, &alpha, hCPU_double.data(), &incx, hGPU_double.data(), &incx);
 
-    float error_float = slange_(&norm_type, &M, &N, hGPU_float, &lda, &work) / cpu_norm;
-    error_double      = static_cast<double>(error_float);
+    double error = dlange_(&norm_type, &M, &N, hGPU_double.data(), &lda, work) / cpu_norm;
 
-    return error_double;
+    return error;
 }
 
 template <>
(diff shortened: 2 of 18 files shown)


library/src/blas_ex/rocblas_gemm_ex.cpp
```diff
@@ -17,6 +17,24 @@
...
+{
+    return val.x;
+}
+
+template <>
+double to_double<rocblas_double_complex>(const rocblas_double_complex& val)
+{
+    return val.x;
+}
+
```
leekillough: Smaller, more generic code **(but see below about `alpha`, `beta` being complex)**:
```
template <typename T, typename = void>
double to_double(const T& val)
{
    return val;
}

template<typename T, typename std::enable_if<is_complex<T>>::type>
double to_double(const T& val)
{
    return val.x;
}
```

library/src/blas_ex/rocblas_gemm_ex.cpp
```diff
@@ -94,6 +112,16 @@ extern "C" rocblas_status rocblas_gemm_ex(rocblas_handle    handle,
...
+                else if(compute_type == rocblas_datatype_f32_c)
+                {
+                    alpha_double = to_double(*static_cast<const rocblas_float_complex*>(alpha));
+                    beta_double  = to_double(*static_cast<const rocblas_float_complex*>(beta));
+                }
+                else if(compute_type == rocblas_datatype_f64_c)
+                {
+                    alpha_double = to_double(*static_cast<const rocblas_double_complex*>(alpha));
+                    beta_double  = to_double(*static_cast<const rocblas_double_complex*>(beta));
+                }
```
leekillough: In complex GEMM, alpha and beta are supposed to be complex, most generally speaking.

So they cannot simply be converted to double, and have their imaginary part ignored.

clients/include/rocblas_random.hpp
```diff
@@ -110,6 +126,22 @@ inline int8_t random_generator<int8_t>()
     return std::uniform_int_distribution<int8_t>(1, 3)(rocblas_rng);
 };
 
+template <>
+inline rocblas_float_complex random_generator<rocblas_float_complex>()
+{
+    float real  = std::uniform_int_distribution<int>(1, 10)(rocblas_rng);
+    float image = std::uniform_int_distribution<int>(1, 10)(rocblas_rng);
+    return {real, image};
```
leekillough: `image` should be spelled `imag`.

`imag` is short for `imaginary`. There is no `e` in `imaginary`.

library/src/include/rocblas_unique_ptr.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef GUARD_ROCBLAS_MANAGE_PTR_HPP
 #define GUARD_ROCBLAS_MANAGE_PTR_HPP
 
+#include "definitions.h"
```
leekillough: `rocblas_unique_ptr.hpp` will be removed with the device memory allocation refactoring, currently in-progress.

We do not need to use `unique_ptr` wrappers for `hipMalloc` and `hipFree` anymore, since we fully manage device memory inside of `rocblas_handle` now.

clients/include/testing_gemm_strided_batched_ex.hpp
```diff
@@ -319,7 +319,9 @@ void testing_gemm_strided_batched_ex(const Arguments& arg)
         h_alpha_Tc = float_to_half(arg.alpha);
         h_beta_Tc  = nantest ? 0 : float_to_half(arg.beta);
     }
-    else if(std::is_same<Tc, float>{} || std::is_same<Tc, double>{} || std::is_same<Tc, int32_t>{})
+    else if(std::is_same<Tc, float>{} || std::is_same<Tc, double>{} || std::is_same<Tc, int32_t>{}
+            || std::is_same<Tc, rocblas_float_complex>{}
+            || std::is_same<Tc, rocblas_double_complex>{})
```
leekillough: `is_complex<Tc>` can be used instead of `std::is_same<Tc, rocblas_float_complex>{} || std::is_same<Tc, rocblas_double_complex>{}`

clients/include/rocblas_random.hpp
```diff
@@ -110,6 +126,22 @@ inline int8_t random_generator<int8_t>()
...
+};
+
+template <>
+inline rocblas_double_complex random_generator<rocblas_double_complex>()
+{
+    double real  = std::uniform_int_distribution<int>(1, 10)(rocblas_rng);
+    double image = std::uniform_int_distribution<int>(1, 10)(rocblas_rng);
+    return {real, image};
```
leekillough: `image` should be `imag`; see similar comment below.

clients/benchmarks/client.cpp
```diff
@@ -113,6 +109,27 @@ struct perf_blas : rocblas_test_invalid
 {
 };
 
+template <typename T>
+struct perf_blas<T,
+                 typename std::enable_if<std::is_same<T, rocblas_float_complex>{}
+                                         || std::is_same<T, rocblas_double_complex>{}>::type>
```
leekillough: `typename std::enable_if<is_complex<T>>::type` can be used instead of `typename std::enable_if<std::is_same<T, rocblas_float_complex>{} || std::is_same<T, rocblas_double_complex>{}>::type`

clients/benchmarks/client.cpp
```diff
@@ -113,6 +109,27 @@ struct perf_blas : rocblas_test_invalid
...
+        if(!strcmp(arg.function, "gemm"))
+            testing_gemm<T>(arg);
+        else if(!strcmp(arg.function, "gemm_strided_batched"))
+            testing_gemm_strided_batched<T>(arg);
+        else
+            throw std::invalid_argument("Invalid combination --function "s + arg.function
+                                        + " --a_type "s + rocblas_datatype2string(arg.a_type));
+    }
+};
+
```
leekillough: Rather than write a separate specialization for complex, the `enable_if` for the current real `gemm` should be expanded to support complex by adding `|| is_complex<T>` to its enabling condition. Then you do not need to duplicate code.

library/src/blas3/Tensile/gemm.h
```diff
@@ -92,3 +92,251 @@ inline rocblas_status validateArgs(rocblas_handle    handle,
...
+    dst = dst + hipBlockIdx_z * str_dst;
+    src = src + hipBlockIdx_z * str_src;
+
+    if(tx < m && ty < n)
+        dst[tx + ld_dst * ty] = src[tx + ld_src * ty].x;
+}
+
+template <typename DT, typename ST>
+__global__ void kernel_copy_real_image_of_complex_matrix(DT*               dst,
+                                                         const rocblas_int ld_dst,
```
leekillough: `imag` and `image` mean different things here. `copy_real_image` suggests a projection, while `copy_real_imag` suggests copying both `real` and `imag` parts. I suggest either `kernel_copy_complex_matrix` or `kernel_copy_real_part_of_complex_matrix` as names with clearer meanings.

clients/common/norm.cpp
```diff
@@ -506,6 +506,92 @@ double norm_check_general<double>(char        norm_type,
...
+        }
+        else if(norm_type == 'O' || norm_type == 'o' || norm_type == 'I' || norm_type == 'i')
+        {
+            cumulative_error = cumulative_error > error ? cumulative_error : error;
+        }
+    }
+
+    return cumulative_error;
+}
+
```
leekillough: It would be better if most of these norm functions could be consolidated and differentiated where necessary, instead of having the same code multiplied several times.

clients/include/rocblas_math.hpp
```diff
@@ -20,7 +20,14 @@ inline __host__ rocblas_half float_to_half(float val)
 }
 
 // Helper routine to convert halfs into their floats equivalent; uses F16C instructions
-inline __host__ float half_to_float(rocblas_half val)
+template <typename T>
+inline __host__ float half_to_float(T val)
+{
+    throw rocblas_status_internal_error;
```
leekillough: We do not throw exceptions in test code right now.

Maybe this was accidentally left in during debugging.

clients/include/rocblas_random.hpp
```diff
@@ -75,6 +75,22 @@ class rocblas_nan_rng
     {
         return random_nan_data<rocblas_bfloat16, uint16_t, 7, 8>();
     }
+
+    explicit operator rocblas_float_complex()
+    {
+        rocblas_float_complex ret{};
+        ret.x = random_nan_data<float, uint32_t, 23, 8>();
+        ret.y = random_nan_data<float, uint32_t, 23, 8>();
+        return ret;
```
leekillough: Simpler single statement:

```
return { float(*this), float(*this) };
```

clients/include/rocblas_random.hpp
```diff
@@ -75,6 +75,22 @@ class rocblas_nan_rng
...
+    }
+
+    explicit operator rocblas_double_complex()
+    {
+        rocblas_double_complex ret{};
+        ret.x = random_nan_data<double, uint64_t, 52, 11>();
+        ret.y = random_nan_data<double, uint64_t, 52, 11>();
+        return ret;
+    }
```
leekillough:
```
return { double(*this), double(*this) };
```

library/src/blas3/Tensile/gemm.h
```diff
@@ -92,3 +92,251 @@ inline rocblas_status validateArgs(rocblas_handle    handle,
...
+    dst = dst + hipBlockIdx_z * str_dst;
+    src = src + hipBlockIdx_z * str_src;
+
+    if(tx < m && ty < n)
+        dst[tx + ld_dst * ty] = src[tx + ld_src * ty].y;
+}
+
+template <typename DT, typename ST>
+void copy_part_of_complex_matrix(const hipStream_t rocblas_stream,
+                                 const bool        real,
```
leekillough: `copy` which `part_of_complex_matrix`?

clients/common/norm.cpp
leekillough: In other words, use templates and make all of the types use the same `norm` code. Then where you need to differentiate between `general`, `rocblas_half`, `rocblas_bf16`, or `rocblas_float_complex`, or `rocblas_double_complex`, you can insert `if` statements (`if constexpr` in C++17) to make code conditional.

Try not to copy-and-paste code. Instead try to refactor existing code to become more general. I can help you with C++ template metaprogramming.

clients/include/rocblas_math.hpp
```diff
@@ -20,7 +20,14 @@ inline __host__ rocblas_half float_to_half(float val)
 }
 
 // Helper routine to convert halfs into their floats equivalent; uses F16C instructions
-inline __host__ float half_to_float(rocblas_half val)
+template <typename T>
+inline __host__ float half_to_float(T val)
+{
```
leekillough: `half_to_float` should not be generalized like this. It merely defers a compile-time error to a run-time error.

This is probably being done because the current templated code calls `half_to_float` in a part of the code guarded by `if` statements but always compiled in regardless of the types, which can lead to compile-time errors if complex types are used.

In C++17 we can use `constexpr if` to produce conditional code and avoid compilation errors, but since we aren't using C++17 yet, we have to use SFINAE or other techniques.

clients/include/rocblas_random.hpp
TonyYHsieh: understand. Thanks for catch this typo error. :)

clients/include/testing_gemm_strided_batched_ex.hpp
TonyYHsieh: Understand. Thanks for catch that. 👍

clients/include/rocblas_random.hpp
TonyYHsieh: understand. thanks for catch that.

clients/benchmarks/client.cpp
TonyYHsieh: understand. thanks for catch that.

clients/benchmarks/client.cpp
TonyYHsieh: Yes, I understand. But most of our function, like trsm, didn't support complex. For preventing use complex in other function, we prefer to write another specialization to handle complex. This is same as how we handle half type.

library/src/blas3/Tensile/gemm.h
TonyYHsieh: understand. thanks for catch that.

clients/include/rocblas_math.hpp
TonyYHsieh: understand. For this function, I will return val directly for non-half type. Is that ok??

library/src/blas3/Tensile/gemm.h
TonyYHsieh: I use "bool real" as a flag to indicate copy real or imag part of complex matrix.
I will modify real to is_real_part.

library/src/include/rocblas_unique_ptr.hpp
TonyYHsieh: Hi @leekillough ,

When can we get new device memory management??
We can modify this part for new mechanism.

Thanks,
Tony

PR #584, opened 2019-06-19
New clang-format for rocm2.5

akilaMD:
- Additional options are now available in clang-format, and they need to be applied to Jenkins after I upgraded CI from rocm2.4 to rocm2.5

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -75,7 +75,7 @@ ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4
 Cpp11BracedListStyle: true
-#SpaceBeforeCpp11BracedList: false
+SpaceBeforeCpp11BracedList: false
 DerivePointerAlignment: false
 ExperimentalAutoDetectBinPacking: false
 ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
--- a/.clang-format
+++ b/.clang-format
@@ -23,12 +23,14 @@ DisableFormat:   false
 Standard: Cpp11
 
 AccessModifierOffset: -4
-AlignAfterOpenBracket: true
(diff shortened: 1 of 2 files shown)


leekillough:
```
YAML:51:21: error: unknown key 'AfterCaseLabel'
    AfterCaseLabel: 'true'
                    ^~~~~~
Error reading /home/lkilloug/rocBLAS/.clang-format: Invalid argument
``

akilaMD:
> ```
> YAML:51:21: error: unknown key 'AfterCaseLabel'
>     AfterCaseLabel: 'true'
>                     ^~~~~~
> Error reading /home/lkilloug/rocBLAS/.clang-format: Invalid argument
> ```
Known issue I'm afraid. If you're on a node that uses an older version of ROCm, there are a few new clang-format file options that are incompatible and cause errors. ROCm 2.5 has an updated clang-format instance, and it has slightly different formatting results when compared to earlier versions. It's best to run the formatting script on nodes with ROCm 2.5 installed already, or update your system clang to the latest clang-format 9.0.0 version.

leekillough:
It happens with clang-9.0.0.

Please fix this regression.

akilaMD:
> It happens with clang-9.0.0.
> Please fix this regression.

I don't think I was clear in my reply above, so I'll try again to explain. Right now, clang 9.0.0 is an [in-progress release](https://clang.llvm.org/docs/ReleaseNotes.html), thus there are revisions being made from time-to-time. It currently isn't listed on the [LLVM download page](http://releases.llvm.org/download.html). It seems that ROCm2.5 contains an updated/patched clang 9.0.0 when compared to ROCm2.4, and a few additional clang-format options have been added. You're probably working on a node with an earlier ROCm version, so your .clang-format doesn't recognize the additional options added in the new file. If you use the previous .clang-format file, there will be a diff for about 5 files for rocBLAS.

PR #585, opened 2019-06-19
Adding CentOS CI support

akilaMD:
Summary of proposed changes:
-	Idea is to run a CentOS instance of rocBLAS in CI
-	CentOS7 image is more barebones than the Ubuntu base image, so additional packages need to be installed to enable CentOS support
-	Some CentOS/rpm packages have slightly different names than Ubuntu/debian packages 
-	cmake3 must be specified for centOS7, as the default cmake is cmake2. 
-	sudo privileges for running tests must be added to the Jenkinsfile to run in CentOS w/o getting a permissions issue (not sure why this is happening, group and user permissions do appear to be the same as Ubuntu)

CentOS support in general:
-	All existing Jenkins nodes can run CentOS jobs because it's being ran in a docker container
-	Currently gfx900 is set to test CentOS7 base image, but this can be changed pretty easily
-	See [this link](https://github.com/ROCmSoftwarePlatform/rocJenkins/blob/master/README.md) for more informatioN

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,7 @@ rocBLASCI:
     rocblas.paths.build_command = './install.sh -c'
 
     // Define test architectures, optional rocm version argument is available
-    def nodes = new dockerNodes(['gfx900 && centos7', 'gfx906'], rocblas)
+    def nodes = new dockerNodes(['gfx900', 'gfx906 && centos7'], rocblas)
 
     boolean formatCheck = true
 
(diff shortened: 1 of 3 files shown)


akilaMD:
@amcamd @zaliu Looks like the CentOS Vega 20 Jenkins job passed CI, same with the Ubuntu Vega 10 jenkins job. Is it ok for me to squash and merge this PR?

PR #603, opened 2019-07-03
Use git ls-files instead of find for clang-format

malcolmroberts:
Summary of proposed changes:
-  Change in documentation to use git ls-files instead of find


malcolmroberts:
Thanks for the heads-up with -z and -0; I added those.
Adding similar changes to the pre-commit hook doesn't seem to be straight-forward.

PR #609, opened 2019-07-09
Complex develop for review [don't merge due to issue SWDEV-194263]

TonyYHsieh:
resolves #___

Summary of proposed changes:
-  1. implement rocblas software flow for complex gemm
-  2. use tensile hip solution for complex gemm in rocblas
-  

Hi all,

This PR contains 2 commit.
First commit implement rocblas complex gemm software flow with dgemm tensile kernel.
Second commit implement rocblas complex gemm cal tensile complex gemm by hip solution.
Tensile part also has a PR for complex gemm. (we will update later)

Due to compile issue on HIP header, we can't compile pass with ROCm 2.5 (ROCm 2.4 is fine).
There has JIRA ticket (http://ontrack-internal.amd.com/browse/SWDEV-194263) for it. We hope that we can go thought code review before compile issue is fixed.


leekillough:
PR #607 removed the dependency on the `<hip/hip_vector_types.h>` because of a similar issue. It also ensures robust `#include` files, by making sure every header file is self-sufficient (it can compile if it is `#include`d from a `main() {}` program with no other `#include`s before it -- it `#include`s all of its dependencies first).

Also, `<hip/hip_runtime_api.h>` has a bug where it uses C++ code even when compiling in C mode, leading to a `<cstdint>` unknown header error.

PR #607 removes all of those dependencies, and makes `rocblas_float_complex` and `rocblas_double_complex` simple `struct`s. If the HIP complex data types are [standard layout](https://en.cppreference.com/w/cpp/types/is_standard_layout) (C++ term), then they should be compatible with rocBLAS complex data types.

`rocblas_unique_ptr.hpp` is going away, and there are hundreds of refactoring changes to the memory allocation system, only being held up by some mysterious `TRSM` failures which are being solved right now. Once that is checked in, 90% of the rocBLAS structure will be synchronization-free.

PR #612, opened 2019-07-11
Fix allocator bug; add one-time warning messages about degraded performance

leekillough:
Fix allocator bug for when size = 0 is requested for one or more arrays.
Add one-time warning messages when degraded performance is encountered in TRSM or TRSV.
Add `total_size()` function for querying allocated memory's total size.


mahmoodw:
LGTM

PR #615, opened 2019-07-17
Fix typo and simplify compatibility logic in rocblas_bfloat16.h; parallelize header file checking

leekillough:
Use a simpler method to ensure C/C++11/C++14 compatibility in `rocblas_bfloat16.h`.

Only if C++14 and HCC are detected enabled, does it include the `__device__` and `constexpr` code. Otherwise it uses a plain old C `struct`.

Fit a typo in the `struct`.

Parallelize the `header_compilation_tests.sh` header file validation script, reducing build times.

### DIFF
--- a/header_compilation_tests.sh
+++ b/header_compilation_tests.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+
 exec >&2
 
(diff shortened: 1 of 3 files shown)


leekillough:
All of the tests (quick, pre_checkin, nightly) passed in a standalone run.

PR #619, opened 2019-07-22
Updating version with new rocm-cmake standard

saadrahim:
resolves FBA-29

Summary of proposed changes:
-  Changing versioning library version for rocm-cmake
-  Incrementing SOVERSION to 1.0 to meet FBA-29 requirements


library/src/buildinfo.cpp
```diff
@@ -10,7 +10,7 @@
 #define TO_STR(x) TO_STR2(x)
 #define VERSION_STRING                                                           \
     (TO_STR(ROCBLAS_VERSION_MAJOR) "." TO_STR(ROCBLAS_VERSION_MINOR) "." TO_STR( \
-        ROCBLAS_VERSION_PATCH) "." TO_STR(ROCBLAS_VERSION_TWEAK) "-" TO_STR(ROCBLAS_VERSION_COMMIT_ID))
+        ROCBLAS_VERSION_PATCH) "." TO_STR(ROCBLAS_VERSION_TWEAK) )
```
amcamd: change
ROCBLAS_VERSION_PATCH) "." TO_STR(ROCBLAS_VERSION_TWEAK) )
to
ROCBLAS_VERSION_PATCH) "." TO_STR(ROCBLAS_VERSION_TWEAK))
in order to pass clang-format. There should not be space between ) ), it should be ))

saadrahim: thanks, I should be checking the dozens of Jenkins emails I am getting.

PR #621, opened 2019-07-23
Using Blis for test validation

mahmoodw:
Summary of proposed changes:
-  Multi thread CPU test BLAS calls to speed up CI

*PR is to test timing for a full run on jenkins


bragadeesh:
Please reopen after release branching

clients/common/blis_interface.cpp
```diff
@@ -0,0 +1,95 @@
...
+              1,
+              lda,
+              (float*)B,
+              1,
+              ldb,
+              (float*)&beta,
+              (float*)C,
+              1,
+              ldc);
+}
```
leekillough: Support for complex needs to be added (eventually).

clients/common/blis_interface.cpp
```diff
@@ -0,0 +1,95 @@
...
+        std::cerr << "rocblas ERROR: trans != N, T, C" << std::endl;
+        exit(1);
+    }
+}
+
+void setup_blis()
+{
+    bli_init();
+    bli_thread_set_num_threads(omp_get_max_threads());
+}
```
leekillough: Rather than explicitly call `setup_blis` from every client, have a static variable which initializes it, e.g.:

```
static int initialize_blis = (setup_blis(), 0);
```
And put that in `blis_interface.cpp`.

If calling BLIS before `main()` starts is required, I can help provide solutions using function-local static variables.

PR #625, opened 2019-07-29
add bf16 to rocblas-bench

amcamd:
- bfloat16 tests are already part of rocblas-test
- small changes are needed to add bfloat16 to rocblas-bench


amcamd:
quick + pre_checkin + nightly tests pass on standalone gfx900 and gfx906 machines

PR #626, opened 2019-07-30
complex gemm implementation

TonyYHsieh:
1. complex gemm(_ex) and gemm_strided_batched(ex) implementation
2. rocblas test and benchmark for complex

resolves #___

Summary of proposed changes:
-  
-  
-


library/include/rocblas_bfloat16.h
```diff
@@ -73,6 +73,11 @@ struct rocblas_bfloat16
         return u.fp32;
     }
 
+    explicit __host__ __device__ constexpr operator double() const
```
amcamd: can you change to

explicit constexpr __host__ __device__ operator double() const

This will make it consistent with the other functions that have constexpr before __host__

library/include/rocblas-functions.h
```diff
@@ -1736,31 +1742,43 @@ rocblas_qgemm_strided_batched(
...
+                                                            rocblas_int                   lda,
+                                                            rocblas_int                   stride_a,
+                                                            const rocblas_double_complex* B,
+                                                            rocblas_int                   ldb,
+                                                            rocblas_int                   stride_b,
+                                                            const rocblas_double_complex* beta,
+                                                            rocblas_double_complex*       C,
+                                                            rocblas_int                   ldc,
+                                                            rocblas_int                   stride_c,
+                                                            rocblas_int batch_count);
```
amcamd: For gemm_ex documentation at line 1876 need to add information on rocblas_datatype_f32_c and rocblas_datatype_f64r to "Supported types are as follows:"

library/include/rocblas_bfloat16.h
TonyYHsieh: modified in commit "add document for gemm_ex and correct func format"

library/include/rocblas-functions.h
TonyYHsieh: modified in commit "add document for gemm_ex and correct func format"

clients/include/testing_gemm.hpp
```diff
@@ -89,18 +89,8 @@ void testing_gemm(const Arguments& arg)
...
-        h_alpha = float_to_half(arg.alpha);
-        h_beta  = float_to_half(rocblas_isnan(arg.beta) ? 0 : arg.beta);
-    }
-    else
-    {
-        h_alpha = arg.alpha;
-        h_beta  = rocblas_isnan(arg.beta) ? 0 : arg.beta;
-    }
+    T h_alpha = arg.get_alpha<T>();
+    T h_beta  = rocblas_isnan(arg.get_beta<T>()) ? 0.0 : arg.get_beta<T>();
```
leekillough: `get_beta<T>()` should already check for NaN and return 0.0 if beta==NaN.

clients/include/testing_gemm.hpp
```diff
@@ -163,7 +153,7 @@ void testing_gemm(const Arguments& arg)
         rocblas_seedrand();
         rocblas_init<T>(hA, A_row, A_col, lda);
         rocblas_init_alternating_sign<T>(hB, B_row, B_col, ldb);
-        if(rocblas_isnan(arg.beta))
+        if(rocblas_isnan(arg.get_beta<T>()))
```
leekillough: `rocblas_isnan()` should be applied to `arg.beta` rather than `arg.get_beta<T>()`.

clients/include/testing_gemm_strided_batched.hpp
```diff
@@ -22,18 +22,8 @@ void testing_gemm_strided_batched(const Arguments& arg)
...
-        h_alpha = float_to_half(arg.alpha);
-        h_beta  = float_to_half(rocblas_isnan(arg.beta) ? 0 : arg.beta);
-    }
-    else
-    {
-        h_alpha = arg.alpha;
-        h_beta  = rocblas_isnan(arg.beta) ? 0 : arg.beta;
-    }
+    T h_alpha = arg.get_alpha<T>();
+    T h_beta  = rocblas_isnan(arg.get_beta<T>()) ? 0.0 : arg.get_beta<T>();
```
leekillough: `arg.get_beta<T>()` should already consider whether `rocblas_isnan(arg.beta)` and return accordingly.

clients/include/testing_gemv.hpp
```diff
@@ -168,7 +168,7 @@ void testing_gemv(const Arguments& arg)
     rocblas_init<T>(hA, M, N, lda);
     rocblas_init<T>(hx, 1, dim_x, abs_incx);
 
-    if(rocblas_isnan(arg.beta))
```
leekillough: `rocblas_isnan(arg.beta)` should be left alone -- it is separate from `arg.get_beta<T>()`.

library/include/rocblas_bfloat16.h
```diff
@@ -73,6 +73,11 @@ struct rocblas_bfloat16
         return u.fp32;
     }
 
+    explicit constexpr __host__ __device__ operator double() const
```
leekillough: We do not want to all all kinds of types to convert to -- converting from `rocblas_bfloat16` to `float` should be sufficient, then convert up to `double` or other types.

whchung:
+@deven-amd +@jerryyin for awareness. CGEMM / ZGEMM in rocBLAS.

library/src/blas3/Tensile/gemm.h
```diff
@@ -4,6 +4,51 @@
...
+    // First letter refers to A, second letter refers to B
+    NN,
+    NT,
+    TN,
+    TT,
+    NC,
+    CN,
+    TC,
+    CT,
+    CC
```
leekillough: We might want a trailing comma, for consistency

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -335,7 +583,8 @@ rocblas_status gemm_ex_handle_transpose(rocblas_handle    handle,
     static const bool arch_lt906 = handle->device_arch_id() < 906;
     const To* c_in;
     unsigned int ldi, stride_i;
-    if(!arch_lt906 && (std::is_same<Ti, float>{} || std::is_same<Ti, double>{}) &&
+
+    if(!arch_lt906 && (std::is_same<Ti, float>::value || std::is_same<Ti, double>::value) &&
```
leekillough: `std::is_same<T1,T2>{}` can be used as shorthand instead of `std::is_same<T1,T2>::value`, but this is not critical.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -351,103 +600,22 @@ rocblas_status gemm_ex_handle_transpose(rocblas_handle    handle,
...
+                                         static_cast<const Ti*>(b),
+                                         alpha, beta,
+                                         static_cast<unsigned int>(ldd), stride_d,
+                                         static_cast<unsigned int>(ldi), stride_i,
+                                         static_cast<unsigned int>(lda), stride_a,
+                                         static_cast<unsigned int>(ldb), stride_b,
+                                         static_cast<unsigned int>(m),
+                                         static_cast<unsigned int>(n),
+                                         static_cast<unsigned int>(batch_count),
+                                         static_cast<unsigned int>(k),
```
leekillough: `unsigned(...)` can be used instead of `static_cast<unsigned int>(...)` for brevity.

clients/include/testing_gemm.hpp
TonyYHsieh: understand I will correct it on third commit "update complex change".  But what if beta is not nan but betai is nan??  Do we have to consider this case??

clients/include/testing_gemm.hpp
TonyYHsieh: understand. This is my mistake. will update it on third commit "update complex change"

clients/include/testing_gemm_strided_batched.hpp
TonyYHsieh: understand. This is my mistake. will update it on third commit "update complex change"

clients/include/testing_gemv.hpp
TonyYHsieh: understand I will correct it on third commit "update complex change".
Do we have to check nan for betai too??

library/include/rocblas_bfloat16.h
TonyYHsieh: understand I will correct it on third commit "update complex change".

library/src/blas3/Tensile/gemm.h
TonyYHsieh: understand I will correct it on third commit "update complex change".

library/src/blas_ex/rocblas_gemm_ex.hpp
TonyYHsieh: understand I will correct it on third commit "update complex change".

library/src/blas_ex/rocblas_gemm_ex.hpp
TonyYHsieh: understand I will correct it on third commit "update complex change".

TonyYHsieh:
> Need to review alpha and beta behavior when they have NaN values

Library side : didn’t treat NaN  as a special case, so NaN operate with other value will generate NaN. 
                     This implementation looks same as cblas implementation.
Client side: set C as NaN if beta is NaN and set beta as 0 when beta is NaN
                   I follow previous NaN implementation when implement complex gemm.
                   So what is the expectation about NaN implementation??

TonyYHsieh:
> Ignore the CI build failures, which are global. Saad is working on a fix for them.
> 
> I have verified that the full rocBLAS tests pass on GFX900 and GFX906.
> 
> I will look at the code some more, but I approve the changes, as long as:
> 
> 1. Complex values for alpha and beta are tested sufficiently.
    We have add alpha and beta imaginary value in rocblas-test for complex gemm
    Which matrix size and other parameter is same real type.
    Do we need more other testing for complex gemm??

> 2. Complex values for A, B, C, D matrices are tested sufficiently.
    Same as item a. We have add complex test item in rocblas-test.
    Which matrix size and other parameters is same  real type.
    All test item is pass.

> 3. Most of the concerns I've raised in comments are addressed or downplayed, such as the `isnan` stuff with `beta`. Some of the suggestions can be addressed in a later refactoring change.
    We have modified implementation according to your comment and update in other commits in this PR (do we need to squash all commits to one commit ??).
    We only met issue on one comment. ”bfloat16 to double conversion”
    We met build error when convert bfloat16 to double if we didn’t add double conversion.
    So we have to add double conversion. Compile didn’t convert bfloat16 to float and to double when  double(bfloat16{1.0}).

library/include/rocblas_bfloat16.h
leekillough: `rocblas_bfloat16.h` has been updated on develop. This version is outdated.

We should not convert from `rocblas_bfloat16` to `double`. The conversion from `rocblas_bfloat16` to `float` is exact, and it can be up-converted to other types from there. We do not need to list every type conversion.

clients/include/testing_gemm.hpp
leekillough: `is_nan` should be defined for `rocblas_float_complex` and `rocblas_double_complex`, which will return `true` if either the real or imaginary part is `NaN`.

leekillough: But since `arg.beta` and `arg.betai` are always floating-point types, you can use `std::isnan(arg.beta) || std::isnan(arg.betai)`.

clients/include/testing_gemv.hpp
leekillough: `rocblas_isnan(arg.beta) || rocblas_isnan(arg.betai)`. It does not matter that betai is only used during complex. It will default to 0 in non-complex cases.

library/include/rocblas_bfloat16.h
leekillough: `float(x)` should be used to convert `rocblas_bfloat16` to `float`, then it can be converted to other types.

clients/gtest/gemm_gtest.cpp
```diff
@@ -167,11 +164,7 @@ namespace
     // When converted to bool, this functor returns true.
     // Complex is not supported yet.
```
leekillough: I would remove comments about "Complex is not supported yet".

clients/gtest/gemm_gtest.yaml
```diff
@@ -150,10 +150,10 @@ Definitions:
...
 
   - &alpha_beta_range
-    - { alpha:  5, beta:  0 }
-    - { alpha:  0, beta:  3 }
-    - { alpha:  1, beta:  3 }
-    - { alpha:  1, beta:  1 }
+    - { alpha:  5, beta:  0, alphai:  0, betai: -3 }
+    - { alpha:  0, beta:  3, alphai: -3, betai:  0 }
+    - { alpha:  1, beta:  3, alphai:  0, betai:  4 }
+    - { alpha:  1, beta:  1, alphai:  4, betai:  0 }
```
leekillough: It is okay if `alpha`, `alpha`, `beta`, `betai` are specified in the same combination. Basically all of the mappings here get merged into the mapping for the test case, and whichever gets merged last, wins. There is no reason to separate `alpha` and `alphai`, but they can be separated if necessary.

clients/include/norm.hpp
```diff
@@ -21,28 +28,260 @@
 
 /*! \brief  Template: norm check for general Matrix: half/float/doubel/complex  */
 
-// see check_norm.cpp for template speciliazation
 // use auto as the return type is only allowed in c++14
```
leekillough: We use C++14 for our internal code (not external C interfaces), so it's okay to use `auto` as a placeholder for function return type.

clients/include/norm.hpp
```diff
@@ -21,28 +28,260 @@
...
+    double cpu_norm = xlang(&norm_type, &M, &N, hCPU, &lda, work);
+    xaxpy(&size, &alpha, hCPU, &incx, hGPU, &incx);
+
+    double error = xlang(&norm_type, &M, &N, hGPU, &lda, work) / cpu_norm;
+
+    return error;
+}
 
-/*! \brief  Template: norm check for hermitian/symmetric Matrix: half/float/double/complex */
+template <>
```
leekillough: TODO: Consolidate these functions into a smaller number of templates , varying them only by template arguments.

clients/include/norm.hpp
```diff
@@ -21,28 +28,260 @@
...
+        }
+    }
+
+    return cumulative_error;
+}
+
+/* ============== Norm Check for Symmetric Matrix ============= */
+/*! \brief compare the norm error of two hermitian/symmetric matrices hCPU & hGPU */
+
+template <typename T, typename std::enable_if<!is_complex<T>>::type* = nullptr>
```
leekillough: Pointers to `void` are not allowed as template arguments. It should be changed to:
`template <typename T, typename std::enable_if<!is_complex<T>, int>::type = 0>`

clients/include/norm.hpp
```diff
@@ -21,28 +28,260 @@
...
+
+    double cpu_norm = dlansy_(&norm_type, &uplo, &N, hCPU_double, &lda, work);
+    daxpy_(&size, &alpha, hCPU_double, &incx, hGPU_double, &incx);
+
+    double error = dlansy_(&norm_type, &uplo, &N, hGPU_double, &lda, work) / cpu_norm;
+
+    return error;
+}
+
+template <typename T, typename std::enable_if<is_complex<T>>::type* = nullptr>
```
leekillough: As above:
`template <typename T, typename std::enable_if<!is_complex<T>, int>::type = 0>`

clients/include/testing_gemm.hpp
```diff
@@ -89,18 +89,8 @@ void testing_gemm(const Arguments& arg)
...
-        h_alpha = float_to_half(arg.alpha);
-        h_beta  = float_to_half(rocblas_isnan(arg.beta) ? 0 : arg.beta);
-    }
-    else
-    {
-        h_alpha = arg.alpha;
-        h_beta  = rocblas_isnan(arg.beta) ? 0 : arg.beta;
-    }
+    T h_alpha = arg.get_alpha<T>();
+    T h_beta  = arg.get_beta<T>();
```
leekillough: This is the correct usage. You call `get_alpha` or `get_beta` with a template argument of `<T>`. It will return `0` if the original value is `NaN`, and will return the original value otherwise.

clients/include/testing_gemm_ex.hpp
```diff
@@ -250,7 +250,7 @@ void testing_gemm_ex(const Arguments& arg)
 
     bool nantest = rocblas_isnan(arg.beta);
```
leekillough: This may need to be changed to:

`bool nantest = rocblas_isnan(arg.beta) || rocblas_isnan(arg.betai);`

For non-complex types, `betai` is `0` and so `rocblas_isnan(arg.betai)` is `false`.

library/src/blas3/Tensile/gemm.cpp
```diff
@@ -54,83 +26,510 @@ const char* tensileGetSolutionName(rocblas_operation trans_a,
...
+    static_assert(std::is_standard_layout<TensileComplexDouble>{},
+                  "rocblas_float_complex is not a standard layout type, and thus is "
+                  "incompatible with C.");
+
+    static_assert(std::is_trivial<TensileComplexDouble>{},
+                  "rocblas_float_complex is not a trivial type, and thus is "
+                  "incompatible with C.");
+
+    static_assert(sizeof(rocblas_double_complex) == sizeof(TensileComplexDouble),
+                  "internal rocblas_float_complex does not match public rocblas_double_complex");
```
leekillough: The comment is a little off.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -206,104 +259,298 @@ TensileStatus tensile_Cijk_Alik_Bjlk_B<TensileHalf, TensileHalf, float>(TENSILE_
...
-TensileStatus tensile_Cijk_Alik_Bjlk_B<TensileInt8x4, TensileInt32, TensileInt32>(
+inline TensileStatus tensile_Cijk_Alik_Bjlk_B<TensileInt8x4, TensileInt32, TensileInt32>(
     TENSILE_IN_ARGS(TensileInt8x4, TensileInt32, TensileInt32))
 {
     return tensile_Cijk_Alik_Bjlk_4xi8BH(TENSILE_OUT_ARGS);
 }
+
+//----- typename_data=rocblas_float_complex ---------- typename_compute = rocblas_float_complex --------------------------
+#define TENSILE_COPMLEX_OUT_ARGS(Ti, To, Tc)                                        \
+    (To*)dataD, (const To*)dataC, (const Ti*)dataA, (const Ti*)dataB,               \
```
leekillough: Typo -- `COPMLEX` should be `COMPLEX`

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -206,104 +259,298 @@ TensileStatus tensile_Cijk_Alik_Bjlk_B<TensileHalf, TensileHalf, float>(TENSILE_
...
+          "rocblas_float_complex is not a trivial type, and thus is "
+          "incompatible with C.");
+
+static_assert(sizeof(rocblas_float_complex) == sizeof(TensileComplexFloat),
+          "internal rocblas_float_complex does not match public rocblas_float_complex");
+template <>
+inline TensileStatus tensile_Cijk_Ailk_Bljk_B<rocblas_float_complex,rocblas_float_complex,rocblas_float_complex>(
+    TENSILE_IN_ARGS(rocblas_float_complex, rocblas_float_complex, rocblas_float_complex))
+{
+    return tensile_Cijk_Ailk_Bljk_CB(TENSILE_COPMLEX_OUT_ARGS(TensileComplexFloat, TensileComplexFloat, TensileComplexFloat));
```
leekillough: Same typo throughout file -- maybe do global search and replacce

clients/common/norm.cpp
```diff
@@ -1,596 +0,0 @@
-/* ************************************************************************
```
leekillough: I agree with moving `norm.cpp` to a header file `norm.hpp`.

clients/include/testing_gemm.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/include/testing_gemv.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/gtest/gemm_gtest.cpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/include/norm.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/include/norm.hpp
TonyYHsieh: Need to wrap rocblas_half to a class form to get conversion operation from half to double.
I think this can be done later.

clients/include/norm.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/include/norm.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/include/testing_gemm_ex.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

library/src/blas3/Tensile/gemm.cpp
TonyYHsieh: understand. update it in commit "update complex change 2".

library/src/blas_ex/rocblas_gemm_ex.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

library/src/blas_ex/rocblas_gemm_ex.hpp
TonyYHsieh: understand. update it in commit "update complex change 2".

clients/common/norm.cpp
TonyYHsieh: Thanks for agreement  :)

amdkila:
@TonyYHsieh Your PR failed CI due to minor formatting differences on just 1 file (rocblas_gemm_ex.cpp), so I re-ran the formatter. It should pass the format check now.

PR #630, opened 2019-08-01
Check hip-clang as well.

mhbliao:
resolves SWDEV-198354

- Need to check hip-clang as well to include the full definition of bfloat16 & complex


mhbliao:
BTW, are we still OK to pull this into 2.7 branch?

leekillough:
CI tests are failing right now. I'm running the tests outside of Jenkins now, and will merge once they pass.

If this needs to go into 2.7, then a special hotfix needs to be approved.

PR #638, opened 2019-08-06
add cmake variable Tensile_COMPILER so rocBLAS can set Tensile compil…

amcamd:


aaronenyeshi:
When --hip-clang flag is passed into install.sh, should this Tensile_COMPILER flag be set to hipcc?

amcamd:
In tests there were diagnostic print statements that indicated below occurs. I removed the print statements, but below should still occur, if not, it will need to be fixed:
- Calling install.sh with --hip-clang will set build_hip_clang=true
- this will set ${compiler} = hipcc
- this will invole cmake with -DTensileCompiler=hipcc
- this will set the cmake cache variable Tensile_COMPILER to hipcc
- this will call TensileCreateLibrary with --cxx-compiler=hipcc
- this will set CXXCompiler in TensileCreateLibrary to hipcc
- this will be passed eventually to buildSourceCodeObjectFile as CxxCompiler, and it will change the compiler based on the if statement in line 125 of Tensile/TensileCreateLibrary.py

PR #641, opened 2019-08-07
merge develop into master for ROCm2.8

amcamd:


wbgilmartin (approved):
LGTM

leekillough (commented):
Without looking at all of the individual changes, if this passes CI, then I approve it.

There have been so many changes lately to Jenkins and rocBLAS/Tensile, that I'd be surprised if this passes in one go.

We're still waiting for the Complex GEMM changes to be checked in; I suspect the time difference is the reason. I've addressed all of my main concerns, and Tony has acknowledged most of them.

PR #652, opened 2019-08-13
ROCm2.8 hot_fix: add tuned logic files for resnet and inception sizes

amcamd:


amcamd:
All tests quick+pre_checkin+nightly pass on gfx900 and gfx906

PR #655, opened 2019-08-15
fp16 and bfloat16 precision for dot

daineAMD:
Summary of proposed changes:
-  add fp16 precision with fp32 accumulation for dot
-  add bfloat16 precision with fp32 accumulation for dot


clients/benchmarks/client.cpp
```diff
@@ -48,14 +48,19 @@ using namespace std::literals;
 #include "testing_trsv.hpp"
 
 // Template to dispatch testing_gemm_ex for performance tests
-// When Ti == void or complex, the test is marked invalid
+// When Ti == void or complex or Ti == To == Tc == bfloat16, the test is marked invalid
 template <typename Ti, typename To = Ti, typename Tc = To, typename = void>
```
amcamd: I do not see a test for Ti == complex. Can the comment be updated if there is no test for complex.

clients/include/rocblas_common.yaml
```diff
@@ -173,6 +175,19 @@ Single double joined: &single_double_complex_real_in_complex_out
   - *single_precision_complex_real_in_complex_out
   - *double_precision_complex_real_in_complex_out
 
+#############################################
+#           Used for Dot (quick)            #
+#############################################
+Half bfloat single double complex real: &half_bfloat_single_double_complex_real_precisions
+  - *half_precision
+  - *bfa_precision
```
amcamd: I think bf16 is more descriptive than bfa. Is it possible to make this bf16?

daineAMD: Good point, changed in #f746ec6.

daineAMD:
All tests (quick, pre-checkin, nightly) pass on gfx900 and gfx906.

PR #658, opened 2019-08-19
Rot functions

AlexBrownAMD:
Added rot, rotg, rotm, rotmg and test code for all relevant data types.


clients/include/cblas_interface.hpp
```diff
@@ -365,6 +365,196 @@ inline void cblas_swap(rocblas_int             n,
...
+inline void cblas_rotmg(float* d1, float* d2, float* b1, const float* b2, float* p)
+{
+    cblas_srotmg(d1, d2, b1, *b2, p);
+}
+
+template <>
+inline void cblas_rotmg(double* d1, double* d2, double* b1, const double* b2, double* p)
+{
+    cblas_drotmg(d1, d2, b1, *b2, p);
+}
```
leekillough: This should be templatized; also, we don't provide Fortran-compatible APIS for any routines.

clients/include/testing_rotm.hpp
```diff
@@ -0,0 +1,189 @@
...
+                << ",norm_error_host_x,norm_error_host_y,norm_error_device_x,norm_error_device_y";
+        std::cout << std::endl;
+        std::cout << N << "," << incx << "," << incy << "," << gpu_time_used << ","
+                  << cpu_time_used;
+        if(arg.norm_check)
+            std::cout << ',' << norm_error_host_x << ',' << norm_error_host_y << ","
+                      << norm_error_device_x << "," << norm_error_device_y;
+        std::cout << std::endl;
+    }
+}
```
leekillough: The testing_rot* functions should be consolidated into one file, not duplicated.

library/src/blas1/rocblas_rotg.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+rocblas_status rocblas_zrotg(rocblas_handle          handle,
+                             rocblas_double_complex* a,
+                             rocblas_double_complex* b,
+                             double*                 c,
+                             rocblas_double_complex* s)
+{
+    return rocblas_rotg(handle, a, b, c, s);
+}
+
```
leekillough: The rot* functions should be consolidated into one file and differentiated by template arguments, not copied-and-edited.

AlexBrownAMD: Not sure I follow your meaning, the rot* functions are all different operations, they are not copies of each other

clients/include/cblas_interface.hpp
AlexBrownAMD: Can you provide an example? I think these are templated the same way as the rest of cblas_interface

leekillough:
When you finally commit your changes, please use "Squash and merge" to minimize the number of commits on the main develop branch.

AlexBrownAMD:
All tests quick+pre_checkin+nightly pass on gfx900 and gfx906

PR #659, opened 2019-08-20
revert to rocBLAS develop calling Tensile develop

amcamd:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ if( BUILD_WITH_TENSILE )
   else()
     # Use the virtual-env setup and download package from specified repot:
     set( tensile_fork "ROCmSoftwarePlatform" CACHE STRING "Tensile fork to use" )
-    set( tensile_tag a472399c5e97ac13b171092c2eb61d87d3a08d53 CACHE STRING "Tensile tag to download" )
+    set( tensile_tag develop CACHE STRING "Tensile tag to download" )
     virtualenv_install("git+https://github.com/${tensile_fork}/Tensile.git@${tensile_tag}")
     message (STATUS "using GIT Tensile fork=${tensile_fork} from branch=${tensile_tag}")
   endif()


pfultz2:
This makes build non-reproducible. Rocblas should always use a commit hash or a tag for tensile. If we want miopen or migraphx to use a newer rocblas version from develop(perhaps we are using a newer rocblas feature) then we need to have rocblas builds to be reproducible.

saadrahim:
@amcamd The fixes in rocBLAS to revert to Tensile develop are not their yet. For CI purposes, it is better to have rocBLAS track a Tensile commit. That way we can create PRs to update the commit and ensure that Tensile never breaks rocBLAS (which happens far too often).

leekillough:
Should we switch Tensile to become a submodule of rocBLAS?

saadrahim:
@leekillough The submodule is the best long term approach.

amcamd:
We use rocBLAS develop as part of testing for Tensile develop. Is it possible to use the rocBLAS master branch for a reproducible build? This is fine for rocBLAS, but there will be delayed testing of Tensile, and Tensile bugs will be caught later.

saadrahim:
> We use rocBLAS develop as part of testing for Tensile develop. Is it possible to use the rocBLAS master branch for a reproducible build? This is fine for rocBLAS, but there will be delayed testing of Tensile, and Tensile bugs will be caught later.

Yes the Tensile bugs are caught immediately, but then you need to use bisect or other methods to find out which commit caused it. As long as we update the Tensile commit ID regularly to force a test, then Tensile can be tested in a controlled manner. The current strategy is too random.

saadrahim:
I also want to add a Jenkins job to test rocBLAS against Tensile develop latest. This should cover most of your concerns. As Paul is commenting, things are not reproducible.

amcamd:
You are going against the principles of agile in favor of waterfall. If you want stable, use master, it you want the latest features use develop.

leekillough:
> 
> 
> You are going against the principles of agile in favor of waterfall. If you want stable, use master, it you want the latest features use develop.

Why not then make Tensile a submodule of rocBLAS, and have master rocBLAS -> master Tensile and develop rocBLAS -> develop Tensile? As for going the other direction, from Tensile -> rocBLAS, I don't know an immediate solution but maybe there's a trick.

It seems like submodule maintenance would be less than these back-and-forth `CMakeLists.txt` changes.

saadrahim:
I think all changes should be tested before it is committed. The system you propose does not test changes in Tensile against rocBLAS. 

At the moment, we need to resolve https://github.com/ROCmSoftwarePlatform/Tensile/pull/636 before changing back to develop.

pfultz2:
>  Is it possible to use the rocBLAS master branch for a reproducible build?

Not always. Sometimes we need to pull from develop as the features have not been merged into master yet.

> Using a Tensile tag or hash is fine for rocBLAS, but there will be delayed testing of Tensile, and Tensile bugs will be caught later.

There are several solutions to this. The `tensile_tag` variable is a cached variable so you can override to pull somewhere else. For example, if tensile want to build against rocblas

 Also, you can install tensile first with `pip` and then build rocblas, as rocblas only pulls tensile if it isnt already installed.

> If you want stable, use master, it you want the latest features use develop.

We want to use the latest feature, but we also want reproducibility.

pfultz2:
> Why not then make Tensile a submodule of rocBLAS, and have master rocBLAS -> master Tensile and develop rocBLAS -> develop Tensile? As for going the other direction, from Tensile -> rocBLAS, I don't know an immediate solution but maybe there's a trick.

You can't override where tensile gets pulled with submodules. It also makes rocblas so the source tarballs from github no longer works.

> It seems like submodule maintenance would be less than these back-and-forth CMakeLists.txt changes.

You still need to commit updates to the submodule. Submodules dont pull from the latest HEAD on the branch on every clone. So instead of committing updates to the CMakeLists.txt it will be updates to the the submodules

amcamd:
The only way you will get reproducibility is to pull a specific rocBLAS tag and a Tensile tag. As you mentioned,  the tensile_tag variable is a cached variable so you can override to pull somewhere. Pulling the rocBLAS develop branch that pulls a specific Tensile tag does not give you reproducibility. The rocBLAS develop branch changes.

sdquiring:
This makes things more difficult:

1. Commit history.

If we need to find out when rocBLAS broke, or otherwise look at an old version of the code, we can’t accurately check out rocBLAS from a particular moment in time.  This makes debugging difficult.

2. Breaking changes.

There’s no way to check in a change to Tensile that breaks rocBLAS while reliably having rocBLAS tests pass.  The only way is to merge the two changes at the same time.  This means we can’t properly test a set of Tensile/rocBLAS changes with CI.

If we include the commit hash, then you can make a corresponding rocBLAS PR that simultaneously fixes the issue and update to the hash with the breaking change.  That PR can be tested in CI so we know it will fix it.

We absolutely should include a Jenkins job that always pulls the latest develop from tensile (just need to run `install.sh --branch develop`). That would essentially be additional CI for Tensile.

sdquiring:
> The only way you will get reproducibility is to pull a specific rocBLAS tag and a Tensile tag.

And how do you know which Tensile tag to pull if it's not checked in?

saadrahim:
Also, developers can modify their build scripts to pass the Tensile branch as develop if they need the latest. CI and 3rd party users would prefer a little more stability.

amcamd:
You have a point, we cannot check out a tag that is not checked in. It it is checked in to a fork you can use the cached variable tensile_fork. This allows for pulling from a fork and tag or a fork and commit ID or a fork and branch.

pfultz2:
> Pulling the rocBLAS develop branch that pulls a specific Tensile tag does not give you reproducibility. The rocBLAS develop branch changes.

Yes, in MIOpen we always pull a commit hash(or tag) of rocBLAS(see [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/master/requirements.txt#L4)). We never pull from a branch since we would lose reproducibility as well. Although, the commit hash maybe from the develop branch of rocBLAS, which is why its important for the develop branch of rocBLAS to use a hash for tensile.

PR #668, opened 2019-08-21
Adding Jenkins job to test Tensile Develop latest

saadrahim:
This will add CI jobs to continuously run changes on Tensile latest develop branch. In the future we will add an automatic PR generation.


saadrahim:
This should run only nightly until we figure out how to scan Tensile to trigger this.

Jenkinsfile.dependency
```diff
@@ -0,0 +1,169 @@
...
+        platform.runCommand(this, command)
+    }
+
+    def testCommand =
+    {
+        platform, project->
+
+        def command
+
+        if(platform.jenkinsLabel.contains('centos'))
```
amdkila: Change this to only run nightly tests

Jenkinsfile.dependency
```diff
@@ -0,0 +1,139 @@
...
+// If you are interested in running your own Jenkins, please raise a github issue for assistance.
+
+import com.amd.project.*
+import com.amd.docker.*
+
+////////////////////////////////////////////////////////////////////////
+// Mostly generated from snippet generator 'properties; set job properties'
+// Time-based triggers added to execute nightly tests, eg '30 2 * * *' means 2:30 AM
+properties([
+    pipelineTriggers([cron('0 1 * * *'), [$class: 'PeriodicFolderTrigger', interval: '5m']]),
```
sdquiring: Should this not include the PeriodicFolderTrigger so we only run nightly?

Also, looking at the help for this suggests we should be using the "H 1 * * *" instead of "0 1 * * *" so the jobs don't all start at once.

Jenkinsfile.dependency
```diff
@@ -0,0 +1,139 @@
...
+                    set -ex
+                    cd ${project.paths.project_build_prefix}
+                    LD_LIBRARY_PATH=/opt/rocm/hcc/lib CXX=/opt/rocm/bin/hipcc ${project.paths.build_command} --hip-clang
+                    """
+        }
+        else
+        {
+            command = """#!/usr/bin/env bash
+                    set -ex
+                    cd ${project.paths.project_build_prefix}
```
saadrahim: I don't think set -ex does anything here as it is overwritten in the compile script

Jenkinsfile.dependency
saadrahim: I removed cron, and just going on a folder trigger so it happens only if Tensile changes.

saadrahim: @sdquiring Also, feedback on install.sh file is going to go in another PR.

saadrahim:
@sdquiring pytorch has a .jenkins folder. I think we should follow that organization structure and move all the Jenkinsfiles to the .jenkins.

Issue #664, opened 2019-08-20
rocBLAS no longer builds


pfultz2:
When building rocBLAS from 2.7 release(ie from [1ac2271e72d21fa745c7af87cd5fd102f90cba45](https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/1ac2271e72d21fa745c7af87cd5fd102f90cba45)) we get an error from cmake:

```
Please run this script after at least one build of rocBLAS.
library/src/CMakeFiles/rocblas.dir/build.make:697: recipe for target 'library/src/librocblas.so.0.1' failed
```

pfultz2:
It seems the `POST_BUILD` process is the problem [here](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/bae39ccb155f009660f9d9047d917449cee55157/library/src/CMakeLists.txt#L212). It seems `header_compilation_tests.sh` should run during testing not during a regular build of rocBLAS.

justxi:
The script https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/f1b6bb43594b6652e00c39d1df248d836b878df9/header_compilation_tests.sh#L8 has a fixed path. It depends where you have your build directory. If you have your build directory outside of the source tree, it fails on my tests.

xdgarrido:
Rocblas is not building for ROCM 2.7 even if using install.sh. Please, try Rocm 2.7 .



The rocBLAS checked out from GitHub branch master is failing to build, this is the output:-

 

[ 11%] Building CXX object library/src/CMakeFiles/Tensile.dir/__/__/Tensile/Tools.cpp.o

clang-9: warning: -amdgpu-target argument 'gfx908' is not recognized; using gfx803 instead [-Winvalid-command-line-argument]

clang-9: warning: -amdgpu-target argument 'gfx908' is not recognized; using gfx803 instead [-Winvalid-command-line-argument]

clang-9: warning: -amdgpu-target argument 'gfx908' is not recognized; using gfx803 instead [-Winvalid-command-line-argument]

clang-9: warning: -amdgpu-target argument 'gfx908' is not recognized; using gfx803 instead [-Winvalid-command-line-argument]

clang-9: warning: -amdgpu-target argument 'gfx908' is not recognized; using gfx803 instead [-Winvalid-command-line-argument]

[ 11%] Built target rocblas-test-data

clang-9: error: unable to execute command: Killed

[...]

PR #670, opened 2019-08-21
Fix path issue in header compilation headers

pfultz2:
This issue #664. It sets the path correctly, and only runs the test when `BUILD_TESTING` is enabled.


leekillough:
Another PR is being created for this issue, without moving it from builds to tests, and with preserving the parallelism in the script on develop.

PR #672, opened 2019-08-23
Updating Tensile Tag

saadrahim:


saadrahim:
> 
> 
> If this is going to be common, perhaps the tag should be stored in a separate file so that `CMakeLists.txt` isn't changed often.
> 
> Or perhaps submodules should be used, so that the tag is stored inside Git.

I concur, we should go to submodules. We can discuss by email.

pfultz2:
> I concur, we should go to submodules. We can discuss by email.

Let me just say that submodules is not a very good solution from a user perspective as rocBLAS will no longer have source tarballs from github that would work, which will prevent us from listing rocBLAS in our requirements.txt. We also would not like to have to use `git clone --recursive` to get rocBLAS because it can be really slow(we need dont need the entire history of rocBLAS and Tensile when building rocBLAS in our CI).

I dont really see what the issue is with updating a CMakeLists.txt file vs another file vs a git submodule.

saadrahim:
Updating the tag for latest Tensile with the fix that broke Tensile for the past few days. Jenkins will confirm this overnight.

PR #675, opened 2019-08-27
Adding flag to install script to choose cpu lib

mahmoodw:
I have added a flag to the install script to be able to choose between the blis and lapack cpu libraries.
- cpu_lib flag is set to blis by default. 

I believe I should update the wiki page to include information on this flag when building the client.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -22,7 +22,7 @@ function display_help()
   echo "    [-o|--cov] Set tensile code_object_version (V2 or V3)"
   echo "    [-t|--test_local_path] Use a local path for tensile instead of remote GIT repot"
 #  echo "    [--cuda] build library for cuda backend"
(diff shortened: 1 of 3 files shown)


saadrahim:
Will lapack build without openmp? I would like to quickly get SLES working.

install.sh
```diff
@@ -22,6 +22,7 @@ function display_help()
   echo "    [-o|--cov] Set tensile code_object_version (V2 or V3)"
   echo "    [-t|--test_local_path] Use a local path for tensile instead of remote GIT repot"
 #  echo "    [--cuda] build library for cuda backend"
+  echo "    [--cpu_lib] specify libary to use for cpu reference code in testing (blis or lapack)"
```
saadrahim: test library?

saadrahim: cpu_ref_lib?

PR #678, opened 2019-08-30
Fix rocblas_bfloat16 conversions and add a __device__ __host__ rocblas_abs() function

leekillough:
I will have to do build and test run manually.

I'm not 100% sure this fixes it, but it should.

### DIFF
--- a/library/include/rocblas_bfloat16.h
+++ b/library/include/rocblas_bfloat16.h
@@ -63,7 +63,7 @@ struct rocblas_bfloat16
     }
 
     // zero extend lower 16 bits of bfloat16 to convert to IEEE float
-    explicit constexpr __host__ __device__ operator float() const
+    constexpr __host__ __device__ operator float() const
     {
         union
         {
@@ -71,11 +71,6 @@ struct rocblas_bfloat16
             float    fp32;
         } u = {uint32_t(data) << 16};
         return u.fp32;
-    }
-
-    explicit constexpr __host__ __device__ operator double() const
-    {
-        return double(float(*this));
     }
 
 private:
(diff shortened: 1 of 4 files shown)


leekillough:
There were three problems addressed:
1. The `rocblas_bfloat16` type should be allowed to be implicitly converted to `float`, not just explicitly converted to `float` or `double`. There is no loss in the conversion from `rocblas_bfloat16` to `float`, so it can be allowed implicitly. If converted to `double`, it will be implicitly converted to `float` first.
2. The `std::abs` overload for `rocblas_bfloat16` was buggy, using the _integer_ value of `data & 0x7fff` converted to `rocblas_bfloat16` as the return value, instead of the absolute value of the BF16 value.
3. The `std::abs` function was being called from `__device__` functions, which isn't generally allowed. As a result, when trying to call `std::abs` from `__device__` functions with `float` or `double` arguments, the compiler only saw the `rocblas_bfloat16` overload of `std::abs` as a viable `__device__` overload candidate function. It then tried to convert the `float` or `double` argument to `rocblas_bfloat16`, but it couldn't, since `rocblas_bfloat16`'s constructor is `explicit` (a conversion to `rocblas_bfloat16` from `float` or `double` is lossy, and thus should not be allowed implicitly).

The fix:
1. Allow `rocblas_bfloat16` to be implicitly converted to `float`, without loss.  The conversion to `double` will then be automatic by the compiler, first converting it to `float` and then to `double`.
2. Remove the `std::abs` overloads for `rocblas_bfloat16`.
3. Add a `rocblas_abs` function to be callable from `__device__` or `__host__` functions, and put the `rocblas_bfloat16` overfload there.
4. Change ROT*g functions to use `rocblas_abs` instead of `std::abs`.

PR #679, opened 2019-09-03
Adding SLES support for Jenkins

saadrahim:


cmake/os-detection.cmake
```diff
@@ -0,0 +1,31 @@
+
+# Modified from https://github.com/OPM/opm-common/blob/master/cmake/Modules/UseSystemInfo.cmake based on GNU General Public License v3.0 from https://github.com/OPM/opm-common/blob/master/LICENSE
```
saadrahim: Need someone to review this change

PR #681, opened 2019-09-04
Fix header file checks to not depend on hardcoded paths

### LINKED ISSUE #664: rocBLAS no longer builds

leekillough:
resolves #664

Change `header_compilation_tests.sh` to use CMake variables and data extracted from files, instead of hard-coded paths.

I have verified that it works in a `/tmp` build directory instead of the standard rocBLAS build tree. Running `install.sh` and running `make` from `build/release` still run the header checks.

We are nearing code-complete and this will be merged into `master` (or the `staging` branch) shortly, which will add parallelism to the existing script on `master`, making it less annoying during builds.

### DIFF
--- a/header_compilation_tests.sh
+++ b/header_compilation_tests.sh
@@ -2,15 +2,22 @@
 
 export LOCALE=C
 set -e
-
 exec 2>&1
 
-if [[ ! -e build/release/include/rocblas-export.h ]]; then
+script=$(realpath "$0")
+
+build_first()
+{
     echo "Please run this script after at least one build of rocBLAS."
     exit 1
-fi
+}
 
-script=$(realpath "$0")
+[[ ! -e CMakeCache.txt ]] && build_first
+
+BUILD_DIR=$(realpath "$(pwd)")
+SOURCE_DIR=$(realpath -m "$(grep CMAKE_HOME_DIRECTORY CMakeCache.txt | sed 's/CMAKE_HOME_DIRECTORY:INTERNAL=//g')")
+
+[[ ! -e $BUILD_DIR/include/rocblas-export.h ]] && build_first
 
 # Returns whether the output file is up to date.
 # Prints the output file.
@@ -20,7 +27,7 @@ script=$(realpath "$0")
 out_uptodate()
 {
     local file="$1_$2"
-    local filename="build/compilation_tests/$file.o"
+    local filename="$BUILD_DIR/compilation_tests/$file.o"
     mkdir -p $(dirname "$filename")
     local out=$(realpath -m "$filename")
     echo "$out"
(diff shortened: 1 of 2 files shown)


pfultz2:
This test should be disabled when using a normal build. Besides adding unnecessary extra build time to rocblas, it also uses hardcoded paths that will break when hcc and hip are in non-standard paths.

The best solution is to create cmake targets that are `EXCLUDE_FROM_ALL` instead of doing this in bash. In MIGraphX, we generate header tests automatically in cmake [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/CMakeLists.txt#L142)(and the target is created with `EXCLUDE_FROM_ALL` [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/CMakeLists.txt#L70)) that also includes the header across TUs so we dont accidentally create duplicate symbols by forgetting an `inline` or `static`. 

Of course, that is more work, so in the short term you can disable the test when tests are disabled in cmake via the `BUILD_TESTING` variable.

leekillough:
If there are more hard coded paths which need to use variables instead, we can add them.

This is intentionally done at build time, because it's closely tied to the build make system, and the rocBLAS **test** build system is very dissimilar. Also, errors in this script are hard build errors, not just test failures, meaning that the product is unreleaseable.

The time is minimal with parallelism (<10 seconds the first time; instantaneous on later builds). You are basing it on the old version, which was not parallel.

The compiler arguments are hard-coded intentionally to test different use-case scenarios. If one of them fails, it is a concern to us. We want to ensure different compilers and different languages can compile all headers.

pfultz2:
> If there are more hard coded paths which need to use variables instead, we can add them.

The cmake targets should be used for the usage requirements not a bunch of variables.

> This is intentionally done at build time, because it's closely tied to the build make system, and the rocBLAS test build system is very dissimilar. 

As a user installing rocBLAS, it is not necessary.

> Also, errors in this script are hard build errors, not just test failures, meaning that the product is unreleaseable.

Yes, but that should've been checked by the CI before merging. As a user, I assume you have already tested all these things, I just need to build and install rocBLAS.

> The compiler arguments are hard-coded intentionally to test different use-case scenarios. If one of them fails, it is a concern to us. We want to ensure different compilers and different languages can compile all headers.

You can make the tests for different languages in cmake, and also you can set different compilers in the clients if you want to test different compilers for the same language.

Ideally, you would have a C test in rocblas to test the headers with vanilla gcc. Another test in rocBLAS to test the headers with `hip::host` and another test with `hip::device`. Then you can have an additional test in the clients with g++(in fact the clients dont need to be changed since they already include the rocblas headers).

leekillough:
As a user, with an unknown ROCm installed, unknown previous version of rocBLAS installed, etc., we need to run these "tests" at build time.

We may add other checks in the future. Two weeks ago, the original issue was taken over by someone who could not build rocBLAS without it hanging, until he removed a previous rocBLAS.

As we transition between different compilers and GPU targets, this kind of checking is best done at build time, to save time later. .githooks is not the place to check it, because Git commands are unrelated to build status -- someone may want to make commits before builds. These checks should be done post-builds, not pre-commit. They verify that rocBLAS is compatible with the user's build environment, while allowing flexibility.

PR #682, opened 2019-09-04
Fixing the install script to throw errors correctly.

saadrahim:


saadrahim:
The commit to simulate failure will be removed obviously.

saadrahim:
This was tested and correctly throws errors in the compile stage of Jenkins.

PR #686, opened 2019-09-09
Addition of scal_batched and scal_strided_batched functions

daineAMD:
- Adds scal_batched and scal_strided_batched functions
- Exposes offset param in .hpp files for further internal extension
- Changed L1 testing for better distribution of testing times

- General file layout (as per Juan's document and discussion of exposing offset for internal use):
    - rocblas_scal.cpp, rocblas_scal_batched.cpp, etc. contain C calls
    - rocblas_scal.hpp, rocblas_scall_batched.hpp, etc. contain logging and error checking
    - scal_host.hpp contains quick return and device calls
    - scal_device.hpp contains kernels and device calls

rocblas-test times for all scal functions:
- quick: 11.5s
- pre-checkin: 41s
- nightly: 185s

Wasiq/Torre/others are welcome to give feedback before Lee & Andrew as requested.


clients/gtest/blas1_gtest.yaml
```diff
@@ -21,70 +25,192 @@ Definitions:
...
+      - scal_batched: *single_double_precisions_complex_real
+      - scal_batched: *single_double_complex_real_in_complex_out
+
+  - name: blas1_strided_batched
+    category: quick
+    N: [ -1, 0, 5, 33792 ]
+    incx_incy: *incx_incy_range
+    alpha_beta: *alpha_beta_range
+    alphai_betai: *alphai_betai_range
+    batch_count: [-1, 5]
```
mahmoodw: You may want to add batch_count of 0 to the quick tests to test the API quick return functionality.

clients/common/rocblas_gentest.py
```diff
@@ -196,7 +196,7 @@ def setdefaults(test):
     # These are only for dynamic defaults
     # TODO: This should be ideally moved to YAML file, with eval'd expressions.
 
-    if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
+    if all([x in test for x in ('M', 'incx', 'strideScale')]) and (test['function']=='ger_strided_batched' or test['function']=='scal_strided_batched'):
```
TorreZuk: Minor but might be worth putting the specific function name check first for readability and to short circuit as it is likely a cheaper test.

clients/include/testing_scal_batched.hpp
```diff
@@ -0,0 +1,223 @@
...
+    rocblas_seedrand();
+    for(int i = 0; i < batch_count; i++)
+    {
+        rocblas_init<T>(hx_1[i], 1, N, incx);
+
+        hx_2[i]    = hx_1[i];
+        hx_gold[i] = hx_1[i];
+    }
+
+    // copy data from CPU to device, does not work for incx != 1
```
TorreZuk: Is the "does not work" still true?

clients/include/testing_scal_batched.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // CPU BLAS
+        cpu_time_used = get_time_us();
+        for(int i = 0; i < batch_count; i++)
+        {
+            cblas_scal<T, U>(N, h_alpha, hx_gold[i], incx);
+        }
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = axpy_gflop_count<T>(N) / cpu_time_used * 1e6 * 1;
```
TorreZuk: scal_gflop_count?

clients/include/testing_scal_strided_batched.hpp
```diff
@@ -0,0 +1,238 @@
...
+                                                                     incx,
+                                                                     stridex,
+                                                                     batch_count));
+
+        return;
+    }
+
+    size_t size_x = N * size_t(incx) * batch_count
+                    + (static_cast<size_t>(stridex) - N * size_t(incx))
+                          * static_cast<size_t>(batch_count - 1);
```
YvanMokwinski: Maybe a short comment to explain this formula.
It seems that here a value range is assumed for stridex since the size is equal to: stridex * (batch_count-1) + N * incx, which makes sense only if stridex > N * incx.
In this case why not simply have stridex * batch_count?

clients/include/testing_scal_strided_batched.hpp
```diff
@@ -0,0 +1,238 @@
...
+                                                                     dx,
+                                                                     incx,
+                                                                     stridex,
+                                                                     batch_count));
+
+        return;
+    }
+
+    size_t size_x = N * size_t(incx) * batch_count
+                    + (static_cast<size_t>(stridex) - N * size_t(incx))
```
TorreZuk: size_t( ) instead of static_cast , maybe we should comment on the stride assumptions when calculating

clients/include/testing_scal_strided_batched.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+        // CPU BLAS
+        cpu_time_used = get_time_us();
+        for(int i = 0; i < batch_count; i++)
+        {
+            cblas_scal<T, U>(N, h_alpha, hx_gold + i * stridex, incx);
+        }
+
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = axpy_gflop_count<T>(N) / cpu_time_used * 1e6 * 1;
```
TorreZuk: axpy_ -> scal_

library/src/blas1/scal_host.hpp
```diff
@@ -0,0 +1,82 @@
...
+template <typename T, typename U>
+rocblas_status rocblas_scal_batched_template(rocblas_handle handle,
+                                             rocblas_int    n,
+                                             const U*       alpha,
+                                             T*             x[],
+                                             rocblas_int    offsetx,
+                                             rocblas_int    incx,
+                                             rocblas_int    batch_count)
+{
+    static constexpr int NB = 256;
```
YvanMokwinski: Seems to be the common practice but I had a discussion with Scott and putting the block size as a template parameter might be better for future tuning.

library/src/blas1/rocblas_scal_batched.hpp
```diff
@@ -0,0 +1,92 @@
...
+constexpr char rocblas_scal_name<rocblas_float_complex, float>[] = "rocblas_csscal_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex, double>[] = "rocblas_zdscal_batched";
+
+template <typename T, typename U>
+rocblas_status rocblas_scal_batched_impl(rocblas_handle handle,
+                                            rocblas_int    n,
+                                            const U*       alpha,
+                                            T*             x[],
+                                            rocblas_int    offsetx,
```
YvanMokwinski: For my understanding, what is the purpose of this offsetx parameter? Seems to be an implementation convenience but I don't see the immediate benefit of it.

library/src/blas1/rocblas_scal_batched.hpp
```diff
@@ -0,0 +1,92 @@
...
+constexpr char rocblas_scal_name<rocblas_float_complex>[] = "rocblas_cscal_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex>[] = "rocblas_zscal_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_float_complex, float>[] = "rocblas_csscal_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex, double>[] = "rocblas_zdscal_batched";
+
+template <typename T, typename U>
+rocblas_status rocblas_scal_batched_impl(rocblas_handle handle,
```
TorreZuk: _impl suffix convention I see in blas2 but nowhere else in blas1.   Do we want it here?

clients/common/rocblas_gentest.py
leekillough: Also, the `all()` test can be done once outside of the `if`, or as separate a nested `if` controlling multiple cases, rather than repeating it.

clients/include/testing_scal_batched.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+        gpu_time_used = get_time_us(); // in microseconds
+
+        for(int iter = 0; iter < number_hot_calls; iter++)
+        {
+            rocblas_scal_batched<T, U>(handle, N, &h_alpha, dx_1, incx, batch_count);
+        }
+
+        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops    = axpy_gflop_count<T>(N) / gpu_time_used * 1e6 * 1;
```
leekillough: `axpy_gflop_count`?

library/src/blas1/rocblas_scal.hpp
```diff
@@ -0,0 +1,83 @@
...
+template <>
+constexpr char rocblas_scal_name<double>[] = "rocblas_dscal";
+template <>
+constexpr char rocblas_scal_name<rocblas_float_complex>[] = "rocblas_cscal";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex>[] = "rocblas_zscal";
+template <>
+constexpr char rocblas_scal_name<rocblas_float_complex, float>[] = "rocblas_csscal";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex, double>[] = "rocblas_zdscal";
```
leekillough: These need to be declared `static` if in a header file. `constexpr` template variables have external linkage by default in a retroactively-applied decision made earlier this year. We don't want to use unnamed namespaces in headers, so that leaves `static` (in C++17 the `constexpr` makes them implicitly `inline`, meaning they can be repeatedly defined if identically, but that doesn't apply to C++14).

library/src/blas1/rocblas_scal_batched.hpp
leekillough: > _impl suffix convention I see in blas2 but nowhere else in blas1. Do we want it here?

`_impl` is used for a C++ templated implementation with logging and argument error checking and device memory allocation (if any) turned on.

`_template` is used for a C++ templated implementation without logging, argument error checking, or device memory allocation, so that `_template` can be called from other rocBLAS without incurring overhead.

`#2` in: https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.github/CONTRIBUTING.md

To be consistent, `_impl` should probably be used everywhere in BLAS1.

library/src/blas1/rocblas_scal_strided_batched.hpp
```diff
@@ -0,0 +1,109 @@
...
+constexpr char rocblas_scal_name<rocblas_float_complex>[] = "rocblas_cscal_strided_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex>[] = "rocblas_zscal_strided_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_float_complex, float>[]
+    = "rocblas_csscal_strided_batched";
+template <>
+constexpr char rocblas_scal_name<rocblas_double_complex, double>[]
+    = "rocblas_zdscal_strided_batched";
+
```
leekillough: `static` needed in header files.

library/src/blas1/scal_device.hpp
```diff
@@ -0,0 +1,40 @@
...
+                                            T*          xa,
+                                            rocblas_int offsetx,
+                                            rocblas_int incx,
+                                            rocblas_int stridex)
+{
+    T* x = xa + hipBlockIdx_y * stridex + offsetx;
+    scal_kernel<T, U>(n, alpha_device_host, x, incx);
+}
+
+#endif // \IncludeGuard
```
leekillough: We have been gradually moving away from `_device` headers, and putting the `__device__` code in the `.cpp` files instead.

library/src/blas1/scal_host.hpp
leekillough: If blocksize is added, it should be the first, not the last, template parameter, so that later ones which are types can still be deduced automatically.

library/src/blas1/scal_host.hpp
```diff
@@ -0,0 +1,82 @@
...
+                           0,
+                           rocblas_stream,
+                           n,
+                           *alpha,
+                           x,
+                           offsetx,
+                           incx,
+                           stridex);
+
+    return rocblas_status_success;
```
leekillough: Why is `scal` implementation duplicated between batched and non-batched? Couldn't non-batched just be expressed with a `bool` template parameter (before any type parameters which can be deduced automatically), so that code is not duplicated and 90% of the batched and non-batched code is shared? With C++17 you can use `constexpr if`, but we're still using C++14, but as long as both halves of an `if` with a constant template parameter are compileable, the compiler optimizes away the false one.

leekillough:
>     * General file layout (as per Juan's document and discussion of exposing offset for internal use):
>       
>       * rocblas_scal.cpp, rocblas_scal_batched.cpp, etc. contain C calls
>       * rocblas_scal.hpp, rocblas_scall_batched.hpp, etc. contain logging and error checking
>       * scal_host.hpp contains quick return and device calls
>       * scal_device.hpp contains kernels and device calls  
 
Where is this "Juan's document", and does it conflict with [CONTRIBUTING.md](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.github/CONTRIBUTING.md)?

We have been moving away from `_host` and `_device` filenames, and implementations in header files.

**We also want to merge batched and non-batched implementations as much as possible** (for example, see `trtri`). If we need to distinguish between them, we can use template parameters (but we cannot use `constexpr if` just yet, since we're restricted to C++14).

We do not want to define variables in header files, but if we do, and they are `template` variables, they need to be declared `static`. If they're not `template` variables, they need to be declared `const`, `constexpr`, or `static`. (In C++17 they can be `inline`, but we are still using only C++14.)

library/src/blas1/rocblas_scal_strided_batched.hpp
leekillough: If `rocblas_scal_strided_batched.hpp` is only used in `rocblas_scal_strided_batched.cpp`, then it should be hand-inlined into `rocblas_scal_strided_batched.cpp`, and `rocblas_scal_strided_batched.hpp` should be removed. Same for non-strided-batched.

Those changes were recently made wholesale across BLAS in the last few months, so this BLAS1 change would be undoing those changes.

We also don't want to define variables in headers, and we do not want a header file to `#include` all of the other headers needed by the corresponding `.cpp` file -- the `.cpp` and `.hpp` files should each only `#include` what they need, and one of them should not rely on another to `#include` files it needs. We have a script which *enforces* this at the header-file level now.

We also want to use templates as much as possible to reduce duplicate code. `batched` and `strided_batched` and non-batched should all be able to be templatized into a common code,.

clients/gtest/blas1_gtest.yaml
daineAMD: Good call, added.

clients/include/testing_scal_batched.hpp
daineAMD: Good catch, fixed.

clients/include/testing_scal_batched.hpp
daineAMD: This comment seems to be everywhere in the blas1 testing files. I'm not sure why it would not work for incx != 1 as, in this case, size_x includes the incx value. Also, we have tests which test for values besides incx == 1. I'm guessing this has been fixed/can be removed but I'm unsure of the original reasoning.

library/src/blas1/rocblas_scal_batched.hpp
daineAMD: As we discussed earlier, the offset (or shift) parameter can be used in batched calls internally to use only subsections of data in each batch. This can be done manually in strided_batched and regular calls, but not as easily in batched calls. I'm not sure of any current use for functions such as scal, but if you look at trsm, it uses offsets when it makes calls to gemm.

leekillough: Why not just pass a pointer to the offseted data, rather than a pointer and offset?

clients/include/testing_scal_batched.hpp
leekillough: I'm guessing it was left over from before my mass-refactoring of `rocblas-test`.

If it works now, and passes tests for which the comment contradicts, we can remove the comment.

library/src/blas1/rocblas_scal_batched.hpp
daineAMD: The batched version uses an array of pointers (for each batch). The offset parameter will offset each batch after being called, rather than having to create an array to the offset data for each batch.

library/src/blas1/scal_host.hpp
daineAMD: The non-batched version currently uses the strided_batched call at the template layer. I have different templates for the batched version and strided_batched version. After talking with Juan, I can move these to a single template file using the preprocessor.

clients/common/rocblas_gentest.py
daineAMD: Rearranged statements.

clients/common/rocblas_gentest.py
```diff
@@ -196,13 +196,13 @@ def setdefaults(test):
     # These are only for dynamic defaults
     # TODO: This should be ideally moved to YAML file, with eval'd expressions.
 
-    if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
+    if (test['function']=='ger_strided_batched' or test['function']=='scal_strided_batched') and all([x in test for x in ('M', 'incx', 'strideScale')]):
         test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
                                     test['strideScale']))
```
leekillough: I would reduce the line lengths to less than 80 (with small exceptions), and use `in` for list membership instead of Boolean expressions. I use `flake8` (a Python `lint`) to validate Python scripts.
```
  if (all([x in test for x in ('M', 'incx', 'strideScale')]):
      if test['function'] in ('ger_strided_batched', 'scal_strided_batched'):
          test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
                                      test['strideScale']))
``` 

Ideally the hierarchical structure of the `if` statements should reflect the structure of the program, and the same test should not be repeated twice in two different places.

clients/include/testing_scal_strided_batched.hpp
daineAMD: In the case that we have the assumption stridex >= N * incx, we can remove the last overhang of stridex (that is, stridex - (N * incx) ) by using this formula. I will add a comment. If we decide against enforcing stridex to be a certain size this can be changed.

library/src/blas1/scal_device.hpp
daineAMD: Removed the _device header, moved the device code into shared rocblas_scal.hpp file (see caacbf0)

library/src/blas1/scal_host.hpp
daineAMD: Created one shared rocblas_scal_template function using the preprocessor to differentiate batched vs strided_batched/non-batched calls. (see caacbf0)

library/src/blas1/scal_host.hpp
daineAMD: Added blocksize as template parameter (with trtri as inspiration). (see 
9fe93bb)

clients/include/testing_scal.hpp
```diff
@@ -62,6 +62,10 @@ void testing_scal(const Arguments& arg)
         }
 
         CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
+        // if(incx <= 0)
+        //     EXPECT_ROCBLAS_STATUS(
+        //         (rocblas_scal<T, U>)(handle, N, &h_alpha, dx, incx), rocblas_status_invalid_size);
+        // else
```
leekillough: `incx <= 0` is not an error for `SCAL`, but just makes it act as a no-op. Rather than doing a separate test and early return, perhaps it should just call `SCAL` normally with `incx <= 0` and then verify that the result has not changed. The case here would be limited to `n <= 0` but `incx <= 0` would fall through to the code which calls CBLAS and compares the results.

library/src/blas1/rocblas_scal.hpp
```diff
@@ -0,0 +1,111 @@
+/* ************************************************************************
```
leekillough: I thought we were merging the `.hpp` files into the `.cpp` files. It will be easier to maintain if the `*_kernel` and `*_device` functions are in the same `.cpp` file as the C and C++ interfaces (the C, `*_impl` and `*_template` functions).

If `*_template` needs to be called from another file, it does not need to be defined (just declared) in a header as long as the template is instantiated with the right arguments in the `.cpp` file.

library/src/blas1/rocblas_scal.hpp
```diff
@@ -0,0 +1,111 @@
...
+    if(handle->is_device_memory_size_query())
+    {
+        if(rocblas_pointer_mode_host == handle->pointer_mode && inca != 0)
+            return handle->set_optimal_device_memory_size(sizeof(V) * batch_count * inca);
+        else
+            return rocblas_status_size_unchanged;
+    }
+
+    if(n <= 0 || incx <= 0 || batch_count <= 0)
+        return rocblas_status_success;
```
leekillough: Redundant with the above check.

library/src/blas1/rocblas_scal.hpp
TorreZuk: I would vote against merging the templates into .cpp if we want to promote reuse or customization.  If one wants to derive a new variant they would want to include an .hpp.

daineAMD: Since the _template function is being called from `rocblas_scal_impl` (in rocblas_scal.cpp), `rocblas_scal_batched_impl` (in rocblas_scal_batched.cpp) and `rocblas_scal_strided_batched_impl` (in rocblas_scal_strided_batched.cpp) doesn't it make more sense to have one .hpp file which contains the shared code?

I think it would be easier to maintain by having a separate .hpp file with the shared functions, rather than putting them into one specific (non-batched) .cpp file, but perhaps I'm missing a disadvantage to this.

library/src/blas1/rocblas_scal.hpp
```diff
@@ -0,0 +1,111 @@
...
+                                     U              x,
+                                     rocblas_int    offsetx,
+                                     rocblas_int    incx,
+                                     rocblas_int    stridex,
+                                     rocblas_int    batch_count)
+{
+    // Quick return if possible. Not Argument error
+    if(n <= 0 || incx <= 0 || batch_count <= 0)
+    {
+        if(handle->is_device_memory_size_query())
```
leekillough: Note that `*_template` functions are not supposed to allocate device memory or do argument error-checking. If they do that, then they cannot be called from other rocBLAS functions (besides their C wrappers and/or other wrappers).

`<name>_impl`: Argument error-checking, logging, and device memory allocation. Called from `extern "C"` wrappers.

`<name>_template`: No error-checking, logging, or device memory allocation. If device memory is needed, it is passed by pointer. Can be called from other rocBLAS routines, which are responsible for allocating device memory.

The separation of `<name>_impl` and `<name>_template` is to allow you to call `<name>_template` from other rocBLAS code without duplicating device memory allocation (which, for speed, is only allowed once per rocBLAS call), without duplicating log entries of routines the rocBLAS user didn't call directly, and to avoid the overhead of arugment-checking, which should only be done at the highest-level routine.

In general, `.hpp` files for implementing `rocBLAS` should be avoided except when the kernel is needed outside of the `.cpp` file, in which case, a declaration rather than definition is preferred. The `.cpp` file should instantiate all necessary template argument combinations, either implicitly as with the C wrappers, or explicitly in case they're needed outside of the `.cpp` file.

clients/common/rocblas_gentest.py
daineAMD: Addressed in 077ef48.

clients/include/testing_scal.hpp
daineAMD: Removed commented-out code. incx <= 0 still does a separate test and returns early. If this does not happen, incx is used for memory alloc, etc which causes issues.

library/src/blas1/rocblas_scal.hpp
daineAMD: Fixed in 077ef48.

library/src/blas1/rocblas_scal.hpp
leekillough: > I would vote against merging the templates into .cpp if we want to promote reuse or customization. If one wants to derive a new variant they would want to include an .hpp.

**That's only true for class templates / containers / etc.**

Reuse can be done even if they are in the `.cpp` file, _for function templates_. 

You simply declare (not define) the function template in the `.hpp` file, and then define it in the `.cpp` file. You can explicitly instantiate it for whatever template arguments in needs in the `.cpp` file, but if they are the same template arguments as the common `extern "C"` wrappers' template arguments, you won't even need to explicitly instantiate them, as they will already be implicitly instantiated.

You then don't need to create the `.hpp` file until a function is needed outside of the `.cpp` file, in which case you create an `.hpp` file and declare, but don't define, the function template there.

For most rocBLAS routines, they aren't called from outside of their implementation, so keeping everything in the `.cpp` file leads to more compact code which is easier to maintain, since it's all in one place rather than scattered among several files.

leekillough: I wouldn't be opposed to declaring the template function in the `.hpp` file, even if it's not currently used outside of the `.cpp` file. I just don't like the definitions of the kernel functions being scattered among `.hpp` and `.cpp` files.

daineAMD: For this PR, all three functions (`rocblas_scal`, `rocblas_scal_batched`, and `rocblas_scal_strided_batched`) all share a single kernel which is defined in the shared .hpp file. This follows how the reduction functions are defined as well - in a shared .h file in that case - along with other functions such as `rocblas_trtri` (rocblas_trtri_batched.hpp).
Moving forward, if we are to implement more batched/strided_batched functions, kernels will likely start to be shared more (between the batched/strided/non-batched versions of functions), moving .cpp implementation of kernels to .hpp files might be a good standard to start to address this.

leekillough: > Since the _template function is being called from `rocblas_scal_impl` (in rocblas_scal.cpp), `rocblas_scal_batched_impl` (in rocblas_scal_batched.cpp) and `rocblas_scal_strided_batched_impl` (in rocblas_scal_strided_batched.cpp) doesn't it make more sense to have one .hpp file which contains the shared code?
> 
> I think it would be easier to maintain by having a separate .hpp file with the shared functions, rather than putting them into one specific (non-batched) .cpp file, but perhaps I'm missing a disadvantage to this.

It could be separated out into a `rocblas_scal_template.cpp` file, but the only advantage over putting it into a `.hpp` file is that compile/link time might be reduced if the same template arguments are used more than once; the functionality remains the same as if it is kept in a `.hpp` file.

**If each `.cpp` file instantiates the template with _different_ template arguments, then each `.cpp` file will generate its own instantiation of the template, and there is no advantage to moving it from a `.hpp` file to a `.cpp` file.** If two or more `.cpp` files instantiate the template with the _same_ template arguments, then each `.cpp` file will have to reinstantiate and rebuild the template, while if the template is only instantiated in one `.cpp` file, it will only be instantiated and built once for each set of arguments, and then it is also available for other `.cpp` files to use.

As long as the calls of the template are already scattered among several `.cpp` and `.hpp` files, then unless it's easy and not too large to merge them into one `.cpp` file, then my concerns about scattered implementations will still remain, regardless of whether you put the `_template` function in a `.hpp` file or a `.cpp` file. 

Putting `_template` in a `.hpp` file:
- Advantages: Declaration = definition; possibly one less file
- Disadvantages: Rebuilds the template every time it is used; separates the implementation in the `.hpp` file away from the `.cpp` file(s) which use it

Putting `_template` in a `.cpp` file:
- Advantages: Builds the template only once for each set of template arguments, possibly reducing compile time; still allows it to be called from any `.cpp` file; allows putting all implementation variants of the kernel code into one `.cpp` file
- Disadvantages: Still requires a `.hpp` file declaring the template if it's used elsewhere, so possibly one extra file; code may be too large if merged into one `.cpp` file, causing the same file to be edited more often than if the files were separate

So, as long as there are multiple `.cpp` files using the same `.hpp` file, I'm neutral about keeping the `.hpp` file.

leekillough:
> 
> 
> Questions:
> 
>     * why is it defined only if GOOGLE_TEST is undefined?

Not speaking about this change in particular, but in other parts of  `rocblas-test`, we only define things if `GOOGLE_TEST` is undefined, because it would cause build errors if `GOOGLE_TEST` is defined. Sometimes we do the opposite, and only define something if `GOOGLE_TEST` is defined, such as use its macros. The Google Test framework is complicated, and we sometimes have to work around it at the fringes.

Note that everything under `clients/*` is basically our in-house code, and not for end-users to use, so we do not need to document it, or make it free of dependencies like Boost or Google Test. We just need to have reasonable code comments and a reasonable `--help` command-line message.

>     * Is 'load' the best naming? why not simply 'get' ? 'block' with 'batch_index' ? Naming is important.

`load`, `get`, `fetch`, are all acceptable names. `load` was partially chosen because, like `load_scalar`, we are taking different actions based on whether the argument is a pointer to our result type or not -- in this case, we either load from the pointer-to-pointer, or we copy the already-loaded pointer.
 
> Suggestion:
> 
>     * Could we have a proper 'doxygen' format documention?

We don't use doxygen for every internal class, function, variable, type, macro, etc. Only public types and APIs, which use C, are documented. Namely, stuff in `library/include/*`. Not `clients/*` or `library/src/*` (with a few exceptions, like sample code).
 
>     * Why don't we have a namespace for utility functions? RocblasUtils for instance.

Right now rocBLAS is, except for a few names which use `namespace rocblas`, a global-namespace library, since it has a C API.

It goes into a shared library for which only certain symbols are exported, so there is no prevention of the contamination of other namespaces if we use our own namespaces, since the symbols will still not be exported.

Also, using our own namespaces makes the code more complicated, requiring `using namespace RocblasUtils;` or `RocblasUtils::name`, while right now we just use `name`.

YvanMokwinski:
> ot speaking about this change in particular, but in other parts of `rocblas-test`, we only define things if `GOOGLE_TEST` is undefined, because it would cause build errors if `GOOGLE_TEST` is defined. Sometimes we do the opposite, and only define something if `GOOGLE_TEST` is defined, such as use its macros. The Google Test framework is complicated, and we sometimes have to work around it at the fringes.
> 
> Note that everything under `clients/*` is basically our in-house code, and not for end-users to use, so we do not need to document it, or make it free of dependencies like Boost or Google Test. We just need to have reasonable code comments and a reasonable `--help` command-line message.

This is not in the CLIENT size, the file is library/src/include/utility.h. Despite that Google Test framework is complicated, why does it need to work around with GOOGLE_TEST in this file? 
"Reasonable" means no documentation. IMHO, we need clean internal documentation and doxygen is one way to do it correctly.

YvanMokwinski:
> We don't use doxygen for every internal class, function, variable, type, macro, etc. Only public types and APIs, which use C, are documented. Namely, stuff in `library/include/*`. Not `clients/*` or `library/src/*` (with a few exceptions, like sample code).

Can we have a more systematic internal documentation?

YvanMokwinski:
> Also, using our own namespaces makes the code more complicated, requiring `using namespace RocblasUtils;` or `RocblasUtils::name`, while right now we just use `name`.

There is trade-off between complicated and a certain structure.

leekillough:
> > ot speaking about this change in particular, but in other parts of `rocblas-test`, we only define things if `GOOGLE_TEST` is undefined, because it would cause build errors if `GOOGLE_TEST` is defined. Sometimes we do the opposite, and only define something if `GOOGLE_TEST` is defined, such as use its macros. The Google Test framework is complicated, and we sometimes have to work around it at the fringes.
> > Note that everything under `clients/*` is basically our in-house code, and not for end-users to use, so we do not need to document it, or make it free of dependencies like Boost or Google Test. We just need to have reasonable code comments and a reasonable `--help` command-line message.
> 
> This is not in the CLIENT side, the file is library/src/include/utility.h. Despite that Google Test framework is complicated, why does it need to work around with GOOGLE_TEST in this file?
> "Reasonable" means no documentation. IMHO, we need clean internal documentation and doxygen is one way to do it correctly.

It is client side. When building client with Google Test, some parts of the library header must be disabled in order to compile client.

It does not affect the build of the library, since GOOGLE_TEST is not defined during library builds. But during client builds, GOOGLE_TEST may be defined, and that interferes with part of the library headers, so we leave that out during the client builds, without affecting either the client build or the library build.

leekillough:
> Can we have a more systematic internal documentation?

That's something for the team to discuss, as far as rocBLAS documentation guidelines, but if you just need doxygen syntax documentation, possibly there are others who can point you to an online reference.

leekillough:
> There is trade-off between complicated and a certain structure.

Right now we use a simple structure similar to rocBLAS's C API, with a shared library exporting only those symbols which need to be public.

leekillough:
> It does not affect the build of the library, since GOOGLE_TEST is not defined during library builds. But during client builds, GOOGLE_TEST may be defined, and that interferes with part of the library headers, so we leave that out during the client builds, without affecting either the client build or the library build.

Note that the rocBLAS client tests / benchmarks use library header files which are not part of the "public library API". You may be wondering why a client (an internal one) uses `library/src/include/utility.h`, but it is to avoid code duplication and to ensure the client uses the exact same definitions as the library does. When the client includes that part of the library, it disables the parts which it doesn't need and which conflict with GOOGLE_TEST client builds.

clients/gtest/blas1_gtest.yaml
```diff
@@ -21,70 +25,193 @@ Definitions:
...
+  - name: blas1_strided_batched
+    category: nightly
+    N: [ 4000000 ]
+    incx_incy: *incx_incy_range_small
+    alpha_beta: *alpha_beta_range
+    alphai_betai: *alphai_betai_range
+    batch_count: [ 2 ]
+    strideScale: [ 1.2 ]
+    function:
+      - scal_batched: *single_double_precisions_complex_real
```
mahmoodw: I think this is meant to be strided_batched?

daineAMD: Good catch, fixed in e906818.

daineAMD:
I believe that everyone's concerns have been addressed, and all of our questions regarding the batched/strided_batched versions have been addressed as well. If anyone has any remaining comments let me know, if not if I can get approval I'll finally get this merged and we can move on with finishing up and merging some of the other batched functions being worked on.

clients/include/testing_scal_strided_batched.hpp
```diff
@@ -0,0 +1,216 @@
...
+        rocblas_status_invalid_pointer);
+    EXPECT_ROCBLAS_STATUS((rocblas_scal_strided_batched<T, U>)(handle,
+                                                               N,
+                                                               &h_alpha,
+                                                               nullptr,
+                                                               incx,
+                                                               stridex,
+                                                               batch_count),
+                          rocblas_status_invalid_pointer);
+}
```
leekillough: Missing newline at EOF

clients/include/testing_scal_batched.hpp
```diff
@@ -0,0 +1,223 @@
...
+        CHECK_HIP_ERROR(hipErrorOutOfMemory);
+        return;
+    }
+
+    EXPECT_ROCBLAS_STATUS((rocblas_scal_batched<T, U>)(handle, N, nullptr, dx, incx, batch_count),
+                          rocblas_status_invalid_pointer);
+    EXPECT_ROCBLAS_STATUS(
+        (rocblas_scal_batched<T, U>)(handle, N, &h_alpha, nullptr, incx, batch_count),
+        rocblas_status_invalid_pointer);
+}
```
leekillough: Missing newline at EOF

library/src/blas1/rocblas_scal.hpp
```diff
@@ -0,0 +1,107 @@
...
+                                     rocblas_stride stride_alpha,
+                                     U              x,
+                                     rocblas_int    offsetx,
+                                     rocblas_int    incx,
+                                     rocblas_stride stridex,
+                                     rocblas_int    batch_count,
+                                     V*             mem)
+{
+    // Memory queries must be in template as _impl doesn't have stride_alpha parameter (for calls from
+    // outside of rocblas)
```
leekillough: > // Memory queries must be in template as _impl doesn't have stride_alpha parameter (for calls from
>  // outside of rocblas)

However, this prevents `rocblas_scal_template` from being called from any other _arbitrary_ rocBLAS routine.

Only one device memory allocation is allowed to be active per rocBLAS C API call, and it must allocate for itself and all of its children.

If another routine were to allocate device memory through rocBLAS, directly or indirectly, then it could not call `rocblas_scal_template`, unless that's the only routine in its call tree which allocates device memory, or unless it frees its allocated memory before calling `rocblas_scal_template`. 

 (`hipMalloc()` and `hipFree()` should not be called from the rocBLAS library except in certain ways, because they are synchronizing operations.)

There are runtime safeguards in the library to detect violations of these restrictions, with an internal error message and an `abort()`.

As long as the test pass right now, the error is not happening, but beware of this restriction.

leekillough:
Approving, pending linefeeds at the end of 2 files, and non-SLES tests passing.

PR #692, opened 2019-09-11
Mintests

YvanMokwinski:
Put back the ignored test on iamin (see the JIRA ticket SWDEV-201525) 
and clean clients/include/rocblas.hpp.

Summary of proposed changes:
- 1/ clients/gtest/blas1_gtest.yaml: uncomment ignored tests

- 2/ clients/include/rocblas.hpp: 
    - a/ remove unnecessary full functions iamax/iamin (see modifications on testing_iamax_iamin.hpp)
    - b/ instead, define the references like ALL the other functions. 

- 3/ clients/include/testing_iamax_iamin.hpp
    - a/ no need for the function to be a template parameter, causing the issue of having to define full functions in rocblas.hpp. Instead, pass the function pointer as a function parameter.

- 4/ library/src/blas1/reduction.h
   - a/ Specify the work group size attribute to remove the non-deterministic behavior causing the iamin test failure. (see the JIRA ticket SWDEV-201525). The min is 128 by default and the max is 256 by default, only changing these values if NB is out of bounds.

### DIFF
--- a/clients/include/testing_iamax_iamin.hpp
+++ b/clients/include/testing_iamax_iamin.hpp
@@ -40,17 +40,16 @@ void testing_iamax_iamin_bad_arg(const Arguments& arg, rocblas_iamax_iamin_t<T> 
 template <typename T>
 void testing_iamax_bad_arg(const Arguments& arg)
 {
-  testing_iamax_iamin_bad_arg<T>(arg, rocblas_iamax<T>);
+    testing_iamax_iamin_bad_arg<T>(arg, rocblas_iamax<T>);
 }
 
 template <typename T>
 void testing_iamin_bad_arg(const Arguments& arg)
 {
-  testing_iamax_iamin_bad_arg<T>(arg, rocblas_iamin<T>);
+    testing_iamax_iamin_bad_arg<T>(arg, rocblas_iamin<T>);
 }
 
-template <typename T,
-          void CBLAS_FUNC(rocblas_int, const T*, rocblas_int, rocblas_int*)>
+template <typename T, void CBLAS_FUNC(rocblas_int, const T*, rocblas_int, rocblas_int*)>
 void testing_iamax_iamin(const Arguments& arg, rocblas_iamax_iamin_t<T> func)
 {
     rocblas_int N    = arg.N;
@@ -257,11 +256,11 @@ namespace rocblas_cblas
 template <typename T>
 void testing_iamax(const Arguments& arg)
 {
-  testing_iamax_iamin<T, rocblas_cblas::cblas_iamax<T>>(arg, rocblas_iamax<T>);
+    testing_iamax_iamin<T, rocblas_cblas::cblas_iamax<T>>(arg, rocblas_iamax<T>);
 }
 
 template <typename T>
 void testing_iamin(const Arguments& arg)
 {
-  testing_iamax_iamin<T, rocblas_cblas::cblas_iamin<T>>(arg, rocblas_iamin<T>);
+    testing_iamax_iamin<T, rocblas_cblas::cblas_iamin<T>>(arg, rocblas_iamin<T>);
 }
--- a/clients/include/testing_iamax_iamin.hpp
+++ b/clients/include/testing_iamax_iamin.hpp
@@ -13,9 +13,8 @@
 #include "unit.hpp"
 #include "utility.hpp"
 
-template <typename T,
-          rocblas_status (&FUNC)(rocblas_handle, rocblas_int, const T*, rocblas_int, rocblas_int*)>
-void testing_iamax_iamin_bad_arg(const Arguments& arg)
+template <typename T>
+void testing_iamax_iamin_bad_arg(const Arguments& arg, rocblas_iamax_iamin_t<T> func)
 {
     rocblas_int         N         = 100;
     rocblas_int         incx      = 1;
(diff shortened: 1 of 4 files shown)


clients/include/testing_iamax_iamin.hpp
```diff
@@ -13,9 +13,8 @@
 #include "unit.hpp"
 #include "utility.hpp"
 
-template <typename T,
-          rocblas_status (&FUNC)(rocblas_handle, rocblas_int, const T*, rocblas_int, rocblas_int*)>
-void testing_iamax_iamin_bad_arg(const Arguments& arg)
+template <typename T>
+void testing_iamax_iamin_bad_arg(const Arguments& arg, rocblas_iamax_iamin_t<T> func)
```
leekillough: There are other places in rocBLAS where the testing function is assumed to contain only a single `Arguments&` parameter. That's why a full function was passed as a template argument, instead of passing a function pointer as a regular argument.

Other code which assumes all tests consist of a single `Arguments&` parameter, are likely to be broken.

YvanMokwinski: testing_ia{max|min} and testing_ia{max|min}_bad_arg are the tests used and take a single Arguments&.
Is testing_iamax_iamin_bad_arg used anywhere? I might be wrong but I think it is only used in testing_ia{max|min} and testing_ia{max|min}_bad_arg.

YvanMokwinski: Should we rename testing_iamax_iamin_bad_arg and testing_iamax_iamin ? There is no routine rocblas_iamax_iamin.

leekillough: Here's places where the single `const Arguments&` parameter is assumed (sorry I left `const` out earlier):

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/c924b9b253ed8cc9f1fc12bdf2c41458d165cc0a/clients/include/type_dispatch.hpp#L28

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/c924b9b253ed8cc9f1fc12bdf2c41458d165cc0a/clients/gtest/blas1_gtest.cpp#L193

You can write templated wrapper functions in `testing_iamax_iamin.hpp` which take a single `Arguments& arg` parameter, and call another function with that `arg` argument and a separate function pointer. In that way, it's similar to the original code, but with the wrapper moved to the wraping the test function rather than wrapping the library function. It still might look cleaner.

Or maybe I'm missing something, and you already created such wrappers.

leekillough: > 
> 
> Should we rename testing_iamax_iamin_bad_arg and testing_iamax_iamin ? There is no routine rocblas_iamax_iamin.

`rocblas_iamax_iamin` is a wrapper function to either `rocblas_iamax` or `rocblas_iamin`.

`*_bad_arg` is shown above with the macro token pasting `testing_##NAME##_bad_arg`.

YvanMokwinski: The original version of testing_iamax_iamin_bad_arg takes 3 template arguments (!). It is only a common piece of code for testing_iamax_bad_arg and testing_iamin_bad_arg. There is no routine rocblas_iamax_iamin in the API, thus testing_iamax_iamin_bad_arg is only a private common code function, used to avoid code redundancy, there is no way that the type_dispatch  would call this function with the TEST<float>{}(arg). This routine could be renamed common_code_of_iamax_iamin_bad_arg without any consequence. 
Thus, the full functions in rocblas.hpp is unnecessary and make unnecessary complications.

leekillough: The template test functions which take a single `const Arguments&` parameter take an arbitrary number of `typename` template arguments, as expressed with variadic templates. The code in `type_dispatch.hpp` and elsewhere is meant to work with functors (classes with `operator()`) which take an arbitrary number of template arguments of `typename`.

The template test functions which can take a variable number of arguments are expected to make later arguments default to earlier ones, so that `type_dispatch.hpp` can call them with as few as one type argument, and the others are defaulted to it. So if you call `iamax_iamin_bad_arg` with only one template argument, the others are assumed to match the first.

In the old code, `iamax_iamin_bad_arg` was used to test both `iamax` and `iamin`, as differentiated by a template arguments. The name was chosen to emphasize that both `iamin` and `iamax` could be tested by the same code. We don't want to copy-and-paste that whole function and have two copies of it -- we can either use a template argument or a function argument to distinguish between `iamax` and `iamin`.

As I said, one way to do it is to use wrapper functions, so you can have a `testing_iamax_bad_arg` and `testing_iamin_bad_arg` and then both of them call a common function `testing_iamax_iamin_bad_arg` with different arguments to distinguish between `iamax` and `iamin`. 

The naming of the test functions is not critical, except that they are used in `blas1_gtest.cpp` when forming names with the `##` tokenizing operator.

The `testing_*_bad_arg` functions are separate from the normal tests, and are not called by the normal tests, but are called by the test driver in `blas1_gtest.cpp` when the function name matches `_bad_arg` suffix.

YvanMokwinski: There is no function "iamax_iamin" to test, only "iamax" or "iamin". Thus, there is no constraint of having one Argument& parameter testing_iamax_iamin_bad_arg, which should anyway be renamed without "testing_" for clarity. Thus, full functions of rocblas.hpp must be removed.

leekillough: > There is no function "iamax_iamin" to test, only "iamax" or "iamin". Thus, there is no constraint of having one Argument& parameter for testing_iamax_iamin_bad_arg, which should anyway be renamed without "testing_" for clarity. Thus, full functions of rocblas.hpp must be removed rather than exposing a pointless difference.

It does **not** follow that full functions in `rocblas.hpp` _must_ be removed; that's an orthogonal issue. The only reason that they were there, was because they were being passed as template arguments; arbitrary function pointers cannot be passed as template arguments.

If you add extra parameters to the functions in `testing_iamax_iamin.hpp`, then you need to write wrappers which take only a single `const Arguments&` parameter, so that you can call them from the rest of the test infrastructure, like `type_dispatch.hpp`, without changing the rest of the test infrastructure.

The functions called from the rest of the test infrastructure are expected to have one or more type template arguments, and under the convention used in `type_dispatch.hpp`, at most one argument is needed -- the extra template type arguments are expected to have defaults based on previous type arguments, so that for a function with the same input, output, compute, etc. types, only the first type has to be specified, and the rest will default to it.

The `iamax_iamin` name (or sub-name) is used to indicate that the function is used for both `iamax` and `iamin`, without specifying how they are differentiated. It is a self-documenting code writing technique. Just calling it `iamax` or `iamin` would appear to limit it to one function even though it can apply to both.

YvanMokwinski: > The only reason that they were there, was because they were being passed as template arguments; arbitrary function pointers cannot be passed as template arguments.

I agree, on everything and on the fact, that the function does not need to be a template argument. It is a trick and the full functions are there to keep the use of the trick.

leekillough: Also, `testing_` is used as a prefix throughout the entire rest of the rocBLAS test framework, so `iamax` and `iamin` would be outliers if they did not use `testing_` in their function prefixes or header filenames.

You can write `testing_iamax_bad_arg`, `testing_iamax`, `testing_iamin_bad_arg` and `testing_iamin`, as the testing functions.

For implementing `testing_iamax_bad_arg` and `testing_iamin_bad_arg`, you can call a common function `testing_iamax_iamin_bad_arg` with an extra argument to indicate `iamax` or `iamin`.

Similarly, for implementing `testing_iamax` and `testing_iamin`, you can call a common function `testing_iamax_iamin` with an extra argument to indicate `iamax` or `iamin`.

All of the template arguments would be the same, and could be passed from the wrappers to the implementation functions.

The "full functions" in `rocblas.hpp` would be a separate issue; they can be replaced with simple function pointers, or left as full functions. I don't object to replacing them with pointers if there's no more need for full functions, but that's not the main motivator; the main motivator is how `testing_iamax_iamin.hpp` gets implemented.

YvanMokwinski: Okay. I think we agree on everything. 

My proposal is a simplification of rocblas.hpp, make the file uniform, i.e. declare everything in the same manner, i.e. with static constexpr pointer references. 

In some aspect, the simplicity is important. This simplification can be done, i.e. using pointer references,  if we don't pass a function as a template parameter when calling testing_iamax_iamin to implement testing_iamax and testing_iamin.

PR #696, opened 2019-09-12
Torrezuk batched and strided nrm2 and asum

TorreZuk:
This adds into blas1 the batched and strided_batched for reduction based asum and nrm2 functions (shared templates).    Follows feedback and pattern from PR #686

It defers the merging of the batched and strided_batched kernels into a single more complicated template as it seemed complicated enough already.  It can likely reuse some code from scal_batched helpers if this refactor is done.   Also requires unification of testing parameter stride_scale with the scal routine (and others).   stride_x parameter checking will also require updating after we decide on API specification.

Example of timing runs:

36 tests from quick/asum_batched (75 ms total)
36 tests from pre_checkin/asum_batched (4006 ms total)
12 tests from nightly/asum_batched (3906 ms total)

216 tests from quick/asum_strided_batched (7132 ms total)
36 tests from pre_checkin/asum_strided_batched (272 ms total)
12 tests from nightly/asum_strided_batched (3743 ms total)

36 tests from quick/nrm2_batched (81 ms total)
36 tests from pre_checkin/nrm2_batched (4017 ms total)
12 tests from nightly/nrm2_batched (3973 ms total)

216 tests from quick/nrm2_strided_batched (7121 ms total)
36 tests from pre_checkin/nrm2_strided_batched (285 ms total)
12 tests from nightly/nrm2_strided_batched (3798 ms total)


library/src/blas1/rocblas_asum_batched.hpp
```diff
@@ -0,0 +1,56 @@
...
+template <rocblas_int NB, typename To>
+size_t rocblas_asum_batched_template_workspace_size(rocblas_int n,
+                                                    rocblas_int batch_count,
+                                                    To*         output_type)
+{
+    return rocblas_reduction_batched_kernel_workspace_size<NB>(n, batch_count, output_type);
+}
+
+template <rocblas_int NB, typename Ti, typename To>
+rocblas_status rocblas_asum_batched_template(rocblas_handle  handle,
```
daineAMD: I think it's probably worth it to combine the templates as well tbh, makes it a bit cleaner I think since they're identical.

library/src/blas1/rocblas_nrm2_batched.hpp
```diff
@@ -0,0 +1,70 @@
+/* ************************************************************************
```
daineAMD: Same comments as in asum.

library/src/blas1/reduction_batched.h
```diff
@@ -0,0 +1,286 @@
...
+// yet. rocBLAS still use the classic standard parallel reduction right now.
+
+// kernel 1 writes partial results per thread block in workspace; number of partial results is
+// blocks
+template <rocblas_int NB,
+          typename FETCH,
+          typename REDUCE = rocblas_reduce_sum,
+          typename Ti,
+          typename To>
+__global__ void rocblas_reduction_batched_kernel_part1(rocblas_int     n,
```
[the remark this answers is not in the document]
daineAMD: I notice that you mentioned most of this in your PR comments, I think it's worthwhile imo, but if others think not than we can talk!

library/src/blas1/reduction_batched.h
```diff
@@ -0,0 +1,286 @@
...
+
+    // bound
+    if(tid < n)
+        tmp[tx] = FETCH{}(x[tid * incx], tid);
+    else
+        tmp[tx] = default_value<To>{}(); // pad with default value
+
+    rocblas_reduction<NB, REDUCE>(tx, tmp);
+
+    __syncthreads();
```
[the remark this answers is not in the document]
TorreZuk: I will delete it and test, it must have been in the original but the recursive reduction already has a syncthreads after each recurse.

library/src/blas1/rocblas_asum_batched.hpp
TorreZuk: Now not sure as I think we might want to doxygen document the _template and keeping one per .cpp seems clearer to me than a single catch all.   If we make 3 reusable c++ template classes I would expect 3 .hpp files.   Can discuss.

leekillough: > 
> 
> Now not sure as I think we might want to doxygen document the _template and keeping one per .cpp seems clearer to me than a single catch all. If we make 3 reusable c++ template classes I would expect 3 .hpp files. Can discuss.

We do not need to document internal interfaces such as `*_impl` or `*_template`. Only public C interfaces need documentation. So only the stuff in `library/include/*` needs documenting, not the stuff in `library/src/*`. For the internal stuff, inline C/C++ comments are sufficient.

leekillough: > If we make 3 reusable c++ template classes I would expect 3 .hpp files.

There is no need for each template to be in its own `.hpp` file.

Templates which aren't used, don't affect program output, so if three templates are in the same header file, and only one of them is used, only the one which is used will appear in the output.

If the three template functions are somewhat related, it may be better to keep them in the same file for maintainability, rather than splitting them into separate headers, especially if the templates partially share some common code, which you don't want spread across several files.

As I said before, internal implementations do not need Doxygen documentation, just C/C++ comments.

So you can put 3 template functions in the same header, document them with only C/C++ comments, and then `#include` the header multiple times, only using a subset of it each time. Only the templates which are used (instantiated) will appear in the output.

library/src/blas1/rocblas_asum_batched.hpp
```diff
@@ -0,0 +1,56 @@
+/* ************************************************************************
```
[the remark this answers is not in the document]
leekillough: > It might be worthwhile to combine all of the .hpp files into a single .hpp file, if we are thinking about combining the templates in the future anyway.

That is a good idea, as long as the separate templates are somewhat related, and could later be refactored into simpler code (such as wrappers to a more general common template).

If a template in a `.hpp` file is not instantiated (used), it will not affect program output, just maybe compile time. So there's no problem with having multiple templates, some of which are unused, in the same header file.

clients/include/testing_asum_strided_batched.hpp
```diff
@@ -0,0 +1,239 @@
...
+    rocblas_int incx        = arg.incx;
+    rocblas_int stridex     = arg.stride_x;
+    rocblas_int batch_count = arg.batch_count;
+
+    double rocblas_error_1;
+    double rocblas_error_2;
+
+    rocblas_local_handle handle;
+
+    rocblas_int abs_incx = incx >= 0 ? incx : -incx;
```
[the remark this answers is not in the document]
leekillough: Yeah, `asum` does not have a second vector argument, so if `n <= 0` or `incx <= 0`, it's considered empty (return value 0). There's no need to compute `abs_incx`, since `incx > 0` is assumed, after the error-checks.
```
 sasum = 0.0e0
 stemp = 0.0e0
 IF (n.LE.0 .OR. incx.LE.0) RETURN
```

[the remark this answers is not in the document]
leekillough: > Since we're returning invalid_size if incx < 0, I don't think we need this variable.

BTW, `asum` and  `nrm2` should not return `invalid` if `n <= 0` or `incx <= 0`. They should return `0`.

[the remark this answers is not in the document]
TorreZuk: Are we sure we want to match that fortran code and cublas?  The old docs you sent say incx 0 is invalid, and also for single input functions incx must be > 0.       I get the argument for N but as incx can sometimes be negative so returning 0 may make you miss the parameter error, is improving the API an option?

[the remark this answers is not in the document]
leekillough: We should match the original BLAS specification (that it is in Fortran is incidental).

When vectors are used with other vectors or matrices, the vectors may have negative increments, meaning the vector is read from the end. This generally only involves adjusting the pointer to the vector to the end, while leaving the negative increment alone.

When vectors are used alone (the only non-scalar argument), then if they have a zero or negative size or increment, they behave as if they are zero-length, and the mathematical definition of the operation on a zero-length vector is used.

Matrix dimensions cannot have negative sizes; they are flagged as invalid. However, zero-sized matrix dimensions are usually allowed as long as other constraints are met.

clients/include/rocblas.hpp
```diff
@@ -145,6 +145,51 @@ static constexpr auto rocblas_asum<rocblas_float_complex, float> = rocblas_scasu
 template <>
 static constexpr auto rocblas_asum<rocblas_double_complex, double> = rocblas_dzasum;
 
+// asum_batched
+template <typename T1, typename T2>
+rocblas_status (*rocblas_asum_batched)(rocblas_handle  handle,
+                                       rocblas_int     n,
+                                       const T1* const x[],
+                                       rocblas_int     incx,
+                                       T2*             result,
```
mahmoodw: Reminder from an earlier email

"Where Legacy BLAS functions have return values, the return value is instead added as the last function argument. ... This applies to the following functions: xDOT, xDOTU, xNRM2, xASUM, IxAMAX, IxAMIN.

For the batch and strided batched functions, note that the last argument should be the return, not batch_count. "

clients/include/testing_asum_batched.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+    // Initial Data on CPU
+    rocblas_seedrand();
+    for(int i = 0; i < batch_count; i++)
+    {
+        hx[i] = host_vector<T1>(size_x);
+        rocblas_init<T1>(hx[i], 1, N, incx);
+    }
+
+    device_batch_vector<T1> dxvec(batch_count, size_x);
```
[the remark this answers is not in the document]
TorreZuk: The arrays are in device memory it is only the pointers on the host the same as non-batch pointers are on the host.   Right?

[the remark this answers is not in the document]
daineAMD: The arrays are in device memory and the pointers to such are on the host here, correct. I'm not sure which pointers you mean for non-batch pointers. I just think we should probably be clear that the `device_batch_vectors` and `device_vectors` cannot be used interchangeably, we don't want to get them confused. Just a nitpick though.

library/src/blas1/reduction_strided_batched.h
```diff
@@ -0,0 +1,386 @@
...
+    ********************************************************************/
+template <rocblas_int NB,
+          typename Ti,
+          typename FETCH,
+          typename REDUCE   = rocblas_reduce_sum,
+          typename FINALIZE = rocblas_finalize_identity,
+          typename TPtrX,
+          typename To,
+          typename Tr>
+rocblas_status rocblas_reduction_strided_batched_kernel(rocblas_handle __restrict__ handle,
```
daineAMD: I find it a bit confusing that these are named `_strided_batched` while they are used by all three scenarios. I would consider renaming - to me the suffix `_strided_batched` implies that it is tailored to `_strided_batched` functions, but maybe that's just me.

library/src/blas1/rocblas_asum.hpp
```diff
@@ -0,0 +1,45 @@
...
+    template <typename Ti>
+    __forceinline__ __device__ To operator()(Ti x, ptrdiff_t)
+    {
+        return {fetch_asum(x)};
+    }
+};
+
+// allocate workspace inside this API
+template <rocblas_int NB, typename Ti, typename To>
+rocblas_status rocblas_asum_template(rocblas_handle handle,
```
daineAMD: Since these templates are _exactly_ (or close to it) the same code, I still think it's a good idea to combine them into one, as Lee stated it does reduce future upkeep and I don't think it will add any complexity in this case.
At the very least we could use the `asum_strided_batched` template with a `batch_count` of 1 here. Unless we're concerned that that will degrade performance.

library/src/blas1/rocblas_nrm2.hpp
```diff
@@ -0,0 +1,58 @@
...
+// allocate workspace inside this API
+template <rocblas_int NB, typename Ti, typename To>
+rocblas_status rocblas_nrm2_template(rocblas_handle handle,
+                                     rocblas_int    n,
+                                     const Ti*      x,
+                                     rocblas_int    shiftx,
+                                     rocblas_int    incx,
+                                     To*            workspace,
+                                     To*            result)
+{
```
daineAMD: The same comment here as for asum. I don't think there's a problem in combining the templates, but if you have a concern let me know!

library/src/blas1/rocblas_asum.hpp
TorreZuk: All 3 call the same template kernel this style just exposes 3 internal APIs for reuse rather than 1 where you have to put in stride 0, and batch count 1 when not using those features.   Fewer parameters that match the C API makes more sense to me than having more parameters.   If we want to add another parameter variation later on do we keep adding on to a single generic one or make a new one?    Could merge later in another PR as well.

library/src/blas1/rocblas_nrm2.hpp
TorreZuk: same response as asum.   There is no problem to combine.

leekillough: There are many ways to combine them. I am speaking generally here, but you could add an extra template parameter which passes a class, or an extra template template parameter which passes a template of classes. If you're just trying to distinguish between `asum` and `nrm2`, you only need one type parameter to differentiate them -- if more than one type or function needs to be passed for the different cases, then they can be wrapped in a class, and only the class is passed. I would prefer this method to using an `enum` to differentiate the functions, because an `enum` requires `constexpr if` or other tricks, and we don't use C++17 yet.

The fetch, reduce, and finalize functor classes can be enclosed in a class, and then that class passed to choose `asum`, `nrm2`, etc.

library/src/blas1/rocblas_asum.hpp
daineAMD: Good point, if we want to keep the internal API clean then having extra parameters for different cases likely isn't ideal. A possible alternative if having these "internal API calls" call a generic function which can handle all three, this could improve code reuse and keep the internal API clean?

library/src/blas1/rocblas_nrm2.hpp
TorreZuk: I don't think Daine or I was suggesting merging the asum, and nrm2.   I don't think it saves much and keeping the functions separate is nice abstraction.    I think one could have merged a lot the framework into a C++ class and avoided all the duplicate patterns, but that might be another discussion.

TorreZuk:
Let me know if there is something important remaining to address as I think I should merge this PR and continue more template reuse in a new PR as some of this reduction is being reused and isn't changing further.

PR #698, opened 2019-09-13
Batched swap

TorreZuk:
This adds into blas1 the batched and strided_batched for swap.  Refactors non batched swap to follow new coding pattern (impl and template).   It defers the merging of the batched and strided_batched kernels until final pattern agreed upon but already shares the simple computation kernel code.

Example combined test times for all 3 cases on Vega20:
372 tests from 6 test cases ran. (31937 ms total)

### DIFF
--- a/clients/common/rocblas_gentest.py
+++ b/clients/common/rocblas_gentest.py
@@ -196,17 +196,26 @@ def setdefaults(test):
     # These are only for dynamic defaults
     # TODO: This should be ideally moved to YAML file, with eval'd expressions.
 
-    if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
-        test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
-                                    test['strideScale']))
+    if test['function']=='swap_strided_batched':    
+        if all([x in test for x in ('N', 'incx', 'stride_scale')]):
+            test.setdefault('stride_x', int(test['N'] * abs(test['incx']) *
+                                            test['stride_scale']))
+        if all([x in test for x in ('N', 'incy', 'stride_scale')]):
+            test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
+                                            test['stride_scale']))
     else:
-       test.setdefault('stride_x', 0)
-
-    if all([x in test for x in ('N', 'incy', 'strideScale')]) and test['function']=='ger_strided_batched':
-        test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
+        if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
+            test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
                                         test['strideScale']))
-    else:
-        test.setdefault('stride_y', 0)
+        else:
+            test.setdefault('stride_x', 0)
+
+        if all([x in test for x in ('N', 'incy', 'strideScale')]) and test['function']=='ger_strided_batched':
+            test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
+                                            test['strideScale']))
+        else:
+            test.setdefault('stride_y', 0)
+
 
     if test['transA'] == '*' or test['transB'] == '*':
         test.setdefault('lda', 0)
--- a/clients/common/rocblas_gentest.py
+++ b/clients/common/rocblas_gentest.py
@@ -196,7 +196,7 @@ def setdefaults(test):
     # These are only for dynamic defaults
     # TODO: This should be ideally moved to YAML file, with eval'd expressions.
 
-    if test['function']=='swap_strided_batched':    
+    if test['function'] in ('swap_strided_batched'):    
         if all([x in test for x in ('N', 'incx', 'stride_scale')]):
             test.setdefault('stride_x', int(test['N'] * abs(test['incx']) *
                                             test['stride_scale']))
--- a/clients/common/rocblas_gentest.py
+++ b/clients/common/rocblas_gentest.py
@@ -200,9 +200,13 @@ def setdefaults(test):
         if all([x in test for x in ('N', 'incx', 'stride_scale')]):
             test.setdefault('stride_x', int(test['N'] * abs(test['incx']) *
                                             test['stride_scale']))
+        else:
+            test.setdefault('stride_x', 0)
         if all([x in test for x in ('N', 'incy', 'stride_scale')]):
             test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
                                             test['stride_scale']))
+        else:
+            test.setdefault('stride_y', 0)
     else:
         if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
             test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
--- a/clients/include/rocblas_arguments.hpp
+++ b/clients/include/rocblas_arguments.hpp
@@ -62,8 +62,7 @@ struct Arguments
     rocblas_int stride_b; //  stride_b > transB == 'N' ? ldb * N : ldb * K
     rocblas_int stride_c; //  stride_c > ldc * N
     rocblas_int stride_d; //  stride_d > ldd * N
-
-    rocblas_int stride_x;
+    rocblas_int stride_x; 
     rocblas_int stride_y;
 
     rocblas_int norm_check;
--- a/clients/include/rocblas_arguments.hpp
+++ b/clients/include/rocblas_arguments.hpp
@@ -62,7 +62,7 @@ struct Arguments
     rocblas_int stride_b; //  stride_b > transB == 'N' ? ldb * N : ldb * K
     rocblas_int stride_c; //  stride_c > ldc * N
     rocblas_int stride_d; //  stride_d > ldd * N
-    rocblas_int stride_x; 
+    rocblas_int stride_x;
     rocblas_int stride_y;
     rocblas_int stride_scale;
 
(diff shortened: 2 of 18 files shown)


clients/include/rocblas_arguments.hpp
```diff
@@ -62,9 +62,9 @@ struct Arguments
     rocblas_int stride_b; //  stride_b > transB == 'N' ? ldb * N : ldb * K
     rocblas_int stride_c; //  stride_c > ldc * N
     rocblas_int stride_d; //  stride_d > ldd * N
-
     rocblas_int stride_x;
     rocblas_int stride_y;
+    rocblas_int stride_scale;
```
mahmoodw: We don't want to add this variable to the Arguments struct. It is just a intermediate variable only found in the yaml files.

clients/common/rocblas_gentest.py
```diff
@@ -196,17 +196,30 @@ def setdefaults(test):
     # These are only for dynamic defaults
     # TODO: This should be ideally moved to YAML file, with eval'd expressions.
 
-    if all([x in test for x in ('M', 'incx', 'strideScale')]) and test['function']=='ger_strided_batched':
-        test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
-                                    test['strideScale']))
+    if test['function']=='swap_strided_batched':
```
leekillough: Should be written as `if test['function'] in ('swap_strided_batched'):` so that more functions can be added later.

clients/include/rocblas_arguments.hpp
leekillough: > 
> 
> We don't want to add this variable to the Arguments struct. It is just a intermediate variable only found in the yaml files.

Correct, `Arguments` is already pretty bloated, and adding members to `Arguments` requires adding them to `rocblas_common.yaml` as well.

clients/include/rocblas_common.yaml
```diff
@@ -252,6 +252,7 @@ Arguments:
   - stride_d: rocblas_int
   - stride_x: rocblas_int
   - stride_y: rocblas_int
+  - stride_scale: rocblas_int
```
leekillough: Just like in `struct Arguments`, this should not be added.

library/src/blas1/rocblas_swap.cpp
```diff
@@ -63,23 +48,8 @@ namespace
...
-
-        if(incx < 0)
-            x -= ptrdiff_t(incx) * (n - 1);
-        if(incy < 0)
-            y -= ptrdiff_t(incy) * (n - 1);
-
-        hipLaunchKernelGGL(swap_kernel, grid, threads, 0, rocblas_stream, n, x, incx, y, incy);
-
-        return rocblas_status_success;
+        constexpr rocblas_int NB = 256;
```
mahmoodw: Maybe we should be consistent about where we are declaring this NB variable, scal send this as a template parameter to the *impl function

daineAMD: I was looking at trtri when I made the change for scal, that's how it does it there. Not sure if there's a standard place to do this outside of that.

library/src/blas1/rocblas_swap.hpp
```diff
@@ -0,0 +1,47 @@
...
+        rocblas_swap_vals(x + tid * incx, y + tid * incy);
+    }
+}
+
+template <rocblas_int NB, typename T>
+rocblas_status rocblas_swap_template(
+    rocblas_handle handle, rocblas_int n, T* x, rocblas_int incx, T* y, rocblas_int incy)
+{
+    // Quick return if possible.
+    if(n <= 0)
```
mahmoodw: There is some inconsistency across the board about n < 0 being an invalid size or success. I am guessing this is because of the Fortran implementation. We should discuss this .

library/src/blas1/rocblas_swap.cpp
leekillough: Instead of hard-coding `constexpr` values inside of functions or files, they can be added as template parameters. Note that if they are added, they should be added at the _beginning_, rather than the _end_, of the template parameter list, so that any _type_ template arguments can still be deduced automatically.

Originally some of these `NB` and similar blocksizes were defined as macros. In my first refactoring, I changed the `#define`s to `constexpr ... =  ...;`. Maybe we should go one step further, and make them template parameters.

TorreZuk: If NB will be different per data type I guess moving it makes sense, otherwise you are duplicating in for each data type...  not sure why NB is used for number of threads either  :-)

clients/include/rocblas_arguments.hpp
TorreZuk: okay I was logging it so I could understand the tests better.   I can remove it and rename it to strideScale as in other PR, but if we don't log it should have a different convention ?

library/src/blas1/rocblas_swap.hpp
leekillough: > There is some inconsistency across the board about n < 0 being an invalid size or success. I am guessing this is because of the Fortran implementation. We should discuss this .

This is consistent -- if `N <= 0`, then swap is a no-op. Fortran code:
```
      IF (N.LE.0) RETURN
```

`N <= 0` means the input vector is empty, so the result depends on the operation:
```
DOT: 0
ASUM: 0
AMAX: 0
SWAP: no-op
COPY: no-op
AXPY: no-op
etc.
```
Matrix dimensions are handled differently from vectors, and cannot have negative values.

BLAS originally consisted only of BLAS1. This way back in the era of Cray vector processors, which could perform nearly optimally on any problem as long as the level-1 vector operations were optimized.

BLAS2 was added later, then BLAS3, when vector processors were declining and superscalar RISC with multi-level caching required special techniques for BLAS2 and BLAS3 performance, rather than simply composing them in terms of BLAS1 vector operations.

BLAS1 was more permissive with sizes and increments, since there were only one-dimensional vectors, so a negative or zero size can simply be considered a "0-length vector". For matrices, it becomes harder to decide what to do when one of the dimensions is negative, so it is flagged as invalid.

library/src/blas1/rocblas_swap.cpp
leekillough: In the case of `TRTRI`, there was a need for the blocksizes to change, not just based on type, but based on the calling routine.

There's not much redundancy in:
```
template<rocblas_int NB, typename T>
rocblas_status func(T* a)
{
}

func<256>(a);
```
instead of:
```
template<typename T>
rocblas_status func(T* a)
{
   static constexpr rocblas_int NB = 256;
}

func(a);
```
The lines of code and compiler optimization efficiency is the same. If you need to change the blocksize, you can change it at the call, and it can be different for each call.

clients/include/rocblas_arguments.hpp
leekillough: We could reverse-compute it from the other arguments and log it, assuming it's deterministic.

library/src/blas1/rocblas_swap_strided_batched.cpp
```diff
@@ -0,0 +1,164 @@
...
+                        stridey,
+                        "batch",
+                        batch_count);
+
+        if(!x || !y)
+            return rocblas_status_invalid_pointer;
+
+        size_t abs_incx = incx >= 0 ? incx : -incx;
+        size_t abs_incy = incy >= 0 ? incy : -incy;
+        if(!incx || !incy || (stridex < n * abs_incx) || (stridey < n * abs_incy) || stridex < 0
```
mahmoodw: Not sure if we've discussed whether or not incx/incy == 0 is an invalid size or not. Does this follow the same logic as negative n? Note the non batched version and Fortran does not explicitly handle this case.

leekillough: > Not sure if we've discussed whether or not incx/incy == 0 is an invalid size or not. Does this follow the same logic as negative n? Note the non batched version and Fortran does not explicitly handle this case.

`n <= 0` causes swap to return immediately with no operation. It is not an error.

`incx == 0` and `incy == 0` are accepted by the original BLAS, but are useless for `SWAP` since they repeatedly swap the same first element of `X` or `Y`.

I say we should not try to detect and flag `incx == 0` or `incy == 0` in `SWAP`.

In other routines like `DOT`, a zero increment would multiply the first element of one vector by another.

It's a degenerate corner case but we don't need to detect and flag it.

clients/common/rocblas_gentest.py
TorreZuk: @daineAMD was the camel caps a convention to indicate an non-argument.   Can unify the naming for this meta argument.

daineAMD: @mahmoodw introduced strideScale I believe. I don't think there's a specific reason for it to be camel case, we can probably change it to stride_scale so it matches the typical naming conventions we use.

amcamd: Please go ahead as in daineAMD's comment and use snake_case rather than camelCase

TorreZuk: sure should add something to indicate it is not directly an argument, ??? meta_ prefix

library/src/blas1/rocblas_swap.cpp
TorreZuk: standardized location as _impl parameter

library/src/blas1/rocblas_swap_strided_batched.cpp
TorreZuk: removed all invalid parameter guards, should match '78 checking but 0 increment is non-deterministic and won't be unit tested

clients/common/rocblas_gentest.py
leekillough: > sure should add something to indicate it is not directly an argument, ??? meta_ prefix

Probably not, because there are dozens of YAML objects which are not arguments.

If you want to indicate it is a "pseudo-argument", you could prepend it with `pseudo_`. This would only apply to YAML objects which are used to create arguments, not to the dozens of existing YAML objects which are not arguments.

PR #710, opened 2019-09-20
Copy batch and strided_batch support

mahmoodw:
Added batched functionality to copy as per the conventions discussed

Test timings for all 3 versions:
quick: 10.09 s
pre_checkin: 41.463 s
nightly: 125.9 s


library/src/blas1/rocblas_copy.hpp
```diff
@@ -0,0 +1,75 @@
...
+                                     rocblas_int    n,
+                                     const U        x,
+                                     rocblas_int    offsetx,
+                                     rocblas_int    incx,
+                                     rocblas_int    stridex,
+                                     V              y,
+                                     rocblas_int    offsety,
+                                     rocblas_int    incy,
+                                     rocblas_int    stridey,
+                                     rocblas_stride batch_count)
```
TorreZuk: We are only using rocblas_stride for stridex and stridey not batch_count so all these need to be flipped.

mahmoodw: oops, good catch!

clients/include/testing_copy_strided_batched.hpp
```diff
@@ -0,0 +1,186 @@
...
+    rocblas_int          stride_y    = arg.stride_y;
+    rocblas_int          batch_count = arg.batch_count;
+    rocblas_local_handle handle;
+    rocblas_int          abs_incx = incx >= 0 ? incx : -incx;
+    rocblas_int          abs_incy = incy >= 0 ? incy : -incy;
+    size_t               size_x   = N * size_t(abs_incx);
+    size_t               size_y   = N * size_t(abs_incy);
+
+    // argument sanity check before allocating invalid memory
+    if(N <= 0 || batch_count <= 0)
```
leekillough: `batch_count == 0` should probably follow normal path -- we want to make sure destination is not copied over

library/src/blas1/rocblas_copy.cpp
```diff
@@ -62,24 +52,10 @@ namespace
...
-        // Quick return if possible.
-        if(n <= 0)
-            return rocblas_status_success;
-
-        int  blocks = (n - 1) / NB + 1;
-        dim3 grid(blocks);
-        dim3 threads(NB);
-
-        hipStream_t rocblas_stream = handle->rocblas_stream;
+        if(!incx || !incy)
```
leekillough: `incx ==0 || incy==0` should be valid but not tested for -- see email thread and: http://www.netlib.org/lapack/lapack-3.1.1/html/dcopy.f.html

library/src/blas1/rocblas_copy.hpp
```diff
@@ -0,0 +1,75 @@
...
+    // bound
+    if(tid < n)
+    {
+        const T* x = load_ptr_batch(xa, hipBlockIdx_y, offsetx, stridex);
+        T*       y = load_ptr_batch(ya, hipBlockIdx_y, offsety, stridey);
+
+        if(incx < 0)
+            x -= ptrdiff_t(incx) * (n - 1);
+        if(incy < 0)
+            y -= ptrdiff_t(incy) * (n - 1);
```
leekillough: The adjustments based on `incx`, `incy` should be moved outside of the kernel, to where they were originally (`_impl`). Not sure how this interacts with `load_ptr_batch`, but the way it's written right now, will be much slower.

library/src/blas1/rocblas_copy.cpp
mahmoodw: must've slipped by, will correct that.

library/src/blas1/rocblas_copy_batched.cpp
```diff
@@ -0,0 +1,149 @@
...
+                        "batch_count",
+                        batch_count);
+
+        if(!x || !y)
+            return rocblas_status_invalid_pointer;
+
+        RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);
+
+        if(!incx || !incy || batch_count < 0)
+            return rocblas_status_invalid_size;
```
leekillough: `incx==0 || incy ==0` should be okay. http://www.netlib.org/lapack/lapack-3.1.1/html/dcopy.f.html

library/src/blas1/rocblas_copy_strided_batched.cpp
```diff
@@ -0,0 +1,187 @@
...
+                        "batch_count",
+                        batch_count);
+
+        if(!x || !y)
+            return rocblas_status_invalid_pointer;
+
+        RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);
+
+        if(!incx || !incy || batch_count < 0)
+            return rocblas_status_invalid_size;
```
leekillough: `incx==0 || incy==0` should be okay. http://www.netlib.org/lapack/lapack-3.1.1/html/dcopy.f.html

library/src/blas1/rocblas_copy.hpp
mahmoodw: This doesn't really seem possible with the way we've chosen to combine the non-batched/batched/strided_batched into a single template. Although, I agree that the non-batched version will be slower. How do you suggest we tackle this?

TorreZuk: I would think going back to the original kernel is valid option for the base case.   Reusing generic kernels isn't a great improvement when there is so little code that is duplicated anyway and making it much more complicated with unused parameters outweighs the potential maintenance improvement in my opinion.

leekillough: > This doesn't really seem possible with the way we've chosen to combine the non-batched/batched/strided_batched into a single template. Although, I agree that the non-batched version will be slower. How do you suggest we tackle this?

I thought we were only merging the batched kernels, and that the non-batched kernels would stay separate.

There should be a way to pass in the "adjusted" x, y, etc. rather than calculating them **_at every single element._**

This should be a separate JIRA task I'm willing to take on -- to refactor all of the rocBLAS kernels' templates so that the overhead is avoided.

It can be done with template specializations (or, in C++17, with `constexpr if`).

It might be as simple as adding a `bool` template parameter which specifies whether to "adjust" the `x` and `y` pointers. If `true`, it's done the way it's done now. If `false`, it would be done once before the GPU kernel.

mahmoodw: The few functions that have been combined to a single template are combining all three, this seems to be the first of which that handle the negative increments. GER and GEMV still need to be refactored to a single template.

If you would like to take on this task, how do you suggest we proceed with this PR? Should I revert back to the original kernel for now on the non batched case?

From what I can tell only the batched version needs to be adjusted at the kernel level.

TorreZuk: You can look at swap which supports negative increments.  I reuse an inline kernel for calculation logic but the ptr_loads are only in the batched kernels

mahmoodw: Swap has not combined any templates. Should I revert back to this style for copy and dot?

TorreZuk: For swap all 3 variants call the same inline template inside their kernels.  It is reuse without wasted parameter overhead.   If you want it to run with the same speed as the original non-batched and also share the code that is an option.

PR #711, opened 2019-09-23
Adding Workaround flag for fortran bug

mahmoodw:
resolves SWDEV-203325

Added in a workaround flag for a known bug for Gfortran versions 7,8,9. Details at the link below:
https://developer.r-project.org/Blog/public/2019/05/15/gfortran-issues-with-lapack/index.html

Would like to see if there is any timing impact on CI with this flag enabled.

### DIFF
--- a/deps/external-lapack.cmake
+++ b/deps/external-lapack.cmake
@@ -37,6 +37,10 @@ ExternalProject_Add(
   INSTALL_COMMAND ""
   LOG_INSTALL 1
 )
+# The fortran flag '-fno-optimize-sibling-calls' has been added as a workaround for a known bug
+# that causes incompatibility issues between gfortran and C lapack calls for gfortran versions 7,8 and 9
+# The ticket can be tracked at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90329 
+
 
 ExternalProject_Get_Property( lapack source_dir )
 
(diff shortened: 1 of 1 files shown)


deps/external-lapack.cmake
```diff
@@ -32,7 +32,7 @@ ExternalProject_Add(
   PREFIX ${CMAKE_BINARY_DIR}/lapack
   GIT_REPOSITORY ${lapack_git_repository}
   GIT_TAG ${lapack_git_tag}
-  CMAKE_ARGS ${lapack_cmake_args} -DCBLAS=ON -DLAPACKE=OFF -DBUILD_TESTING=OFF
+  CMAKE_ARGS ${lapack_cmake_args} -DCBLAS=ON -DLAPACKE=OFF -DBUILD_TESTING=OFF -DCMAKE_Fortran_FLAGS='-fno-optimize-sibling-calls'
```
TorreZuk: I would add a comment line saying why this flag is being added.     You can refer to the external PR https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90329 as perhaps this can be removed in the near future.

TorreZuk: Now that I look at this again not sure if you are getting the indended FLAGS += result.  If you see huge differences in speed this might be something to verify.

leekillough:
> Would like to see if there is any timing impact on CI with this flag enabled.

Should be insignificant impact; it just changes a CPU calling sequence, and does not affect critical path.

deps/external-lapack.cmake
mahmoodw: Results look fine to me. Merging the fix in.

PR #712, opened 2019-09-23
trtri_batched and trtri_strided_batched

daineAMD:
Summary of proposed changes:
-  Renames trtri_batched into trtri_strided_batched
-  Implements trtri_batched
-  File structure changed to be the same as other functions which support batched/strided_batched

Testing for trtri is somewhat limited as it's mostly internal. Testing for trsm has coverage for trtri. Approximate times for trtri tests are as follows:

quick: 6 seconds
pre-checkin: 11 seconds
nightly: none


clients/include/testing_trtri_batched.hpp
```diff
@@ -132,8 +148,11 @@ void testing_trtri_batched(const Arguments& arg)
     }
```
mahmoodw: You want to multiply with batch_count to get the correct gflops

clients/include/testing_trtri_batched.hpp
```diff
@@ -158,31 +177,29 @@ void testing_trtri_batched(const Arguments& arg)
         }
```
mahmoodw: Again, you want to multiply with batch_count to get the correct gflops

clients/include/testing_trtri_strided_batched.hpp
```diff
@@ -0,0 +1,216 @@
...
+        handle, uplo, diag, N, dA, lda, bsa, dinvA, lda, bsa, batch_count));
+
+    // Test in place
+    CHECK_ROCBLAS_ERROR(rocblas_trtri_strided_batched<T>(
+        handle, uplo, diag, N, dA, lda, bsa, dA, lda, bsa, batch_count));
+
+    if(arg.timing)
+    {
+        gpu_time_used  = get_time_us() - gpu_time_used;
+        rocblas_gflops = trtri_gflop_count<T>(N) / gpu_time_used * 1e6;
```
mahmoodw: Same as batched - multiply with batch_count

clients/include/testing_trtri_strided_batched.hpp
```diff
@@ -0,0 +1,216 @@
...
+        for(size_t i = 0; i < batch_count; i++)
+        {
+            rocblas_int info = cblas_trtri<T>(char_uplo, char_diag, N, hB + i * bsa, lda);
+            if(info != 0)
+                printf("error in cblas_trtri\n");
+        }
+        if(arg.timing)
+        {
+            cpu_time_used = get_time_us() - cpu_time_used;
+            cblas_gflops  = trtri_gflop_count<T>(N) / cpu_time_used * 1e6;
```
mahmoodw: Same as batched - multiply with batch_count

library/include/rocblas-functions.h
```diff
@@ -2088,7 +2088,68 @@ ROCBLAS_EXPORT rocblas_status rocblas_dtrtri(rocblas_handle   handle,
 /*! \brief BLAS Level 3 API
 
     \details
-    trtri  compute the inverse of a matrix  A
+    trtri  compute the inverse of a matrices  A
```
mahmoodw: trtri_batched*

clients/include/testing_trtri_strided_batched.hpp
```diff
@@ -0,0 +1,216 @@
...
+#include "utility.hpp"
+
+template <typename T>
+void testing_trtri_strided_batched(const Arguments& arg)
+{
+    rocblas_int N           = arg.N;
+    rocblas_int lda         = arg.lda;
+    rocblas_int batch_count = arg.batch_count;
+
+    rocblas_int bsa    = lda * N;
```
mahmoodw: May want to use stride_a argument

clients/gtest/trtri_gtest.cpp
```diff
@@ -47,7 +50,8 @@ namespace
     enum trtri_kind
     {
         trtri_k,
-        trtri_batched_k
+        trtri_batched_k,
+        trtri_strided_batched_k
```
leekillough: Use trailing comma on the last entry so that all entries are uniform and can be expanded later without changing the last line.

clients/include/testing_trtri_batched.hpp
```diff
@@ -22,8 +22,7 @@ void testing_trtri_batched(const Arguments& arg)
     rocblas_int lda         = arg.lda;
     rocblas_int batch_count = arg.batch_count;
 
-    rocblas_int bsa    = lda * N;
-    size_t      size_A = size_t(bsa) * batch_count;
+    size_t size_A = size_t(lda * N);
```
leekillough: Shouldn't that be `size_t size_A = size_t(lda) * N;` ?

clients/include/testing_trtri_batched.hpp
```diff
@@ -36,33 +35,42 @@ void testing_trtri_batched(const Arguments& arg)
 
     // argument sanity check, quick return if input parameters are invalid before allocating invalid
     // memory
-    if(N < 0 || lda < 0 || lda < N || batch_count < 0)
+    if(N < 0 || lda < 0 || lda < N || batch_count <= 0)
```
leekillough: `batch_count == 0` is a valid size and not an error, just a no-op.

clients/include/testing_trtri_strided_batched.hpp
leekillough: > May want to use stride_a argument, and use rocblas_stride variable type.

If you do so, you'll need to define the `rocblas_stride` type in `rocblas_common.yaml` as a "YAML typedef" name.

clients/include/testing_trtri_strided_batched.hpp
```diff
@@ -0,0 +1,216 @@
...
+    // char_uplo = 'U';
+    rocblas_fill     uplo = char2rocblas_fill(char_uplo);
+    rocblas_diagonal diag = char2rocblas_diagonal(char_diag);
+
+    rocblas_local_handle handle;
+
+    // argument sanity check, quick return if input parameters are invalid before allocating invalid
+    // memory
+    if(N < 0 || lda < 0 || lda < N || batch_count <= 0)
+    {
```
leekillough: Not sure why `batch_count == 0` is handled here.

The main code should work with `batch_count == 0`, maybe with guards to make sure that a null pointer allocated with `batch_count` is only really null if `batch_count != 0`.

library/include/rocblas-functions.h
leekillough: `a matrices A` -> `matrices A[???]`

library/src/blas3/trtri_trsm.hpp
```diff
@@ -258,7 +265,7 @@ rocblas_status rocblas_trtri_trsm_template(rocblas_handle   handle,
         size_t blockSize            = 128;
         size_t tri_elements_to_zero = num_non_tri_elements(rem);
         size_t numBlocks            = (tri_elements_to_zero + blockSize - 1) / blockSize;
-        hipLaunchKernelGGL(rocblas_trtri_batched_fill,
+        hipLaunchKernelGGL(rocblas_trtri_fill<T>,
```
leekillough: Template argument `<T>` is probably unnecessary, especially if it can be deduced from the actual arguments.

library/src/include/utility.h
```diff
@@ -59,6 +59,33 @@ __forceinline__ __device__ __host__ T load_scalar(const T* xp)
...
+}
+
+// For device array of device pointers
+template <typename T>
+__forceinline__ __device__ __host__ T*
+                                    load_ptr_batch(T** p, rocblas_int block, rocblas_int offset, rocblas_stride stride)
+{
+    return p[block] + offset;
+}
+
```
leekillough: Why was this moved above `load_scalar`'s overload for `rocblas_half2`?

clients/include/testing_trtri_strided_batched.hpp
daineAMD: Got rid of the batch_count == 0 check here, the rest of the code already seems to work with batch_count == 0. However it doesn't really do anything interesting as none of the copies do anything (meaning the output of rocblas_trtri isn't checked) and cblas doesn't run with the current code.

library/include/rocblas-functions.h
daineAMD: Improved documentation in 41463e5.

library/src/include/utility.h
daineAMD: Must have happened in a merge, rearranged stuff in 41463e5.

library/src/blas3/trtri_trsm.hpp
daineAMD: Argument `A` will either be a T* or a T**, is there a simple way to retrieve type `T` from this, as I need the type to cast a 0 into `rocblas_trti_fill_lower` from this function?

YvanMokwinski: using traits?..

clients/include/testing_trtri_strided_batched.hpp
mahmoodw: Removing from here doesn't really help make any additional checks. Do we want to make it a special case and verify that the destination is not being written to? The same goes for the other BLAS1 PRs, otherwise all are just checking for rocblas_status_success as is.

clients/include/testing_trtri_batched.hpp
leekillough: We need `batch_count == 0` to work correctly. Usually this simply means returning `rocblas_status_success`, or `rocblas_status_size_unchanged` if `handle->is_device_memory_size_query()` is `true`.

I don't understand the code comments or your comments here ("we are checking for `rocblas_status_success` here when `batch_count ==0`"). You should not be _checking_ for `rocblas_status_success` when `batch_count == 0` -- you should be _returning_ `rocblas_status_success` (or `rocblas_status_size_unchanged`).

```
if(!batch_count)
    return handle->is_device_memory_size_query() ? rocblas_status_size_unchanged : rocblas_status_success;
```
Memory allocation and memory size queries are done in very specific ways. I can help.

It is not generally a good idea to ignore argument checking and go ahead and compute memory size based on possibly invalid arguments. Device memory size queries are allowed to return invalid argument errors.

The TRTRI _kernel_ code is not allowed to allocate device memory if it is called from another routine line TRSM. Its callers, like TRSM, must provide it buffer space. But a self-contained TRTRI which has a C wrapper, can allocate device memory. TRTRI is not a part of BLAS, so I don't know why we are extending a batch versioned for it. Only batched code to support batched TRSM is necessary. It's unfortunate that TRTRI was exposed in the public C interface, because there's no TRTRI in BLAS.

It took weeks to get allocation working correctly for TRTRI. It is very delicate code. As I recall, the batched code already existed when the device memory allocation code was added, but there were no C APIs for the batched versions.

I will need to look at this PR offline (not on GitHub) to properly review it. TRSM and TRTRI are the most complicated routines in BLAS.

daineAMD: We are returning rocblas_status_success from trtri when batch_count == 0, in the testing code here we are _checking_ that that is the case. We aren't carrying out _all_ of the testing code when batch_count == 0 because this would complicate the code as the setup would have to have multiple checks that batch_count != 0 before allocating memory anyway, thus we are caling trtri early here and ensuring that it still returns rocblas_status_success. If we want to run through all of the testing code when batch_count == 0 we can do this, it might just add unnecessary complications to the test code and I'm unsure that it will be useful.

You mentioned that trtri kernel code is not allowed to allocate device memory. I am currently only allocating memory in the _impl functions, not in the _template as was recommended in the guidelines. So, when trsm calls trtri, for example, it will have to allocate memory before calling the _template. I'm not sure if you're referring to something else in another file that I'm missing.

The code for trtri_strided_batched had already existed, but was named trtri_batched, I changed the name of this to match the rest of the library. It was already present in the C api when I began this; I added trtri_strided_batched to the C api to be consistent.

Thanks for taking the time to review it offline, let me know if you have any further concerns that I can address.

leekillough: > We are returning rocblas_status_success from trtri when batch_count == 0, in the testing code here we are checking that that is the case. We aren't carrying out all of the testing code when batch_count == 0 because this would complicate the code as the setup would have to have multiple checks that batch_count != 0 before allocating memory anyway, thus we are caling trtri early here and ensuring that it still returns rocblas_status_success. If we want to run through all of the testing code when batch_count == 0 we can do this, it might just add unnecessary complications to the test code and I'm unsure that it will be useful.

Okay, if the testing code cannot currently handle `batch_count == 0` without a lot of guards, then maybe it's okay to do it in the same code as `batch_count < 0` or `n < 0` (with the "safe size" allocation of arrays), just with it expecting `rocblas_status_success` instead of `rocblas_status_invalid_size`.

> You mentioned that trtri kernel code is not allowed to allocate device memory. I am currently only allocating memory in the _impl functions, not in the _template as was recommended in the guidelines. So, when trsm calls trtri, for example, it will have to allocate memory before calling the _template. I'm not sure if you're referring to something else in another file that I'm missing.

That is the correct interpretation.

> The code for trtri_strided_batched had already existed, but was named trtri_batched, I  changed the name of this to match the rest of the library. It was already present in the C api when I began this; I added trtri_strided_batched to the C api to be consistent.

That is fine, to be consistent. I don't know how many users are actually calling the `TRTRI` C API. `TRTRI` is mainly important for `TRSM`, since it inverts the diagonal blocks.

leekillough:
Does [strided] batched TRSM call TRTRI, and if so, does it pass allocated device memory to TRTRI?

One of the comments says that the C API for TRTRI is called by TSRM.

One of the difficulties with TRSM, was that device memory had to be allocated for all functions TRSM calls, including TRTRI.

The TRTRI C API calls an `_impl` function which allocates device memory and then calls a `_template` function, passing it the allocated memory. TRSM allocates memory for itself and TRTRI, and passes the allocated memory to the TRTRI `_template` function. If TRSM calls the C API for TRTRI, then TRTRI will allocate and free device memory during the call, something which cannot be done if TRSM already allocated device memory -- TRSM must pass the allocated memory to TRTRI.

If device memory is already allocated by TRSM, and it calls TRTRI, and then TRTRI attempts to allocate device memory again, it will fail.

daineAMD:
> Does [strided] batched TRSM call TRTRI, and if so, does it pass allocated device memory to TRTRI?
> 
> One of the comments says that the C API for TRTRI is called by TSRM.
> 
> One of the difficulties with TRSM, was that device memory had to be allocated for all functions TRSM calls, including TRTRI.
> 
> The TRTRI C API calls an `_impl` function which allocates device memory and then calls a `_template` function, passing it the allocated memory. TRSM allocates memory for itself and TRTRI, and passes the allocated memory to the TRTRI `_template` function. If TRSM calls the C API for TRTRI, then TRTRI will allocate and free device memory during the call, something which cannot be done if TRSM already allocated device memory -- TRSM must pass the allocated memory to TRTRI.
> 
> If device memory is already allocated by TRSM, and it calls TRTRI, and then TRTRI attempts to allocate device memory again, it will fail.

No, that must be a typo. In my trsm implementation, the C API is not called internally, only the template level of trtri is called from trsm.

library/src/blas3/rocblas_trtri_strided_batched.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+}
+
+/* ============================================================================================ */
+
+/*
+ * ===========================================================================
+ *    C interface
+ *    This function is called by trsm
```
leekillough: ` *    This function is called by trsm`

Typo?

daineAMD: Looks like this was copied from the original trtri_batched.cpp file, will remove.

PR #721, opened 2019-09-25
Build without tensile

TorreZuk:
Summary of proposed changes:
Restores the development use only non-tensile build of rocblas via new install.sh flag -n [-no_tensile] for use with -c [--clients] flag.
This will allows rocblas builds of blas1 and blas2 (less trsv) and rocblas-test but is only intended for quick development rebuilds.


install.sh
```diff
@@ -24,6 +24,7 @@ function display_help()
 #  echo "    [--cuda] build library for cuda backend"
   echo "    [--cpu_ref_lib] specify libary to use for cpu reference code in testing (blis or lapack)"
   echo "    [--hip-clang] build library for amdgpu backend using hip-clang"
+  echo "    [-n|--no_tensile] build subset of library that doesn't require tensile (testing)"
```
leekillough: The `(testing)` remark at the end should probably be rewritten as `(beta)`, `(under development)` or `(experimental)`, unless you really are referring to Tensile or other "testing".

leekillough: Oh, and `Tensile` is usually capitalized.

TorreZuk: I had already merged, I'll adjust these later in another PR if okay

PR #720, opened 2019-09-25
IAMAX / IAMIN (batched/strided batched)

YvanMokwinski:
BLAS level 1 iamax/iamin batched and strided batched versions.

PR #722, opened 2019-09-26
Change BLAS functions to return immediately without error when possible

leekillough:
Partially resolves [SWDEV-203994](http://ontrack-internal.amd.com/browse/SWDEV-203994), by exiting early with `rocblas_status_success` when `n==0` or similar, without checking other arguments for `nullptr` or other invalid values. "Short circuit on success".

This is the Fortran philosophy: If a mathematical result can be determined based on some arguments, then other arguments do not matter, and _do not even need to be checked for validity_. For example, if `m==0 || n==0`, then GEMM is a no-op, and `alpha` and `beta` can be `nullptr` without error, and `lda`, `ldb`, `ldc` do not need to be positive.

This is also the CuBLAS behavior.

When `alpha == nullptr` or `beta == nullptr`, then `NaN` is printed in logging .

Still need to go through all of BLAS1. Held off because of pending PRs.

Still need to add tests. Right now the YAML has no way of entering `nullptr` values. Alternatively, the `*_bad_args` functions can be modified to add test cases where some of the arguments are invalid, but others should cause it to short-circuit success. This is probably the easier solution.

Also, minor refactoring: Rather than using an unused function parameter for `rocblas_reduction_kernel_workspace_size` to deduce the `To` type, pass `To` as an explicit template argument. Also, make `batch_count` an optional argument with a default value of `1`.


library/src/blas3/Tensile/gemm_strided_batched.cpp
```diff
@@ -98,13 +98,13 @@ namespace
...
                         alphass << " --alphai " << std::imag(*alpha);
 
                     std::stringstream betass;
-                    betass << "--beta " << std::real(*beta);
-                    if (std::imag(*beta) != 0)
+                    betass << "--beta " << (beta ? std::real(*beta) : std::numeric_limits<T>::quiet_NaN());
+                    if (beta && std::imag(*beta) != 0)
                         betass << " --betai " << std::imag(*beta);
 
                     log_bench(handle,
```
TorreZuk: Also wondered if we could make some argument class for blas1, blas2, blas3 to indicate what standard arguments they take.   Then a helper class with static methods could maybe do all the logging, layer, … work for all functions.     Don't know if we care about minor efficiency changes for all these methods but it would keep things consistent and reduce code.

YvanMokwinski:
"Also, minor refactoring: Rather than using an unused function parameter for rocblas_reduction_kernel_workspace_size to deduce the To type, pass To as an explicit template argument." 

This is something that have already been done in the PR #720

library/src/blas3/Tensile/gemm.cpp
```diff
@@ -70,25 +69,25 @@ namespace
                             m,
                             n,
                             k,
-                            *alpha,
+                            alpha ? *alpha : std::numeric_limits<T>::quiet_NaN(),
```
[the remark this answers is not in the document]
leekillough: We could add a helper function.

The above logging function call might not even be correct; I'll have to make sure.

library/src/blas1/reduction_strided_batched.h
```diff
@@ -286,8 +285,8 @@ __global__ void
     batch_count rocblas_int
               number of instances in the batch
     @param[out]
-    workspace To*  
-              temporary GPU buffer for inidividual block results for each batch
+    workspace To*
+              temporary GPU buffer for individual block results for each batch
               and results buffer in case result pointer is to host memory
               Size must be (blocks+1)*batch_count*sizeof(To)
```
YvanMokwinski: Is 'blocks' defined ?

library/src/blas3/Tensile/gemm_strided_batched.cpp
leekillough: Every routine is separate, and there is not a "standard argument list" for BLAS1, BLAS2, or BLAS3. With very few exceptions in BLAS1, the argument list is different for every routine.

If we merged several routines' argument lists into a universal struct like `Arguments`, and printed that, then we would be printing a lot of extraneous arguments.

That's why the logging system is flexible, allowing arbitrary argument lists to be specified for each routine. The logging functions accept an arbitrary number of arguments.

It took around 2 months to implement the current logging system. With other priorities, we don't have the time to rewrite the logging system.

Related to that, the current `Arguments` struct is _universal_ for all of rocBLAS, and splitting it into separate classes for BLAS1, BLAS2, BLAS3, etc. would require **extensive** modifications across multiple languages (C++, Python, YAML).

leekillough:
> This is something that has already been done in the PR #720

Then it will need to be merged in one PR or another, along with the change to make `batch_count = 1` default argument.

PR #724, opened 2019-09-27
handle negative inc via pre-kernel shift

TorreZuk:
Wasiq pointed out the negative increment handling offset is constant across batches so pulled it out of the kernel for both batched variants via the offset/shift argument we already added for rocSOLVER.    So do we expect the negative shift being large and so should make offsets rocblas_stride or assume the += N*inc is still fine as int32.   I would lean toward making it a rocblas_stride to allow this optimization.

### DIFF
--- a/library/src/blas1/rocblas_swap_batched.hpp
+++ b/library/src/blas1/rocblas_swap_batched.hpp
@@ -9,10 +9,10 @@
 template <typename T>
 __global__ void rocblas_swap_kernel_batched(rocblas_int n,
                                             T*          x[],
-                                            rocblas_int shiftx,
+                                            ptrdiff_t   shiftx,
                                             rocblas_int incx,
                                             T*          y[],
-                                            rocblas_int shifty,
+                                            ptrdiff_t   shifty,
                                             rocblas_int incy)
 {
     ssize_t tid = blockIdx.x * blockDim.x + threadIdx.x; // only dim1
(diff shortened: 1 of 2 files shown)


leekillough:
In all of the increment calculations, we use `ptrdiff_t(inc) * (n-1)`, so we convert it to 64 bits before doing the multiplication.

TorreZuk:
Yes that is why if everyone agrees we could promote shift parameters to an rocblas_int64, or
I can just make an int64 combined result and pass this off to the kernel.   (not in current PR, I can push in a minute)
This will make the strided_batched and batched templates template reuse even simpler.
Wondering what people think is best.

leekillough:
We can change _internal_ interfaces at will.

Only the public C APIs need to be handled carefully.

If the multiplication of a 32-bit value can overflow 32 bits, while remaining valid in 64 bits, then the 32-bit value should be converted to 64-bit first.

`ptrdiff_t(inc)` is how we handle it now, so that it gets converted to whatever integer type is the same size as pointers. We don't need to use `int64_t` or `rocblas_int64`, since those are always assuming 64 bits. `ptrdiff_t` (signed) and `size_t` (unsigned) are the right types to use.

YvanMokwinski:
> Yes that is why if everyone agrees we could promote shift parameters to an rocblas_int64, or
> I can just make an int64 combined result and pass this off to the kernel. (not in current PR, I can push in a minute)
> This will make the strided_batched and batched templates template reuse even simpler.
> Wondering what people think is best.

We could stay consistent with the convention and introduce a rocblas_shift.

TorreZuk:
Okay update the PR with what it sounds like is acceptable

TorreZuk:
Build cleanly earlier, current failures all seem to be unrelated python issues so merging.

PR #728, opened 2019-10-01
Master to develop

saadrahim:
@amcamd I don't know the best practices for rocBLAS, but we need to get the fixes in master into develop. This is required to build clients on rocm 2.8.


saadrahim:
@amdcamd I will push it to develop as suggested by the command line instructions in this PR as the master branch is protected. Please let me know if you have any objections to this.

PR #731, opened 2019-10-01
Removed centos jobs

amdkila:
- CentOS jobs are currently failing due to a mismatch in intrinsic sse headers (i.e. xmmintrin.h), so this PR disables them from CI
- Added Vega 20 Ubuntu job as well


saadrahim (approved):
Temporary work around until centos tests can be fixed.

PR #735, opened 2019-10-03
Refactor Ger and Gemv

mahmoodw:
resolves SWDEV-202362

Refactor focused around Ger and Gemv to:
- Unify template function across batched and non batched routines
- Add support for features like offset and vectors of scalars to the template
- Remove stride checks and appropriate tests
- Unify Gemv yamls
- Fix Ger bad_arg calls
- Use rocblas_stride for stride

Also fixed scal bandwidth calculation for batched routines


library/include/rocblas-functions.h
```diff
@@ -190,12 +190,14 @@ ROCBLAS_EXPORT rocblas_status rocblas_zdscal_batched(rocblas_handle
...
     @param[in]
-    stride_x     stride form the start of one vector (x_i) and the next one (x_i+1).
-                 There are no restrictions placed on stride_x, however the user should
-                 take care to ensure that stride_x is of appropriate size, for a typical
-                 case this means stride_x > n * incx.
+    stride_x    rocblas_stride
+                stride form the start of one vector (x_i) and the next one (x_i+1).
+                There are no restrictions placed on stride_x, however the user should
+                take care to ensure that stride_x is of appropriate size, for a typical
+                case this means stride_x > n * incx.
```
TorreZuk: >=   ???

library/src/blas2/gemv_device.hpp
```diff
@@ -383,199 +383,96 @@ __device__ void gemvt_kernel(rocblas_int m,
...
-                                     U              alpha_device_host,
-                                     const T* const Aa[],
-                                     rocblas_int    lda,
-                                     const T* const xa[],
-                                     rocblas_int    incx,
-                                     U              beta_device_host,
-                                     T* const       ya[],
-                                     rocblas_int    incy)
+template <rocblas_int NB_X, typename T, typename U, typename V, typename W>
+__global__ void gemvt_kernel_test(rocblas_int    m,
```
TorreZuk: is this a test ?   _test

mahmoodw: Need to rename that

clients/common/rocblas_gentest.py
```diff
@@ -214,6 +214,22 @@ def setdefaults(test):
...
+                test.setdefault('stride_y', int(test['M'] * abs(test['incy']) *
+                                                test['stride_scale']))
+        else:
+            if all([x in test for x in ('M', 'incx', 'stride_scale')]):
+                test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
+                                                test['stride_scale']))
+            if all([x in test for x in ('N', 'incy', 'stride_scale')]):
+                test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
+                                                test['stride_scale']))
+
```
leekillough: We should try to avoid duplication as much as possible. Is there a way to combine `stride_x` and `stride_y` calculations for different functions?

library/src/include/utility.h
```diff
@@ -103,6 +103,22 @@ __forceinline__ __device__ __host__ T*
     return p[block] + offset;
 }
 
+// For device pointers
```
daineAMD: The above functions can be removed right, since these are replacing them? Duplication probably happened during a merge.

library/src/blas2/rocblas_gemv.hpp
```diff
@@ -7,214 +7,40 @@
...
+                                     rocblas_stride    stridex,
+                                     const U*          beta,
+                                     W*                y,
+                                     rocblas_int       offsety,
+                                     rocblas_int       incy,
+                                     rocblas_stride    stridey,
+                                     rocblas_int       batch_count)
 {
     //quick return
     if(!m || !n)
```
daineAMD: Should this be `if(!m || !n || !batch_count)` ?

library/src/blas2/rocblas_ger.cpp
```diff
@@ -88,8 +88,8 @@ namespace
         if(!m || !n)
```
daineAMD: I think quick return should be at the _template level, not at _impl level.

library/src/blas2/rocblas_ger_batched.cpp
```diff
@@ -115,8 +115,8 @@ namespace
         if(!m || !n || !batch_count)
```
daineAMD: Same here, should be at _template.

library/src/blas2/rocblas_ger.cpp
```diff
@@ -1,10 +1,10 @@
 /* ************************************************************************
  * Copyright 2016-2019 Advanced Micro Devices, Inc.
  * ************************************************************************ */
+#include "rocblas_ger.hpp"
```
daineAMD: I can't seem to find this file, am I somehow missing it or is it not added?

library/src/include/utility.h
mahmoodw: good catch

library/src/blas2/rocblas_ger.cpp
mahmoodw: Looks like it I forgot to add it, doing that now

library/src/blas2/rocblas_gemv.hpp
mahmoodw: good catch

PR #737, opened 2019-10-04
Rot(m)(g) batched and strided_batched

daineAMD:
Adds the following functions:
- rot_batched, rot_strided_batched
- rotm_batched, rotm_strided_batched
- rotg_batched, rotg_strided_batched
- rotmg_batched, rotmg_strided_batched

Edit:
<s>Does not support batched `param` in rotm and rotmg from API. Supports batched `param` at _template level. Unsure if we want to support batched `param` on the host as this forces a hipMemcpy in the _template function.</s>

DOES support batched `param` in rotm and rotmg from API. Does NOT support `param` located on the host for (strided_)batched versions of rotm.

**Testing**
The testing times for all rot functions, along with Andrew's estimations for 4 BLAS1 functions testing time follows:

| Functions     | Quick  | Pre-checkin  | Nightly       |
|------|------|----------|--------------|
| All rot funcs  | 5.1s    | 68.7s             |  347s          | 
| 4 L1 Funcs | 8 - 48s   | 48 - 144s      | 288s - 864s |


clients/include/testing_rotm_batched.hpp
```diff
@@ -0,0 +1,274 @@
...
+    device_batch_vector<T> by(batch_count, size_y);
+
+    for(int b = 0; b < batch_count; b++)
+    {
+        hx[b] = host_vector<T>(size_x);
+        hy[b] = host_vector<T>(size_y);
+    }
+
+    int last = batch_count - 1;
+    if((!bx[last] && size_x) || (!by[last] && size_y))
```
TorreZuk: Probably need to make batch vector fail if any allocation fail to avoid such things as this last test.

library/src/blas1/rocblas_rot.hpp
```diff
@@ -0,0 +1,143 @@
...
+                           offset_y,
+                           incy,
+                           stride_y,
+                           *c,
+                           c_stride,
+                           *s,
+                           s_stride);
+
+    return rocblas_status_success;
+}
```
TorreZuk: Not sure if we want newlines at EOF

library/src/blas1/rocblas_rotm.hpp
```diff
@@ -0,0 +1,168 @@
...
+                           stride_y,
+                           param[0],
+                           param[1],
+                           param[2],
+                           param[3],
+                           param[4],
+                           0);
+    else // array of params on host, copy to device
+    {
+        // This should NOT happen from calls from the API currently.
```
TorreZuk: I vote we don't support this.

daineAMD: This has been addressed in 474e196. We now support:
- batched `param` for rotm(g)_(strided_)batched.
- `param` can be located on either host or device for original rotm(g)
- `param` can only be located on device for (strided_)batched versions.

library/src/blas1/rocblas_rot.hpp
daineAMD: Added.

clients/include/testing_rotm_batched.hpp
daineAMD: Hm, I'm not sure I follow what you mean. This is the check that was done for device_batch_vectors in the original gemv_batched so I've been following that.

TorreZuk: One shouldn't have to check one of the list for failure, the allocator should return failure if any element of a vector fails to allocate.  This would be addressed in a new jira issue and PR.   Not sure if you are keen to create the issue.   It could be considered technical debt.  Can resolve this conversation.

clients/include/testing_rot_batched.hpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    rocblas_seedrand();
+    for(int b = 0; b < batch_count; b++)
+    {
+        rocblas_init<T>(hx[b], 1, N, incx);
+        rocblas_init<T>(hy[b], 1, N, incy);
+    }
+    rocblas_init<U>(hc, 1, 1, 1);
+    rocblas_init<V>(hs, 1, 1, 1);
```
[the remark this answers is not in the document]
TorreZuk: Doesn't this make sense?

clients/include/testing_rotm_batched.hpp
daineAMD: Ah, I see. Iirc Yvan is looking at improving device_vector and device_batch_vector? Could be brought to his attention.

clients/include/testing_rot_batched.hpp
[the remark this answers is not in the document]
daineAMD: Ah yes, that would make sense as that's the use case for this. I'll make this happen.

clients/include/testing_rotm_batched.hpp
TorreZuk: Sure I will fwd this to him to add to the issue he created.

clients/common/rocblas_gentest.py
```diff
@@ -215,6 +216,20 @@ def setdefaults(test):
             test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
                                             test['stride_scale']))
 
+    if test['function'] in ('rotg_strided_batched'):
+        if 'stride_scale' in test:
+            test.setdefault('stride_a', int(test['stride_scale']))
+            test.setdefault('stride_b', int(test['stride_scale']))
+            test.setdefault('stride_c', int(test['stride_scale']))
+            test.setdefault('stride_d', int(test['stride_scale']))
```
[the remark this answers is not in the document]
daineAMD: Added comments in 8f7c408.

clients/include/testing_rot_batched.hpp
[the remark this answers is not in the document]
daineAMD: Done in 8f7c408.
I take a random alpha (in range 0 - 10) and have c = cos(alpha); s = sin(alpha). This is in rads so I think this should be good enough.

PR #738, opened 2019-10-04
integration of the new tensile client

wbgilmartin:
This will add the initial hooks to call the new tensile client host library. 

I am still testing locally to see if the last set of changes added did not break anything but this will enable the start of the code review.

to build I added a switch `-s`, so use `./install.sh -cds`

by default the build will build the old client.

when running using the new host (In this iteration) add a `-lib` option to the call to `rocblas-bench` i.e.
`./rocblas-bench ...  -lib /path/to/host/library.`

a normal rocblas build will create the library and place it in `rocBLAS/build/release/Tensile/library`


library/src/tensile_host.cpp
```diff
@@ -0,0 +1,398 @@
...
+                                     unsigned long     stride_a,
+                                     const T*          B,
+                                     unsigned long       ld_b,
+                                     unsigned long     stride_b,
+                                     const T*          beta,
+                                     T*                C,
+                                     unsigned long       ld_c,
+                                     unsigned long     stride_c,
+                                     unsigned long     batchSize)
+{
```
sdquiring: We should figure out why this layer is necessary.

library/src/tensile_host.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    bool transposeA = false;
+    if (trans_a == rocblas_operation_conjugate_transpose)
```
sdquiring: How is normal transpose handled?

Conjugate transpose needs to be handled with a [TensorOp](https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Source/lib/include/Tensile/TensorOps.hpp).

library/src/tensile_host.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    return inputs;
+}
+
+class TensileHostImpl : public TensileHost
+{
+public:
+  void initializeHost(const char* lib_path)
+  {
+    //std::string dir ("/home/wgilmart/dev/wbgilmartin/tasks/new_client_integration/iteration3/Tensile/build/1_BenchmarkProblems/Cijk_Ailk_Bljk_SB_00/00_Final/source/library/*hsaco");
```
sdquiring: Cleanup

leekillough: > 
> 
> Cleanup

Being done.

library/src/tensile_host.cpp
leekillough: > 
> 
> How is normal transpose handled?
> 
> Conjugate transpose needs to be handled with a [TensorOp](https://github.com/ROCmSoftwarePlatform/Tensile/blob/develop/Tensile/Source/lib/include/Tensile/TensorOps.hpp).

Being refactored, but right now transpose and conjugate transpose are treated the same, and no `TensorOp` is created.

leekillough:
When building rocBLAS with `install.sh -cds -lasm_ci`, I get this error:
```
CMake Error at library/src/CMakeLists.txt:33 (target_compile_definitions):
  Cannot specify compile definitions for target "Tensile" which is not built
  by this project.

-- Configuring incomplete, errors occurred!
See also "/home/lkilloug/tmp/rocBLAS/build/release/CMakeFiles/CMakeOutput.log".
See also "/home/lkilloug/tmp/rocBLAS/build/release/CMakeFiles/CMakeError.log".
+ check_exit_code 1
+ ((  1 != 0  ))
+ exit 1
```

library/src/rocblas_auxiliary.cpp
```diff
@@ -59,6 +59,28 @@ extern "C" rocblas_status rocblas_set_pointer_mode(rocblas_handle handle, rocbla
     return rocblas_status_success;
 }
 
+#ifdef USE_TENSILE_HOST
+extern "C" rocblas_status rocblas_create_host_handle(rocblas_handle* handle, const char* lib_path)
+{
```
leekillough: We should not add  `rocblas_create_host_handle` -- `rocblas_create_handle` should work with the new or old Tensile code. If `lib_path` is needed for a library path, then it needs to be in an environment variable, such as `ROCBLAS_TENSILE_LIBPATH` or something else.

The normal `rocblas_create_handle` API and mechanism should initialize Tensile Host without requiring the user to call `rocblas_create_host_handle`.

leekillough: Being addressed in refactoring.

leekillough:
> 
> 
> When building rocBLAS with `install.sh -cds -lasm_ci`, I get this error:
> 
> ```
> CMake Error at library/src/CMakeLists.txt:33 (target_compile_definitions):
>   Cannot specify compile definitions for target "Tensile" which is not built
>   by this project.
> 
> -- Configuring incomplete, errors occurred!
> See also "/home/lkilloug/tmp/rocBLAS/build/release/CMakeFiles/CMakeOutput.log".
> See also "/home/lkilloug/tmp/rocBLAS/build/release/CMakeFiles/CMakeError.log".
> + check_exit_code 1
> + ((  1 != 0  ))
> + exit 1
> ```

Fixed in refactoring.

library/src/tensile_host.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    unsigned int batchCount = 1;
+
+    a.appendDim(batchCount);
+    b.appendDim(batchCount);
+    d.appendDim(batchCount);
+
+    batchIndices.push_back({2,2,2,2});
+
+    if(*beta != 0.0)
```
leekillough: `alpha` and `beta` can exist on either the device or host, depending on handle pointer mode.

If they exist on device, then `*alpha` and `*beta` cannot be dereferenced on host. They either need to wait and dereference on device, or `hipMemcpy` (synchronizing call) must be used to copy from device to host. This is best avoided if possible.

leekillough:
The `ContractionProblem` constructor only accepts `double` for beta, so we cannot pass complex to it.

sdquiring:
At this location it’s only used to differentiate between beta == 0, beta == 1, and other values.

On Oct 5, 2019, at 8:58 AM, Lee Killough <notifications@github.com<mailto:notifications@github.com>> wrote:

[CAUTION: External Email]

The ContractionProblem constructor only accepts double for beta, so we cannot pass complex to it.

leekillough:
> At this location it’s only used to differentiate between beta == 0, beta == 1, and other values.

We cannot read it from the host if it's on the device. We either need to copy from device to host, which is a synchronizing operation, or we have check for beta==0 or beta==1 in the device code. alpha and beta can either be on device or host, controlled by the handle pointer mode.

> The ContractionProblem constructor only accepts double for beta, so we cannot pass complex to it. 

ZGEMM and CGEMM require complex alpha, beta. When the Complex GEMM code was first added, we noticed that alpha, beta were not complex. This is unacceptable. I thought that alpha, beta were changed to complex. Many LAPACK algorithms use complex values of alpha, beta.

sdquiring:
Once again, the value sent in to the ContractionProblem constructor is ONLY used for branching based on the relevant values for beta.  The real value for beta is sent in via the ContractionInputs, which is templated, and expects a complex value for complex GEMMs.

It's only used to select a more efficient kernel in the case that beta is a special value.  For the case where the value of beta is on the GPU, we can use a value of e.g. 2.0 and we will get a kernel that will work for all cases, but it may be slightly less than optimal if beta happens to be 0.

I realize taking a double there is confusing and it should probably be replaced with an enumeration.

leekillough:
> 
> 
> Once again, the value sent in to the ContractionProblem constructor is ONLY used for branching based on the relevant values for beta. The real value for beta is sent in via the ContractionInputs, which is templated, and expects a complex value for complex GEMMs.
> 
> It's only used to select a more efficient kernel in the case that beta is a special value. For the case where the value of beta is on the GPU, we can use a value of e.g. 2.0 and we will get a kernel that will work for all cases, but it may be slightly less than optimal if beta happens to be 0.
> 
> I realize taking a double there is confusing and it should probably be replaced with an enumeration.

If `beta` is complex, then `imag(beta) != 0` should make it not compare equal to `0`, `1`, or other special real values.

We need to know what values of `beta` it considers "special", and compare complex `beta` to them, passing the special values as `double` when they are found:
```
(*beta == 0 ? 0.0 : *beta == 1 ? 1.0 : -12345.0)
```
`*beta == 0` and `*beta == 1` will do a complex compare and will be false if `imag(beta) != 0`. The `-12345.0` value is just a dummy value which won't compare equal to any special value.

However, if `beta` is on the device, then we cannot read it on the host without copying it from device to host first, causing a synchronization.

I much prefer `enum`s rather than passing `double`, because, as with complex and integer, we use many types different than `double`. An `enum` describing the value category is clearer and does not tie beta to a particular type.

leekillough:
> 
> 
> Once again, the value sent in to the ContractionProblem constructor is ONLY used for branching based on the relevant values for beta. The real value for beta is sent in via the ContractionInputs, which is templated, and expects a complex value for complex GEMMs.
> 
> It's only used to select a more efficient kernel in the case that beta is a special value. For the case where the value of beta is on the GPU, we can use a value of e.g. 2.0 and we will get a kernel that will work for all cases, but it may be slightly less than optimal if beta happens to be 0.
> 
> I realize taking a double there is confusing and it should probably be replaced with an enumeration.

A `value_category` function has been added to categorize a value. Right now it returns `double`, but it can be changed to return an `enum`.

library/src/rocblas_auxiliary.cpp
wbgilmartin: Is there an example of how to handle environment variables in rocblas?

wbgilmartin:
I updated this to get it to build and run but I had to disable the complex and half data types due to type conflicts between rocblas and tensile host library.

This may be challenging to resolve.

leekillough:
> 
> 
> I updated this to get it to build and run but I had to disable the complex and half data types due to type conflicts between rocblas and tensile host library.
> 
> This may be challenging to resolve.

See my PR: https://github.com/wbgilmartin/rocBLAS/pull/1

library/src/rocblas_auxiliary.cpp
leekillough: > 
> 
> Is there an example of how to handle environment variables in rocblas?

See my PR: https://github.com/wbgilmartin/rocBLAS/pull/1

library/src/blas3/Tensile/gemm_strided_batched.cpp
```diff
@@ -195,20 +182,92 @@ namespace
...
+                                             C,
+                                             ld_c,
+                                             stride_c,
+                                             b_c);
+
+        return handle->host->runContractionProblem(problem);
+
+#else
+        rocblas_status validArgs = validateArgs(handle,
+                                                trans_a,
```
leekillough: This argument validation needs to be moved out of the `#ifdef USE_TENSILE_HOST` and executed for both the old and new client code.

PR #739, opened 2019-10-04
trsm_batched and trsm_strided_batched

daineAMD:
Adds:
- trsm_batched
- trsm_ex_batched
- trsm_strided_batched
- trsm_ex_strided_batched

Potentially supports arrays of alphas at the _template level, although this is not tested as it is not supported in the API.

It also appears that moving gemm_ex functions into a shared _template function made its way into this PR as well.

This PR moves around code from files (mainly moving code from rocblas_trsm.cpp to rocblas_trsm.hpp) so it may be hard to follow the changes on GitHub. The most interesting code changes are likely located in rocblas_trsm.hpp. The memory management code was changed to support the batched version which is a little complicated, so extra review on that would be appreciated.


library/src/blas3/rocblas_trsm.hpp
```diff
@@ -0,0 +1,1743 @@
...
+                                         rocblas_fill      uplo,
+                                         rocblas_operation transA,
+                                         rocblas_diagonal  diag,
+                                         rocblas_int       m,
+                                         rocblas_int       n,
+                                         const T*          alpha,
+                                         rocblas_stride    stride_alpha,
+                                         U                 A,
+                                         rocblas_int       offset_Ain,
+
```
mahmoodw: I noticed this shows up twice now, in the _mem and template. Not sure if we need it in both, just thought I'd point it out.

clients/include/testing_trsm_batched.hpp
```diff
@@ -0,0 +1,424 @@
...
+    rocblas_int K      = side == rocblas_side_left ? M : N;
+    size_t      size_A = lda * size_t(K);
+    size_t      size_B = ldb * size_t(N);
+
+    rocblas_local_handle handle;
+    // check here to prevent undefined memory allocation error
+    if(M < 0 || N < 0 || lda < K || ldb < M || batch_count <= 0)
+    {
+        static const size_t     safe_size = 100; // arbitrarily set to 100
+        rocblas_int             num_batch = batch_count < 0 ? 1 : batch_count;
```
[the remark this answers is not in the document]
daineAMD: Removed.

clients/include/testing_trsm_batched.hpp
```diff
@@ -0,0 +1,424 @@
...
+                                                          diag,
+                                                          M,
+                                                          N,
+                                                          &alpha_h,
+                                                          dA,
+                                                          lda,
+                                                          dXorB,
+                                                          ldb,
+                                                          batch_count),
+                                  rocblas_status_invalid_size);
```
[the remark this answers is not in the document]
daineAMD: Good call, changed this.

clients/include/testing_trsm_batched.hpp
```diff
@@ -0,0 +1,424 @@
...
+        rocblas_gflops = trsm_gflop_count<T>(M, N, K) / gpu_time_used * 1e6;
+
+        // CPU cblas
+        cpu_time_used = get_time_us();
+
+        for(int b = 0; b < batch_count; b++)
+            cblas_trsm<T>(side, uplo, transA, diag, M, N, alpha_h, hA[b], lda, cpuXorB[b], ldb);
+
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = trsm_gflop_count<T>(M, N, K) / cpu_time_used * 1e6;
```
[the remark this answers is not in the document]
daineAMD: Ah, I missed this everywhere haha.

library/src/blas3/rocblas_trsm.hpp
daineAMD: I removed the logging in _mem for now since it's only used with the _template right now.
As for the prefix, I'll leave it for now since we log to stderr in other places I believe, but likely a good idea to implement in the future.

clients/include/testing_trsm_strided_batched.hpp
```diff
@@ -0,0 +1,453 @@
...
+                        res_1 += std::abs((hXorB_1[idx] - hB[idx]) / hB[idx]);
+                        res_2 += std::abs((hXorB_2[idx] - hB[idx]) / hB[idx]);
+                    }
+                    else
+                    {
+                        res_1 += std::abs(hXorB_1[idx]);
+                        res_2 += std::abs(hXorB_2[idx]);
+                    }
+                }
+                max_res_1 = max_res_1 > res_1 ? max_res_1 : res_1;
```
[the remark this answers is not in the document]
daineAMD: I'm sure using std::max wouldn't be a problem, I can go ahead and change it, I'm guessing this logic is used elsewhere as well, though.

library/src/blas3/rocblas_trsm.hpp
leekillough: Logging should only occur in the `_impl` functions.

clients/include/testing_trsm_strided_batched.hpp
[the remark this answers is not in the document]
leekillough: > 
> 
> I'm sure using std::max wouldn't be a problem, I can go ahead and change it, I'm guessing this logic is used elsewhere as well, though.

`std::` functions cannot generally be used in `__device__` code (and thus `__global__` kernels).

`std::abs` was replaced with `rocblas_abs`. If `max` is needed, then a `rocblas_max` should be added to `library/src/include/utility.h`

clients/include/testing_trsm_strided_batched.hpp
```diff
@@ -0,0 +1,439 @@
...
+    char char_diag   = arg.diag;
+    T    alpha_h     = arg.alpha;
+
+    rocblas_side      side   = char2rocblas_side(char_side);
+    rocblas_fill      uplo   = char2rocblas_fill(char_uplo);
+    rocblas_operation transA = char2rocblas_operation(char_transA);
+    rocblas_diagonal  diag   = char2rocblas_diagonal(char_diag);
+
+    rocblas_int K      = side == rocblas_side_left ? M : N;
+    size_t      size_A = lda * size_t(K) * batch_count + size_t(stride_a) * size_t(batch_count - 1);
```
mahmoodw: correct if I am wrong but I think this should be:
lda * size_t(K) + size_t(stride_a) * size_t(batch_count - 1);

leekillough: > 
> 
> correct if I am wrong but I think this should be:
> lda * size_t(K) + size_t(stride_a) * size_t(batch_count - 1);
```
lda * size_t(K) + stride_a * (batch_count -1)
```
There is no need to cast `stride_a` or any other `rocblas_stride` variables, since they are supposed to already be `ptrdiff_t` size. `batch_count - 1`  does not need to be cast because it will be automatically promoted.

library/src/blas3/rocblas_trsm.hpp
daineAMD: Moved stderr logging out of template function into _impl functions.

clients/include/testing_trsm_strided_batched.hpp
daineAMD: Good catch, fixed.

clients/include/testing_trsm_strided_batched.hpp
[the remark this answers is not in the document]
daineAMD: This is host/testing code so std:: functions can be used. I'm not going to change it here as we use this logic in quite a few other files for testing; this PR is big enough. If we want to change these to std::max (or otherwise) we should do another refactoring PR I think.

library/src/blas3/rocblas_trsm.hpp
```diff
@@ -0,0 +1,1737 @@
...
```
mahmoodw: I expect the cast type for x_temparr/x_temp in these 2 spots to be V since they both hold the output values.

clients/include/testing_trsm_ex_strided_batched.hpp
```diff
@@ -0,0 +1,494 @@
...
+    rocblas_int batch_count = arg.batch_count;
+
+    rocblas_side      side   = char2rocblas_side(char_side);
+    rocblas_fill      uplo   = char2rocblas_fill(char_uplo);
+    rocblas_operation transA = char2rocblas_operation(char_transA);
+    rocblas_diagonal  diag   = char2rocblas_diagonal(char_diag);
+
+    rocblas_int K      = side == rocblas_side_left ? M : N;
+    size_t      size_A = lda * size_t(K) + stride_A * batch_count - 1;
+    size_t      size_B = ldb * size_t(N) + stride_B * batch_count - 1;
```
TorreZuk: Brackets got dropped on batch_count - 1

clients/include/testing_trsm_batched.hpp
```diff
@@ -0,0 +1,403 @@
...
+        double gpu_time_used, cpu_time_used;
+        double rocblas_gflops, cblas_gflops;
+
+        // GPU rocBLAS
+        for(int b = 0; b < batch_count; b++)
+        {
+            CHECK_HIP_ERROR(
+                hipMemcpy(XorBv[b], hXorB_1[b], sizeof(T) * size_B, hipMemcpyHostToDevice));
+        }
+        CHECK_HIP_ERROR(hipMemcpy(dXorB, hXorB_1, sizeof(T*) * batch_count, hipMemcpyHostToDevice));
```
mahmoodw: I think hXorB_1 should be XorBv

clients/gtest/trsm_ex_gtest.cpp
```diff
@@ -7,13 +7,72 @@
...
+#include "testing_trsm_ex_batched.hpp"
+#include "testing_trsm_ex_strided_batched.hpp"
 #include "type_dispatch.hpp"
 #include <cctype>
 #include <cstring>
 #include <type_traits>
 
 namespace
 {
+    // possible trsm_ex test cases
```
mahmoodw: Just noticed that gemm and gemm_ex are combined into a single *gtest.cpp file. The same could be done for trsm.

clients/include/rocblas.hpp
```diff
@@ -1273,6 +1273,111 @@ static constexpr auto rocblas_trsm<float> = rocblas_strsm;
 template <>
 static constexpr auto rocblas_trsm<double> = rocblas_dtrsm;
 
+// trsm_ex
```
mahmoodw: Looks like you forgot to handle these comments.

clients/include/testing_trsm_batched.hpp
```diff
@@ -0,0 +1,403 @@
...
+
+        if(arg.norm_check)
+            std::cout << ",CPU-Gflops,us,norm_error_host_ptr,norm_error_dev_ptr";
+
+        std::cout << std::endl;
+
+        std::cout << M << ',' << N << ',' << lda << ',' << ldb << ',' << char_side << ','
+                  << char_uplo << ',' << char_transA << ',' << char_diag << ',' << rocblas_gflops
+                  << "," << gpu_time_used;
+
```
mahmoodw: Probably want to add batch_count to the timing print of each testing_batched file.

clients/include/testing_trsm_batched.hpp
daineAMD: All good points. Correcting them all, should all be fixed in the next commit, just waiting for tensile to build to make sure I didn't break anything.

library/src/blas3/rocblas_trsm.hpp
daineAMD: On line 1719 I'm essentially casting `const` to a non-`const` value since the copy function is expected a const value, so I think it's fine there. I'm not sure which other spot you're referring to.

TorreZuk: maybe a inline comment to improve readability,

clients/include/testing_trsm_batched.hpp
daineAMD: Fixed in 9d378b0.

clients/gtest/trsm_ex_gtest.cpp
daineAMD: Combined trsm_gtest.cpp and trsm_ex_gtest.cpp into a single trsm_gtest.cpp in 9d378b0.

clients/include/rocblas.hpp
daineAMD: Removed in 9d378b0. Don't need them since the _ex functions don't have template specializations.

clients/include/testing_trsm_batched.hpp
daineAMD: Done, 9d378b0.

clients/include/testing_trsm_ex_batched.hpp
```diff
@@ -0,0 +1,466 @@
...
+
+        if(!dA || !dXorB)
+        {
+            CHECK_HIP_ERROR(hipErrorOutOfMemory);
+            return;
+        }
+
+        // TODO change this to trsm_ex_batched
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
+        rocblas_status status = rocblas_trsm_batched<T>(
```
mahmoodw: should be the ex version?

clients/include/testing_trsm_ex_strided_batched.hpp
```diff
@@ -0,0 +1,481 @@
...
+        device_vector<T>    dA(safe_size);
+        device_vector<T>    dXorB(safe_size);
+        if(!dA || !dXorB)
+        {
+            CHECK_HIP_ERROR(hipErrorOutOfMemory);
+            return;
+        }
+        // TODO change this to trsm_ex_strided_batched
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
+        rocblas_status status = rocblas_trsm_strided_batched<T>(handle,
```
mahmoodw: Again here

clients/include/testing_trsm_ex_batched.hpp
daineAMD: Yeah, see 3d57ee7.

clients/common/rocblas_gentest.py
```diff
@@ -247,6 +247,24 @@ def setdefaults(test):
...
+                test.setdefault('stride_a', int(test['M'] * test['lda'] *
+                                                test['stride_scale']))
+                test.setdefault('stride_c', int(test['M'] * test['lda'] *
+                                                test['stride_scale']))
+        else:
+            if all([x in test for x in ('N', 'lda', 'stride_scale')]):
+                test.setdefault('stride_a', int(test['N'] * test['lda'] *
+                                                test['stride_scale']))
+                test.setdefault('stride_a', int(test['N'] * test['lda'] *
+                                                test['stride_scale']))
```
leekillough: `stride_a` set twice?

The same expressions used twice? Better (unsure of the logic and the order of the if-nests -- indentation is critical):
```
     if all([x in test for x in ('M', 'lda', 'stride_scale')]):
         ldM = int(test['M'] * test['lda'] * test['stride_scale'])
         test.setdefault('stride_a", ldM)
         test.setdefault('stride_c', ldM)
```
and:
```
         ldN = int(test['N'] * test['lda'] * test['stride_scale'])
         test.setdefault('stride_a', ldN)
         test.setdefault('stride_a', ldN)    # ??? stride_a twice

Comment (leekillough) [reply clients/common/rocblas_gentest.py]: Also, if there is no overlap between functions, they should be exclusive (this is assuming the above indention is correct):

```
    if test['function'] in ('asum_strided_batched', 'nrm2_strided_batched',
                            'scal_strided_batched', 'swap_strided_batched',
                            'copy_strided_batched', 'dot_strided_batched',
                            'dotc_strided_batched', 'rot_strided_batched',
                            'rotm_strided_batched'):
        if all([x in test for x in ('N', 'incx', 'stride_scale')]):
            test.setdefault('stride_x', int(test['N'] * abs(test['incx']) *
                                            test['stride_scale']))
        if all([x in test for x in ('N', 'incy', 'stride_scale')]):
            test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
                                            test['stride_scale']))
        # we are using stride_c for param in rotm
        if all([x in test for x in ('stride_scale')]):
            test.setdefault('stride_c', int(test['stride_scale']) * 5)

    elif test['function'] in ('gemv_strided_batched', 'ger_strided_batched'):
        if test['function'] in ('ger_strided_batched') or test['transA'] in ('T', 'C'):
            if all([x in test for x in ('M', 'incx', 'stride_scale')]):
                test.setdefault('stride_x', int(test['M'] * abs(test['incx']) *
                                                test['stride_scale']))
            if all([x in test for x in ('N', 'incy', 'stride_scale')]):
                test.setdefault('stride_y', int(test['N'] * abs(test['incy']) *
                                                test['stride_scale']))
        else:
            if all([x in test for x in ('N', 'incx', 'stride_scale')]):
                test.setdefault('stride_x', int(test['N'] * abs(test['incx']) *
                                                test['stride_scale']))
            if all([x in test for x in ('M', 'incy', 'stride_scale')]):
                test.setdefault('stride_y', int(test['M'] * abs(test['incy']) *
                                                test['stride_scale']))

    # we are using stride_c for arg c and stride_d for arg s in rotg
    # these are are single values for each batch
    elif test['function'] in ('rotg_strided_batched'):
        if 'stride_scale' in test:
            test.setdefault('stride_a', int(test['stride_scale']))
            test.setdefault('stride_b', int(test['stride_scale']))
            test.setdefault('stride_c', int(test['stride_scale']))
            test.setdefault('stride_d', int(test['stride_scale']))

    # we are using stride_a for d1, stride_b for d2, and stride_c for param in
    # rotmg. These are are single values for each batch, except param which is
    # a 5 element array
    elif test['function'] in ('rotmg_strided_batched'):
        if 'stride_scale' in test:
            test.setdefault('stride_a', int(test['stride_scale']))
            test.setdefault('stride_b', int(test['stride_scale']))
            test.setdefault('stride_c', int(test['stride_scale']) * 5)
            test.setdefault('stride_x', int(test['stride_scale']))
            test.setdefault('stride_y', int(test['stride_scale']))

    elif test['function'] in ('trsm_strided_batched', 'trsm_ex_strided_batched'):
        if all([x in test for x in ('N', 'ldb', 'stride_scale')]):
            test.setdefault('stride_b', int(test['N'] * test['ldb'] *
                                            test['stride_scale']))
        if test['side'].upper() == 'L':
            if all([x in test for x in ('M', 'lda', 'stride_scale')]):
                test.setdefault('stride_a', int(test['M'] * test['lda'] *
                                                test['stride_scale']))
                test.setdefault('stride_c', int(test['M'] * test['lda'] *
                                                test['stride_scale']))
        else:
            if all([x in test for x in ('N', 'lda', 'stride_scale')]):
                test.setdefault('stride_a', int(test['N'] * test['lda'] *
                                                test['stride_scale']))
                test.setdefault('stride_a', int(test['N'] * test['lda'] *
                                                test['stride_scale']))
```

File: library/src/blas3/rocblas_trsm.hpp (hunk 7 above)
Comment (leekillough) [inline library/src/blas3/rocblas_trsm.hpp]: We should be avoiding `hipMemcpy`.

Comment (leekillough) [inline library/src/blas3/rocblas_trsm.hpp]: We should be avoiding `hipMemcpy`. 

We should be computing it on device rather than host.

Comment (leekillough) [inline library/src/blas3/rocblas_trsm.hpp]: These should be computed on device rather than host.

File: library/src/blas3/rocblas_trsm.hpp [hunk 16]
```diff
@@ -0,0 +1,1737 @@
...
+    /* ===============left==================================================== */
+
+    template <rocblas_int BLOCK, bool BATCHED, typename T, typename U, typename V>
+    rocblas_status rocblas_trsm_left(rocblas_handle    handle,
+                                     rocblas_fill      uplo,
+                                     rocblas_operation transA,
+                                     rocblas_int       m,
+                                     rocblas_int       n,
+                                     const T*          alpha,
+                                     rocblas_stride    stride_alpha,
```
leekillough: `alpha` should be scalar. I thought we decided not to support multiple scalars.

library/src/blas3/rocblas_trsm.hpp
leekillough: We can pass pointers to device instead of using `hipMemcpy`.

leekillough: This introduces synchronization at the end of every `hipMemcpy()` call, and may negate the advantages of batched.

daineAMD:
I'm not sure that not having A as a batched parameter is a solution. We would have to talk to Juan to see if he wants A batched. Currently, only the batched version (not the strided batched version) introduced a hipMemcpy. I can look at it and hopefully find a way to avoid it.

library/src/blas3/rocblas_trsm.hpp
daineAMD: This hipMemcpy was not introduced here, it was in the previous version of trsm as well.

daineAMD: This one was introduced by me. I will work on a way around this.

library/src/blas3/rocblas_trsm.hpp
daineAMD: Only a single scalar is supported through the API, multiple scalars are supported at the template level. This similar functionality was supported in some of the L1 functions as Juan specifically requested it, although we are assured that in that case they will be only on the device, which I did not deal with in the PR at this time.

leekillough:
> 
> 
> I'm not sure that not having A as a batched parameter is a solution. We would have to talk to Juan to see if he wants A batched (or see what is done elsewhere in other libraries). Currently, only the batched version (not the strided batched version) introduced a hipMemcpy. I can look at it and hopefully find a way to avoid it.

Shoreline should be the driver, not rocSolver.

library/src/blas3/rocblas_trsm.hpp
leekillough: > 
> 
> This hipMemcpy was not introduced here, it was in the previous version of trsm as well.
> 
> Edit: That's not completely true, the one directly below this was in the previous version, this is just the port of that for multiple alphas. This would only ever be called internally as it is not available through the API.

Well, it is a synchronization point, and we should try to minimize these.

clients/common/rocblas_gentest.py
daineAMD: Unsure of the duplicate or why I ever had stride_c in trsm since it isn't used. Made if statements exclusive. Made indentation easier to read as per suggested. See 64aa460.

library/src/blas3/rocblas_trsm.hpp
daineAMD: I removed the `hipmemcpy`s which copied the pointers from host to device, introduced a device function to do so instead. See 1aa6d9f.

library/src/blas3/rocblas_trsm.hpp
```diff
@@ -86,6 +86,35 @@ namespace
...
+        hipLaunchKernelGGL(setup_batched_array_kernel<T>,
+                           grid,
+                           threads,
+                           0,
+                           handle->rocblas_stream,
+                           src,
+                           src_stride,
+                           dst);
+    }
+
```
mahmoodw: Maybe this should be moved to a common file to be used with trsv

library/src/blas3/rocblas_trsm.hpp
daineAMD: Removed multiple scalars in 66c771e since they will be removed in gemm anyway.

library/src/blas3/rocblas_trsm.hpp
daineAMD: Removed multiple scalars so reverted to what it used to be.

library/src/blas3/rocblas_trsm.hpp
daineAMD: I moved it to the trtri_trsm.hpp file in 11313d6. If anyone thinks it should be moved somewhere else let me know.

PR #741, opened 2019-10-07
Fixing SLES tests LD_LIBRARY_PATH and refactoring tests

saadrahim:


TorreZuk:
Merged so can be integrated into 2.10

saadrahim:
Thanks

PR #748, opened 2019-10-10
Handle whitespace at the end of lines, and add newline at end of file if it doesn't exist

leekillough:
`clang-format` does not, AFAIK, have options for removing trailing whitespace at the end of lines, or adding a missing newline at the end of a file.

The `emacs` `.dir-locals.el` file I added, automatically performs these edits upon saving a file, but for people who use `vim` or other editors, we need a consistent solution.

This changes the `pre-commit` script to run `sed` commands to strip trailing whitespace at the end of lines, and to add missing newlines at the end of files, on the same files for which `clang-format` is run.

There are also some simplifications of the `pre-commit` script, which previously used `/bin/sh` but for which `/bin/bash` adds more features.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,30 +1,24 @@
-#!/bin/sh
+#!/bin/bash
 #
 # This pre-commit hook checks if any versions of clang-format
 # are installed, and if so, uses the installed version to format
 # the staged changes.
 
-base=/opt/rocm/hcc/bin/clang-format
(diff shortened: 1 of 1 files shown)


leekillough:
The relevant commands to add:

```
        sed -i -e 's/[[:space:]]*$//' "$file" # Remove whitespace at end of lines
        sed -i -e '$a\' "$file" # Add missing newline to end of file
```

PR #750, opened 2019-10-15
Switching CI back to full_asm build

mahmoodw:
This change ensures all kernels are being built and tested

Please inform if there are memory restrictions that need to be considered for a full rocBLAS build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ rocBLASCI:
 
     def rocblas = new rocProject('rocBLAS')
     // customize for project
-    rocblas.paths.build_command = './install.sh -lasm_ci -c'
+    rocblas.paths.build_command = './install.sh -c'
 
     // Define test architectures, optional rocm version argument is available
     def nodes = new dockerNodes(['ubuntu && gfx900', 'centos7 && gfx900', 'centos7 && gfx906', 'sles && gfx906'], rocblas)


amdkila (dismissed):

saadrahim:
Let us see how long this takes. Will thins increase the test suite size?

amdkila:
I assume we're going to get rid of asm_ci? Also @saadrahim it will increase the compile time as this adds more Tensile kernels than asm_ci and will increase the test suite size as well. We should add a gfx908 job for rocBLAS as well

zaliu (changes requested):
As I mentioned previously, there is absolutely no reason to go for asm_full build on gfx900 and gfx906 CI machines.  We want to do asm_full build only on gfx908 CI machines.

The memory footprint of a asm_full build is about 44GB.  If a CI machine has 8 GPUs and 256GB physical memory, on average a fully loaded 8 CI runs, each gets only about 32GB, which might cause occasional spurious build failures.

mahmoodw:
> As I mentioned previously, there is absolutely no reason to go for asm_full build on gfx900 and gfx906 CI machines. We want to do asm_full build only on gfx908 CI machines.
> 
> The memory footprint of a asm_full build is about 44GB. If a CI machine has 8 GPUs and 256GB physical memory, on average a fully loaded 8 CI runs, each gets only about 32GB, which might cause occasional spurious build failures.

Can @saadrahim or @amdkila please make the changes to build on the appropriate architectures?

amdkila:
> > As I mentioned previously, there is absolutely no reason to go for asm_full build on gfx900 and gfx906 CI machines. We want to do asm_full build only on gfx908 CI machines.
> > The memory footprint of a asm_full build is about 44GB. If a CI machine has 8 GPUs and 256GB physical memory, on average a fully loaded 8 CI runs, each gets only about 32GB, which might cause occasional spurious build failures.
> 
> Can @saadrahim or @amdkila please make the changes to build on the appropriate architectures?

Memory should be ok as we have 64GB of RAM per GPU on our gfx908 node. One issue we have is related to packaging and the number of GPUs- in order to create rocBLAS debian and rpm packages with asm_full, we'd need 2 gfx908 jobs. We don't have the resources to do that right now.

PR #751, opened 2019-10-15
Fix build for hipclang

AlexBrownAMD:
Function definitions in hpp files need inline.
Templated functions with multiple template params need HIP_KERNEL_NAME.
Fix rotm kernel templating.

### DIFF
--- a/library/src/blas1/reduction_strided_batched.h
+++ b/library/src/blas1/reduction_strided_batched.h
@@ -148,7 +148,7 @@ struct rocblas_default_value
     }
 };
 
-size_t rocblas_reduction_kernel_block_count(rocblas_int n, rocblas_int NB)
+inline size_t rocblas_reduction_kernel_block_count(rocblas_int n, rocblas_int NB)
 {
     if(n <= 0)
         n = 1; // avoid sign loss issues
--- a/library/src/blas1/rocblas_dot.hpp
+++ b/library/src/blas1/rocblas_dot.hpp
@@ -91,7 +91,7 @@ rocblas_status rocblas_dot_template(rocblas_handle __restrict__ handle,
     dim3        grid(blocks, batch_count);
     dim3        threads(NB);
 
-    hipLaunchKernelGGL(HIP_KERNEL_NAME(dot_kernel_part1<NB, CONJ, T>),
+    hipLaunchKernelGGL((dot_kernel_part1<NB, CONJ, T>),
                        grid,
                        threads,
                        0,
--- a/library/src/blas1/rocblas_dot.hpp
+++ b/library/src/blas1/rocblas_dot.hpp
@@ -91,7 +91,7 @@ rocblas_status rocblas_dot_template(rocblas_handle __restrict__ handle,
     dim3        grid(blocks, batch_count);
     dim3        threads(NB);
 
-    hipLaunchKernelGGL(dot_kernel_part1<NB, CONJ, T>,
+    hipLaunchKernelGGL(HIP_KERNEL_NAME(dot_kernel_part1<NB, CONJ, T>),
                        grid,
                        threads,
                        0,
--- a/library/src/blas1/rocblas_rotm.hpp
+++ b/library/src/blas1/rocblas_rotm.hpp
@@ -86,7 +86,7 @@ __global__ void rotm_kernel_batched(rocblas_int    n,
                      h22);
 }
 
-template <typename T>
+template <typename T, typename U>
 __global__ void rotm_kernel_regular(rocblas_int    n,
                                     T*             x_in,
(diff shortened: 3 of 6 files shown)


daineAMD (approved):
Looks like you might need to do a clang format for jenkins, looks good to me otherwise.

leekillough:
> Function definitions in hpp files need inline.

... or `static`, `constexpr` or `template`.

Function or class templates can be redefined if the same tokens are used in each translation unit. You do not need `inline` in every case.

Variable templates must be `inline` or `static` if defined in multiple header files. `constexpr` no longer implies `static` for variable templates (this is a retroactive language defect fix for C++14).

> Templated functions with multiple template params need HIP_KERNEL_NAME.

Correction:  **Macro arguments with commas inside of them, such as in template argument lists, must be parenthesized to avoid syntactic ambiguity.**

`HIP_KERNEL_NAME` is a no-op which just adds confusion. You can just use parentheses:
```
 hipLaunchKernelGGL((dot_kernel_part1<NB, CONJ, T>),
```

leekillough (changes requested):
Use simple parentheses around macro arguments to disambiguate nested commas, instead of using the confusing `HIP_KERNEL_NAME` macro..

AlexBrownAMD:
> Function or class templates can be redefined if the same tokens are used in each translation unit. You do not need `inline` in every case.

This does not appear to be the case in hipclang: functions defined in headers without a qualifier like inline cause "multiple definition" errors at link time.

> `HIP_KERNEL_NAME` is a no-op which just adds confusion. You can just use parentheses:

Pushed an update removing HIP_KERNEL_NAME. Though, I personally think having it there makes the code more clear and readable. It calls attention to the fact that the parentheses are there for a reason.

leekillough:
> > Function or class templates can be redefined if the same tokens are used in each translation unit. You do not need `inline` in every case.
> 
> This does not appear to be the case in hipclang: functions defined in headers without a qualifier like inline cause "multiple definition" errors at link time.

Templates should be an exception, if they are defined with the same tokens, such as including the same header with the same template definition multiple times:

> There can be more than one definition in a program of each of the following: class type, enumeration type, inline function, inline variable (since C++17), class template, **non-static function template**, static data member of a class template, member function of a class template, partial template specialization, concept (since C++20), as long as all of the following is true: 
> * each definition appears in a different translation unit 
> ...
> * each definition consists of the same sequence of tokens (typically, appears in the same header file)
> ...
 > * the language linkage is the same (e.g. the include file isn't inside an extern "C" block) 
> ...
> * if the definition is for a template, then all these requirements apply to both names at the point of definition and dependent names at the point of instantiation 
>
> If all these requirements are satisfied, the program behaves as if there is only one definition in the entire program. Otherwise, the behavior is undefined. 
>
> https://en.cppreference.com/w/cpp/language/definition

**Non-templated** functions need to be `inline`, `static` or `constexpr`, or in an unnamed namespace (which isn't usually a good idea in headers), to appear in multiple translation units. I usually use `static` if there's no benefit to inlining it and I really want it defined in a header file; otherwise I use `inline`. If the function is simple and can appear in a constant expression, I use `constexpr`, which implies `inline`.

(If a function template is declared `static` in a header file and appears in multiple compilation units, then it is internal to each compilation unit, so there is no conflict between them, no multiple definition, and no requirement that they have the same token sequence.)

> > `HIP_KERNEL_NAME` is a no-op which just adds confusion. You can just use parentheses:
> 
> Pushed an update removing HIP_KERNEL_NAME. Though, I personally think having it there makes the code more clear and readable. It calls attention to the fact that the parentheses are there for a reason.

I don't think it makes it any clearer -- it sounds like a macro which has to be called to "get" the name of a kernel from something passed to it. We used to use "get kernel name" type functions in GEMM, for a different purpose (to get the names of low-level Tensile kernels which get called for particular GEMM arguments), but they are slowly being obsoleted.

Also, is `HIP_KERNEL_NAME` defined by HIP, or by us? We should not misappropriate `HIP_*` names.

leekillough (approved):
Approved, but with the caution to keep in mind that there are multiple ways of defining functions in header files `#include`'d in multiple compilation units:

1. `inline`
2. `constexpr` (implies `inline`)
3. `static`
4. `template` (does not require `inline` or `static`)
5. Combinations of the above
6. (Not recommended.) `#define` macros to change the name in each compilation unit, use unnamed namespaces, etc.

Which one to use for a particular function, may depend on the circumstances.

leekillough:
Looks like some `clang-format`ting issues in`library/src/blas3/Tensile/gemm.hpp`

AlexBrownAMD:
Whoops, somehow missed gemm.hpp in my clang-format pass - got it updated.

PR #752, opened 2019-10-16
SWDEV-205693-hip-platform-warning

TorreZuk:
Summary of proposed changes:
Lets hip include auto set HIP_PLATFORM_HCC when using HIP tools


clients/samples/CMakeLists.txt
```diff
@@ -51,7 +51,9 @@ foreach( exe ${sample_list} )
   else( )
     # Unfortunatley, we can't link to hip::hip_hcc because it pulls in flags specific to the
```
leekillough: `Unfortunatley`

TorreZuk: It is now fixed.

PR #753, opened 2019-10-16
correct gemm k==0 quick return

amcamd:
- correct gemm quick return for k==0 so it is the same as in the Legacy Netlib blas

### DIFF
--- a/library/src/blas3/Tensile/gemm.hpp
+++ b/library/src/blas3/Tensile/gemm.hpp
@@ -842,7 +842,11 @@ inline rocblas_status rocblas_gemm_template(rocblas_handle    handle,
             trans_a, trans_b, m, n, k, ld_a, &stride_a, ld_b, &stride_b, ld_c, &stride_c);
     }
 
-    if(m == 0 || n == 0 || k == 0 || b_c == 0)
+    if(m == 0 || n == 0 || b_c == 0)
+    {
+        return rocblas_status_success;
+    }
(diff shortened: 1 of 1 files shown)


leekillough:
`k == 0` should be handled the same as `alpha == 0`, meaning `c = c * beta` (or `c = 0` if `beta == 0`).

There are `m * n` results in `c`. If `k == 0` or `alpha == 0`, then all of those results are `c * beta` (or `0` if `beta == 0`).

So `k == 0` is an early-exit, but it's not the same as a no-op, since `c` is still multiplied by `beta`.

```
*
*     Quick return if possible.
*
      IF ((M.EQ.0) .OR. (N.EQ.0) .OR.
     +    (((ALPHA.EQ.ZERO).OR. (K.EQ.0)).AND. (BETA.EQ.ONE))) RETURN
*
*     And if  alpha.eq.zero.
*
      IF (ALPHA.EQ.ZERO) THEN         ! Should also consider K.EQ.0
          IF (BETA.EQ.ZERO) THEN
              DO 20 J = 1,N
                  DO 10 I = 1,M
                      C(I,J) = ZERO
   10             CONTINUE
   20         CONTINUE
          ELSE
              DO 40 J = 1,N
                  DO 30 I = 1,M
                      C(I,J) = BETA*C(I,J)
   30             CONTINUE
   40         CONTINUE
          END IF
          RETURN
      END IF
```

Some of these changes ideally need to be handled in Tensile, such as testing `alpha == 0` when alpha is in device memory; others can be handled in rocBLAS host code.

This is being examined right now across all of rocBLAS; see #722

PR #755, opened 2019-10-17
Vectors proposal

YvanMokwinski:
Here is a draft of extended functionalities on vectors for testings, to make the code simpler.
It compiles (and used succesfully in one of my branch on axpy)  and is still compatible with its current use. There is some code redundancies I voluntarily kept. 

Notes:
-a/ constructors in (device|host)_strided_batched_vector take these parameters (n,inc,stride,batch_count) we could do the same for all the other classes. 

-b/ the cast operators on the device batched vector should either be removed or it should return a pointer on device rather than a pointer on host. 

Please give feed back to improve these objects that should make life easier when writing tests.


clients/include/d_vector.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+                // Point to guard before allocated memory
+                d -= PAD;
+
+                // Copy device memory after allocated memory to host
+                hipMemcpy(host, d, sizeof(guard), hipMemcpyDeviceToHost);
+
+                // Make sure no corruption has occurred
+                EXPECT_EQ(memcmp(host, guard, sizeof(guard)), 0);
+            }
```
daineAMD: This section looks like it's the same as device_vector_check(), is there a reason we aren't just calling it (or getting rid of the funtion)?

clients/include/device_batch_vector.hpp
```diff
@@ -0,0 +1,236 @@
...
+class device_batch_vector : private d_vector<T, PAD, U>
+{
+public:
+    //!
+    //! @brief Disallow copying.
+    //!
+    device_batch_vector(const device_batch_vector&) = delete;
+
+    //!
+    //! @brief Disallow or assigning.
```
daineAMD: typo "or".

clients/include/device_batch_vector.hpp
```diff
@@ -0,0 +1,236 @@
...
+    //! @brief Constructor.
+    //! @param batch_count The number of vectors.
+    //! @param size_vector The size of each vectors.
+    //!
+    explicit device_batch_vector(rocblas_int batch_count, size_t size_vector)
+        : m_batch_count(batch_count)
+        , m_size_vector(size_vector)
+        , d_vector<T, PAD, U>(size_vector)
+    {
+        if(false == this->try_initialize_memory())
```
daineAMD: `if(!try_initialize_memory())`?

clients/include/device_batch_vector.hpp
```diff
@@ -0,0 +1,236 @@
...
+    //! @param that The host_batch_vector to copy.
+    //!
+    inline hipError_t transfer_from(const host_batch_vector<T>& that) noexcept
+    {
+        //
+        // Check sizes.
+        //
+        if((this->batch_count() != that.batch_count()) || (this->size() != that.size()))
+        {
+            return hipErrorInvalidContext;
```
TorreZuk: Is this consistent with hip interpretation?  Maybe hipErrorUnknown ?   Or hipErrorInvalidValue which is what  would happen if one array is extended with null_ptr to match.

clients/include/device_strided_batch_vector.hpp
```diff
@@ -0,0 +1,235 @@
...
+        , m_stride(stride)
+        , m_batch_count(batch_count)
+    {
+        bool valid_parameters = true;
+
+        switch(this->m_storage)
+        {
+        case storage::block:
+        {
+            if(this->m_stride < this->m_n * this->m_inc)
```
TorreZuk: so only positive stride for now. okay

clients/include/device_strided_batch_vector.hpp
```diff
@@ -0,0 +1,235 @@
...
+            {
+                valid_parameters = false;
+            }
+            break;
+        }
+        case storage::interleave:
+        {
+            if(this->m_inc < this->m_stride * this->m_batch_count)
```
TorreZuk: Not sure what you are proposing here, but the m_inc type is smaller than stride type

clients/include/host_batch_vector.hpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+        return this->m_data;
+    }
+
+    //!
+    //! @brief Copy from a host batched vector.
+    //! @param that the vector the data is copied from.
+    //! @return true if the copy is done successfully, false otherwise.
+    //!
+    bool copy_from(const host_batch_vector<T>& that) noexcept
```
TorreZuk: Wouldn't operator= be acceptable here?

clients/include/device_batch_vector.hpp
YvanMokwinski: ! is too small sometimes :)

clients/include/device_batch_vector.hpp
YvanMokwinski: thx!

clients/include/device_batch_vector.hpp
YvanMokwinski: I found this one meaningful, but I'm open to any other.

clients/include/device_strided_batch_vector.hpp
YvanMokwinski: No, needs a correction. Good point.

clients/include/device_strided_batch_vector.hpp
YvanMokwinski: block or interleaved data

clients/include/d_vector.hpp
YvanMokwinski: > Just began, will keep reviewing on Monday. It might be useful to see an example of testing with the new vectors (your axpy testing, perhaps).

Right, I will give examples

YvanMokwinski: > This section looks like it's the same as device_vector_check(), is there a reason we aren't just calling it (or getting rid of the funtion)?

good point, didn't touch this class though (except adding a getter).

clients/include/host_batch_vector.hpp
YvanMokwinski: Originally. assignment operators have been  specified as 'deleted'.
Moreover, it's been written in this way to be able to catch that something wrong is happening during the copying operation without throwing an exception.

YvanMokwinski:
> Just began, will keep reviewing on Monday. It might be useful to see an example of testing with the new vectors (your axpy testing, perhaps).

an example: 

host_batch_vector<T> hx(size_x, batch_count,);
...
device_batch_vector<T> dx(batch_count, size_x);
if (!dx) CHECK_HIP_ERROR(hipErrorOutOfMemory);


...
// TRANSFER
CHECK_HIP_ERROR(dx.transfer_from(hx));
...

// TRANSFER
CHECK_HIP_ERROR(hx.transfer_from(dx));


 // COPY
host_batch_vector<T> htmp(size_x, batch_count);
 if (!htmp.copy_from(hx))
 {
     CHECK_HIP_ERROR(hipErrorInvalidContext);
 }

clients/include/device_batch_vector.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+    //!
+    //! @brief Constant random access.
+    //! @param batch_index The batch index.
+    //! @return Constant pointer to the array on device.
+    //!
+    inline const T* operator[](rocblas_int batch_index) const noexcept
+    {
+        assert(0 <= batch_index && this->m_batch_count > batch_index);
+
```
leekillough: `assert()` should never be used to catch user errors -- it becomes a no-op in release mode.

If it needs to still be there in release mode, some other way of indicating error is needed (but Google tests makes it hard since it doesn't use exceptions to indicate or throw errors).

clients/include/device_batch_vector.hpp
leekillough: We should define our own `rocblas_status_*` error codes (or use ones which exist), instead of returning `hipError_t`. There are macros for converting from `hipError_t` to `rocblas_status` if we have to detect errors in HIP calls.

clients/include/device_batch_vector.hpp
YvanMokwinski: Would it be better?

return (0 <= batch_index && this->m_batch_count > batch_index)
 ? m_data[batch_index]
 :  nullptr;

clients/include/device_batch_vector.hpp
```diff
@@ -0,0 +1,236 @@
...
+        return hipSuccess;
+    }
+
+private:
+    rocblas_int m_batch_count{0};
+    size_t      m_size_vector{0};
+    T**         m_data{nullptr};
+    T**         m_device_data{nullptr};
+
```
leekillough: `{}` works for all 4 too.
```
    rocblas_int m_batch_count{};
    size_t      m_size_vector{};
    T**         m_data{};
    T**         m_device_data{};
```
minimalistic, but maybe not as documenting.

clients/include/device_batch_vector.hpp
YvanMokwinski: Currently, CHECK_HIP_ERROR is widely used in testings, especially from HIP calls (transfer).  So I am trying to keep tight with the existing 'philosophy'. But I'm willing to change the error type.

clients/include/device_batch_vector.hpp
YvanMokwinski: I prefer explicit values rather than implicit ones.

clients/include/device_strided_batch_vector.hpp
```diff
@@ -0,0 +1,235 @@
...
+        {
+            this->m_data = this->device_vector_setup();
+        }
+    };
+
+    //!
+    //! @brief Destructor.
+    //!
+    inline ~device_strided_batch_vector() noexcept
+    {
```
leekillough: `inline` can be removed to save typing (that's the only reason I leave it out).

clients/include/host_batch_vector.hpp
leekillough: [Rule of 3/5/0](https://en.cppreference.com/w/cpp/language/rule_of_three)

Assignment operator was intentionally deleted in the original implementation. Better to just get a compilation error than to deal with runtime complications. 

We could define it like a "copy" function if we wanted to, so long as it was well-defined and handled different sizes, etc.

clients/include/host_batch_vector.hpp
```diff
@@ -148,7 +188,82 @@ class host_batch_vector
         }
     };
 
+    //!
+    //! @brief Initialize with the rocblas random number generator.
+    //! @param seedReset if true reset the seed.
+    //!
+    inline void random_init(bool seedReset = true) noexcept
```
TorreZuk: Not sure if the memory helpers should introduce dependency on random.   Other random helpers exist so could handle the batch_vectors there?

clients/include/device_batch_vector.hpp
```diff
@@ -49,7 +71,7 @@ class device_batch_vector : private d_vector<T, PAD, U>
     inline ~device_batch_vector() noexcept
     {
         this->free_memory();
-    }
+    };
```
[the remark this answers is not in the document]
leekillough: ... or begin with `inline` if they are defined inside class.

clients/include/host_batch_vector.hpp
leekillough: Yeah, I'm not sure why random number stuff is here.

Also, `noexcept` is mostly redundant nowadays.

clients/include/device_batch_vector.hpp
leekillough: `{}` is pretty explicit. Leaving `{}` out is implicit.

clients/include/device_batch_vector.hpp
leekillough: It depends on what the desired behavior is.

1. If you only want to check bounds during debug builds, use `assert()`.
2. If you want to check bounds with an error exit, use `abort()` or `exit()`, possibly with an error message.
3. If you want to check bounds using exceptions **(not detected by Google Tests)**, use `throw`.
4. (Not applicable here.) If you want to check bounds returning an error code, return a `rocblas_status_invalid_size` or other error where applicable.

clients/include/device_batch_vector.hpp
leekillough: `CHECK_HIP_ERROR()` checks the HIP return type passed as its argument, and if it is not `hipSuccess`, prints an error message, and calls `exit(1)`. I do not see how this is similar to this context, where a different return value is returned if there is a HIP error.

HIP is separate from rocBLAS, and there is no need to compare them, or to try to make them behave similarly.

All rocBLAS APIs should return rocBLAS return values, and by extension, rocBLAS tests should return rocBLAS status codes instead of HIP status codes. If there is no return value, and an exceptional condition needs to be indicated, then it depends on the context what the correct action should be.

Note that Google Tests do not detect C++ exceptions, and `exit()` and `abort()` should only be used for fatal error conditions, not just failing tests.

If a function you're writing can indicate an error in its return value, then it should do so where possible, even if it means returning nullptr.

clients/include/d_vector.hpp
```diff
@@ -0,0 +1,131 @@
...
+private:
+    size_t size, bytes;
+
+public:
+    inline size_t nmemb() const noexcept
+    {
+        return size;
+    }
+
+public:
```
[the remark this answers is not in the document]
leekillough: I don't mind multiple sections, but I don't like `public:` or `private:` being specified when they are already implied.

clients/include/testing_asum.hpp
```diff
@@ -24,25 +25,21 @@ void testing_asum_bad_arg_template(const Arguments& arg)
...
+    EXPECT_ROCBLAS_STATUS((func(nullptr, N, dx, incx, h_rocblas_result)),
                           rocblas_status_invalid_handle);
 }
 
 template <typename T1, typename T2 = T1>
 void testing_asum_template(const Arguments& arg)
 {
-    rocblas_int N    = arg.N;
-    rocblas_int incx = arg.incx;
+    constexpr auto& func = rocblas_asum<T1, T2>;
```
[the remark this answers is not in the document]
leekillough: I don't see the shorthand in this context. 
```
 rocblas_int N    = arg.N;
 rocblas_int incx = arg.incx;
```
Is very different from:
```
constexpr auto& func = rocblas_asum<T1, T2>;
```
I prefer shorthand, when possible.

clients/include/device_strided_batch_vector.hpp
leekillough: The raison d'être of this change, was to support negative strides during tests.

clients/include/d_vector.hpp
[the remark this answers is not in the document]
YvanMokwinski: I'm okay to remove it. It's my own habit and taste to specify privacy for each method, as others language. Yes it is redundant, but "to me" some redundancies increase readability.

clients/include/device_batch_vector.hpp
YvanMokwinski: CHECK_HIP_ERROR is used in the tests to check the validity of hip operations.
It seems natural to return a hip error if a hip operation is used.
I don't see where the problem is. If this is about the specificity of the hipErrorInvalidContext, this is the one I found with the closest meaning.

Is this a rocblas api? No, this is utility classes for testing.
Question: Can we converge to a solution ?

- 1/ return hip error?
- 2/ return rocblas error ? That would mean more work when refactoring of tests (by not using CHECK_HIP_ERROR)

clients/include/device_strided_batch_vector.hpp
YvanMokwinski: absolute values are now used.

clients/include/host_batch_vector.hpp
YvanMokwinski: For now, the behavior is performing the copy if all the parameters are the same.
We can start with that and extend if needed.

copy_from can return a flag to indicate if the copy has been performed successfully or not.
An assignment operator cannot.

clients/include/device_batch_vector.hpp
YvanMokwinski: Originally I want 1.
Can we live with 1?

clients/include/device_batch_vector.hpp
YvanMokwinski: Let's use {} then.

clients/include/device_batch_vector.hpp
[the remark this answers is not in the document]
YvanMokwinski: this is details..

clients/include/testing_asum.hpp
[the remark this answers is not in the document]
YvanMokwinski: You don't see the shorthand because this is a caption of the code, the line is 'constexpr ... func = ...'

clients/include/host_batch_vector.hpp
YvanMokwinski: This is utility classes for tests. Why using random is an issue?..

TorreZuk: Not an issue, but putting all the factory like initializers in a separate file might be a cleaner pattern to follow.   Keeping the memory by itself seems like it is a lot of code already, especially if we add in irregular batch helpers, maybe they will require the random, but could be a separate class/file anyway.

clients/include/device_strided_batch_vector.hpp
YvanMokwinski: detail.

clients/include/testing_asum.hpp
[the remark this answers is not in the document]
leekillough: `constexpr auto& func = rocblas_asum<T1, T2>;` is clear enough to me. As explained in another PR, `auto` can be used inside larger declarations as a placeholder for a type derived the same way as template argument deduction.

PR #757, opened 2019-10-17
Adding batch support to trsv

mahmoodw:
- Added trsv_batched and trsv_strided_batched.
- Changes are very similar to batched versions of trsm

As Daine previously mentioned, there has been some changes to the memory management style so would like to request Lee to review these changes on the trsm PR.


library/src/blas3/rocblas_trsm.cpp
```diff
@@ -2,6 +2,7 @@
  * Copyright 2019 Advanced Micro Devices, Inc.
```
daineAMD: Hmm, I think these changes are just from my trsm PR and don't need to be added in this one, hopefully?

library/src/blas2/rocblas_trsv_strided_batched.cpp
```diff
@@ -0,0 +1,273 @@
...
+                            diag_letter,
+                            "M",
+                            m,
+                            "lda",
+                            lda,
+                            "stride_A",
+                            stride_A,
+                            "incx",
+                            incx,
+                            "stride_B",
```
daineAMD: In our testing code we're using stride_x I believe, rather than stride_B. stride_x makes more sense since it's a vector not a matrix imo.

library/src/blas2/rocblas_trsv.hpp
```diff
@@ -0,0 +1,924 @@
...
+                                       rocblas_int    abs_incx,
+                                       rocblas_int    offset,
+                                       rocblas_stride stride)
+    {
+        ptrdiff_t tx = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+        if(tx < size)
+        {
+            T*   data         = load_ptr_batch(dataa, hipBlockIdx_y, offset, stride);
+            T*   data_end     = data + (m - 1) * abs_incx;
+            auto offset       = tx * abs_incx;
```
daineAMD: Maybe rename this so it doesn't have the same name as the input parameter.

library/src/blas2/rocblas_trsv.hpp
```diff
@@ -0,0 +1,924 @@
...
+                           grid,
+                           threads,
+                           0,
+                           handle->rocblas_stream,
+                           dst,
+                           dst_incx,
+                           src,
+                           src_incx,
+                           size);
+    }
```
daineAMD: Can probably remove these if we aren't using them anymore.

library/src/blas2/rocblas_trsv.hpp
```diff
@@ -0,0 +1,924 @@
...
+
+        // For user-supplied invA, check to make sure size is large enough
+        // If not large enough, indicate degraded performance and ignore supplied invA
+        if(supplied_invA && supplied_invA_size / BLOCK < m)
+        {
+            static int msg = fputs("WARNING: TRSV invA_size argument is too small; invA argument "
+                                   "is being ignored; TRSV performance is degraded\n",
+                                   stderr);
+            perf_status    = rocblas_status_perf_degraded;
+            supplied_invA  = nullptr;
```
daineAMD: I might have missed it, but this check should be in either the _impl functions or the _template as well to set supplied_invA = nullptr for the actual execution aside from the memory management.

library/src/blas3/rocblas_trsm.cpp
mahmoodw: yes but tritri_trsm.hpp is modified to get this working, I will wait for your PR to get in first and then handle the merge. There shouldn't be many conflicts.

clients/gtest/trsv_gtest.yaml
```diff
@@ -4,26 +4,26 @@ include: known_bugs.yaml
...
+    - { M:   10, lda:   10, stride_a: 100 }
+    - { M:   128, lda:   128, stride_a: 32768 }
 
   - &medium_matrix_size_range
-    - { M:   192, lda:   192 }
-    - { M:   256, lda:   256 }
-    - { M:   384, lda:   384 }
-    - { M:   600, lda:   600 }
-    - { M:   800, lda:   801 }
+    - { M:   192, lda:   192, stride_a: 36864 }
```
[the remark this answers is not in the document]
mahmoodw: Not really, it just made the generation script cleaner since I could just add trsv to a previous case. Since it's defined in the same data set as it's requirements we aren't coming up with excessive permutations.

clients/include/testing_trsv_batched.hpp
```diff
@@ -0,0 +1,365 @@
...
+                hA[b][idx] = AAT[b][idx];
+                t += AAT[b][idx] > 0 ? AAT[b][idx] : -AAT[b][idx];
+            }
+            hA[b][i + i * lda] = t;
+        }
+    }
+    //  calculate Cholesky factorization of SPD matrix hA
+    for(int b = 0; b < batch_count; b++)
+        cblas_potrf<T>(char_uplo, M, hA[b], lda);
+
```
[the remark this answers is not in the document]
mahmoodw: yea always nice to simplify.

clients/gtest/trsv_gtest.cpp
```diff
@@ -36,39 +46,78 @@ namespace
...
+            name << rocblas_datatype2string(arg.a_type) << '_' << (char)std::toupper(arg.uplo)
+                 << (char)std::toupper(arg.transA) << (char)std::toupper(arg.diag) << '_' << arg.M
+                 << '_' << arg.lda;
+
+            if(TRSV_TYPE == TRSV_STRIDED_BATCHED)
+                name << '_' << arg.stride_a;
+
+            name << '_' << arg.incx;
+
+            if(TRSV_TYPE != TRSV)
```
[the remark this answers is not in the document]
daineAMD: Looks like CI wants a clang_format as well.

PR #759, opened 2019-10-22
SWDEV-208056-single-template-swap

TorreZuk:
Summary of proposed changes:
* use single template pattern for 3 variants


library/src/blas1/rocblas_swap.hpp
```diff
@@ -13,35 +13,66 @@ __forceinline__ __device__ __host__ void rocblas_swap_vals(T* x, T* y)
...
 {
+    T*        x   = load_ptr_batch(xa, hipBlockIdx_y, offsetx, stridex);
+    T*        y   = load_ptr_batch(ya, hipBlockIdx_y, offsety, stridey);
     ptrdiff_t tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+
     if(tid < n)
     {
         rocblas_swap_vals(x + tid * incx, y + tid * incy);
     }
 }
```
YvanMokwinski: rocblas_swap_kernel:

You can probably skip the template parameter T by using "auto x = load_ptr_batch ..." rather than "T * x = ".
Here, T is doing nothing.

library/src/blas1/rocblas_swap.hpp
```diff
@@ -13,35 +13,66 @@ __forceinline__ __device__ __host__ void rocblas_swap_vals(T* x, T* y)
...
+                       shiftx,
+                       incx,
+                       stridex,
+                       y,
+                       shifty,
+                       incy,
+                       stridey);
 
     return rocblas_status_success;
 }
```
YvanMokwinski: if T in rocblas_swap_kernel is skipped, then T can be skipped here too.

library/src/blas1/rocblas_swap_batched.cpp
```diff
@@ -66,7 +66,7 @@ namespace
         RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);
 
         constexpr rocblas_int NB = 256;
-        return rocblas_swap_batched_template<NB>(handle, n, x, 0, incx, y, 0, incy, batch_count);
+        return rocblas_swap_template<NB, T>(handle, n, x, 0, incx, 0, y, 0, incy, 0, batch_count);
     }
```
YvanMokwinski: I like naming constant values: 

static constexpr ptrdiff_t zero_offset =0;
rocblas_swap_template<NB, T>(handle, n, x, zero_offset , incx, zero_offset , y, zero_offset , incy, zero_offset , batch_count)

library/src/blas1/rocblas_swap.cpp
```diff
@@ -48,7 +48,7 @@ namespace
 
         RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);
 
-        return rocblas_swap_template<NB>(handle, n, x, incx, y, incy);
+        return rocblas_swap_template<NB, T>(handle, n, x, 0, incx, 0, y, 0, incy, 0, 1);
     }
```
YvanMokwinski: I like naming constant values:

static constexpr ptrdiff_t zero_offset =0;
rocblas_swap_template<NB, T>(handle, n, x, zero_offset , incx, zero_offset , y, zero_offset , incy, zero_offset , 1)

library/src/blas1/rocblas_swap.hpp
TorreZuk: Changed it but now I think I preferred the original.   Was simpler to see it is computing a pointer, auto here is more like a trick than helping avoid future changes.   Overuse of auto may make maintenance worse rather than better.

library/src/blas1/rocblas_swap.cpp
TorreZuk: I had some of those but to be consistent with other functions I removed them.   Intellisense can show you the parameter names so that helps a bit too.

library/src/blas1/rocblas_swap.hpp
YvanMokwinski: Even if written like this ?
    ptrdiff_t tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
    if(tid < n)
    {
        rocblas_swap_vals(load_ptr_batch(xa, hipBlockIdx_y, offsetx + tid * incx, stridex), 
                                      load_ptr_batch(ya, hipBlockIdx_y, offsety +  tid * incy, stridey) );
    }

TorreZuk: Close but that makes the function call look too crowded  :-)     I'll leave it with auto unless someone else spots an issue.

library/src/blas1/rocblas_swap.hpp
```diff
@@ -13,35 +13,66 @@ __forceinline__ __device__ __host__ void rocblas_swap_vals(T* x, T* y)
     *x    = tmp;
 }
 
-template <typename T>
-__global__ void rocblas_swap_kernel(rocblas_int n, T* x, rocblas_int incx, T* y, rocblas_int incy)
+template <typename U>
```
daineAMD: Is there a reason this is `U` instead of `T`? Not that it matters haha

library/src/blas1/rocblas_swap.cpp
leekillough: I prefer `batch_count` to have a default value of `1` in the function header, rather than passing 1 explicitly in the call.

`rocblas_swap_template<NB,T>(...)` can be simplified to `rocblas_swap_template<NB>(...)`. That's the reason for putting type template parameters at the end.

library/src/blas1/rocblas_swap.hpp
leekillough: > 
> 
> Changed it but now I think I preferred the original. Was simpler to see it is computing a pointer, auto here is more like a trick than helping avoid future changes. Overuse of auto may make maintenance worse rather than better.

You can document and force it to be a pointer:

```
auto* X = load_ptr_batch...
```

leekillough: `auto` is merely a placeholder; template argument deduction rules do the work. So if:
```
template <typename T>
void foo( T**** X);
````
will work, then
```
auto**** X = ...
```
will work similarly.

Comment (leekillough) [reply library/src/blas1/rocblas_swap.hpp]: > 
> 
> rocblas_swap_kernel:
> 
> You can probably skip the template parameter T by using "auto x = load_ptr_batch ..." rather than "T * x = ".

auto* x = load_ptr_batch

File: library/src/blas1/rocblas_swap.hpp [hunk 6]
```diff
@@ -13,35 +13,66 @@ __forceinline__ __device__ __host__ void rocblas_swap_vals(T* x, T* y)
...
+                                    ptrdiff_t      offsetx,
+                                    rocblas_int    incx,
+                                    rocblas_stride stridex,
+                                    U              ya,
+                                    ptrdiff_t      offsety,
+                                    rocblas_int    incy,
+                                    rocblas_stride stridey)
 {
+    auto      x   = load_ptr_batch(xa, hipBlockIdx_y, offsetx, stridex);
+    auto      y   = load_ptr_batch(ya, hipBlockIdx_y, offsety, stridey);
```
leekillough:
```
    auto*     x    = load_ptr_batch(xa, hipBlockIdx_y, offsetx, stridex);
    auto*     y    = load_ptr_batch(ya, hipBlockIdx_y, offsety, stridey);
```

library/src/blas1/rocblas_swap.cpp
leekillough: https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/b4a317d2f3777cd656770a85fad22745d5d10618/library/src/blas1/rocblas_swap.hpp#L48

```
         rocblas_int    batch_count = 1)
```
This documents the default `batch_count` pretty clearly without having to use named constants like `one_batch`, and it avoids passing `1` at the end of a call.

library/src/blas1/rocblas_swap.hpp
TorreZuk: Well the original reason was 2 template parameters, probably should have a convention that T is a base type and anything more complex has some hints in the name, as I had used XPtr for another case.  This style I copied from another existing merged template but as you can see using the pre-existing pattern generated a lot of comments.

library/src/blas1/rocblas_swap.hpp
TorreZuk: Good point Lee, that is better.

PR #760, opened 2019-10-23
fix clang formatting

TorreZuk:
* builds were broken and I saw this clang format diff


TorreZuk:
Format check passes so merged

PR #762, opened 2019-10-23
Use near check for all rot* tests, fixes test run when built with hipcc

AlexBrownAMD:


clients/include/testing_rotg_strided_batched.hpp
```diff
@@ -147,7 +148,6 @@ void testing_rotg_strided_batched(const Arguments& arg)
 
             if(arg.unit_check)
             {
-                double rel_error = std::numeric_limits<U>::epsilon() * 100;
```
daineAMD: Looks like the scalar here was 100 before, this error wasn't enough for hipcc or is it just changed for consistency?

PR #763, opened 2019-10-23
Add V340L as a supported device

amdkila:
V340L's deviceID has an alias of Vega, so I've added Vega to all the vega 10 yaml files


amdkila:
Oh yes, need to add that, thanks for the feedback. I've added the device number to all the files now- let me know if everything looks good

amdkila:
Will do, thanks!

PR #764, opened 2019-10-25
Changed packages that are installed with running with the -d flag

amdkila:
- Removes gtest and lapack as library dependency
- Removes libgomp as a library dependency
- Removes blis as a library dependency 

All are now client dependencies

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -136,7 +136,7 @@ install_packages( )
   local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "pkg-config"
                                       "python2.7" "python3" "python-yaml" "python3-yaml"
                                       "llvm-6.0-dev" "hip_hcc" "rocm_smi64" "zlib1g-dev"
-				      "libboost-program-options-dev" )
+				                      "libboost-program-options-dev" )
   local library_dependencies_centos=( "epel-release" "boost-devel"
                                       "make" "cmake3" "rpm-build"
                                       "python34" "PyYAML" "python3*-PyYAML"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -364,8 +362,10 @@ esac
 # dependencies
 # #################################################
 if [[ "${install_dependencies}" == true ]]; then
-
   install_packages
+fi
+
+if [[ "${build_clients}" == true ]]; then
```
TorreZuk: Would only want this on install_dependencies case no?

install.sh
```diff
@@ -135,17 +135,15 @@ install_packages( )
   # dependencies needed for rocblas and clients to build
```
leekillough: Comment is misleading: Does this part of the script add dependencies for `rocblas` _and_ `clients` to build? If so, why is `libgomp` being removed? Or is this part of the script only for `rocblas`, and not for `clients`?

zaliu (changes requested):
Can we please pull rocm/rocm-terminal:2.9 and build rocBLAS with this PR and `install.sh -d` to see if it succeeds?  I just tried it and it failed when building TensileHost for a missing header <boost/algorithm/string/split.hpp>.  Thanks.
Moving the libboost-program-options-dev package from client_dependencies to library_dependencies made `install.sh -d` work - at least on Ubuntu.

install.sh
amdkila: This is the part of the script for building without the client. I have fixed this in my latest commit

amdkila:
Yeah, I added boost as a client dependency because I thought CI would fail without it as we only build the client and then define the dependencies in the dockerfile. However, it turns out boost was already defined in the dockerfile (which is where it should be defined for CI, not added as a client dependency). I removed that commit, and it has passed CI now (only failure was an I/O error on Ubuntu), so this should be safe to merge

leekillough:
rocblas client, but not the library, does depend on Boost.

It is used to parse command-line arguments for rocblas-test and rocblas-bench.

amdkila:
It isn't. I removed it in my last commit: https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/764/commits/0abd28a054e6c080beac8d241324e44e257b4696

amcamd (approved):
LGTM

amdkila:
Looks like @zaliu has to remove his request for changes before I can merge the PR

zaliu (approved):

PR #766, opened 2019-10-30
trmm, gemm based

amcamd:
Add gemm based trmm

- only for precisions float and double
- no batched or strided_batched
- code is functional, algorithm will need to be modified to improve performance
- Inside of trmm there are calls to
  - rocblas_set_pointer_mode
  - rocblas_scal_template
  - rocblas_copy_template
  - rocblas_gemv_template
  - rocblas_gemm_template

  if any of the internal calls return status other than rocblas_status_succcess, then
  the trmm will return rocblas_status_internal_error. In addition the macro
  CHECK_INTERNAL_ROCBLAS_ERROR will print to stderr
  - error string
  - error number
  - filename
  - line number

  For the definition of the macro, see line 254 of library/src/include/utility.h.
- Device memory is allocated for work arrays dt1[RB,CB] and dt2[CB,CB]. CB and RB are
  - constexpr rocblas_int RB = 128;
  - constexpr rocblas_int CB = 128;

  See line 140 of rocblas_trmm.cpp for memory device memory allocation.


clients/include/flops.hpp
```diff
@@ -120,6 +120,48 @@ constexpr double geam_gflop_count(rocblas_int m, rocblas_int n)
...
+    }
+}
+
+template <>
+constexpr double
+    trmm_gflop_count<rocblas_float_complex>(rocblas_int m, rocblas_int n, rocblas_side side)
+{
+    if(rocblas_side_left == side)
+    {
+        return (1.0 * m * n * (m + 1)) / 1e9;
```
daineAMD: I'm thinking the complex flops should be 4x the real ones if I'm not mistaken. Doesn't really matter since they're not implemented yet.

Come to think of it, I likely missed complex calcs for other functions (I'm thinking dot (4x), scal(2x), axpy(2x)?).

library/src/include/utility.h
```diff
@@ -233,6 +233,40 @@ constexpr size_t rocblas_sizeof_datatype(rocblas_datatype type)
...
+    case rocblas_status_invalid_pointer: return "rocblas_status_invalid_pointer";
+    case rocblas_status_invalid_size: return "rocblas_status_invalid_size";
+    case rocblas_status_memory_error: return "rocblas_status_memory_error";
+    case rocblas_status_internal_error: return "rocblas_status_internal_error";
+    case rocblas_status_perf_degraded: return "rocblas_status_perf_degraded";
+    case rocblas_status_size_query_mismatch: return "rocblas_status_size_query_mismatch";
+    case rocblas_status_size_increased: return "rocblas_status_size_increased";
+    case rocblas_status_size_unchanged: return "rocblas_status_size_unchanged";
+    default: return "rocblas_status_internal_error";
+    };
```
TorreZuk: Should be part of the "C" API as I needed this in examples too, hip has the equivalent in the API.   Note it exists in rocblas_test.hpp as inline const char* rocblas_status_to_string(rocblas_status status)

clients/include/testing_trmm.hpp
```diff
@@ -27,138 +28,177 @@ void testing_trmm(const Arguments& arg)
...
-    /* =====================================================================
-           ROCBLAS
-    =================================================================== */
-    if(arg.timing)
+    if(arg.unit_check || arg.norm_check)
     {
-        gpu_time_used = get_time_us(); // in microseconds
-    }
+        // calculate dB <- A^(-1) B   rocblas_device_pointer_host
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
```
daineAMD: Looks like we're only doing host_mode tests and not any device_mode tests, is that intended?

clients/include/testing_trmm.hpp
```diff
@@ -27,138 +28,177 @@ void testing_trmm(const Arguments& arg)
...
-        if(arg.norm_check)
+        gpu_time_used = get_time_us(); // in microseconds
+        for(int i = 0; i < number_hot_calls; i++)
         {
-            cout << cblas_gflops << "(" << cpu_time_used << "),";
-            cout << rocblas_error;
+            rocblas_trmm<T>(handle, side, uplo, transA, diag, M, N, &alpha_h, dA, lda, dB, ldb);
         }
+        gpu_time_used  = get_time_us() - gpu_time_used;
+        rocblas_gflops = gemm_gflop_count<T>(M, N, side) * number_hot_calls / gpu_time_used * 1e6;
```
daineAMD: Should this be trmm_gflop_count?

clients/include/flops.hpp
TorreZuk: I don't think I updated flops to be accurate either.   complex mult is 4 fp mult + 2 fp add.   I thought these were only O(?) approximations but if we want to plot them we probably should update them all to be more accurate in a consistent way.

clients/gtest/CMakeLists.txt
```diff
@@ -180,7 +181,7 @@ endif( )
 set( ROCBLAS_TEST_DATA "${PROJECT_BINARY_DIR}/staging/rocblas_gtest.data")
 add_custom_command( OUTPUT "${ROCBLAS_TEST_DATA}"
                     COMMAND ../common/rocblas_gentest.py -I ../include rocblas_gtest.yaml -o "${ROCBLAS_TEST_DATA}"
-                    DEPENDS ../common/rocblas_gentest.py rocblas_gtest.yaml ../include/rocblas_common.yaml known_bugs.yaml blas1_gtest.yaml gemm_gtest.yaml gemm_batched_gtest.yaml gemm_strided_batched_gtest.yaml gemv_gtest.yaml symv_gtest.yaml syr_gtest.yaml ger_gtest.yaml trsm_gtest.yaml trtri_gtest.yaml geam_gtest.yaml set_get_vector_gtest.yaml set_get_matrix_gtest.yaml trsv_gtest.yaml logging_mode_gtest.yaml set_get_pointer_mode_gtest.yaml
+                    DEPENDS ../common/rocblas_gentest.py rocblas_gtest.yaml ../include/rocblas_common.yaml known_bugs.yaml blas1_gtest.yaml gemm_gtest.yaml gemm_batched_gtest.yaml gemm_strided_batched_gtest.yaml gemv_gtest.yaml symv_gtest.yaml syr_gtest.yaml ger_gtest.yaml trsm_gtest.yaml trtri_gtest.yaml geam_gtest.yaml set_get_vector_gtest.yaml set_get_matrix_gtest.yaml trsm_gtest.yaml trsv_gtest.yaml logging_mode_gtest.yaml set_get_pointer_mode_gtest.yaml
```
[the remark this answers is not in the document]
amcamd: Good catch, thanks, corrected

clients/include/flops.hpp
amcamd: Thanks, daineAMD is correct, trmm complex flops is 4x real flops. For most of BLAS flops is from (multiply_add), and complex (multiply_add) has 4 additions and 4 multiplications, so 4x the flops compared to read (multiply_add). I need to check, but I think it is dot(4x), cscal(6x), daxpy(4x).

clients/include/testing_trmm.hpp
amcamd: Thanks for catching this, fixed.

clients/include/flops.hpp
TorreZuk: I thought FLOPS is not an instruction count but operation count.  Malcolm has suggested not reporting FLOPS and only report time to avoid these issues.  Can discuss at meeting.

library/src/blas3/dcld.hpp
```diff
@@ -0,0 +1,119 @@
+bool dcld(rocblas_int ld)
```
leekillough: Should be `inline` or `constexpr`.

library/src/blas3/rocblas_trmm.cpp
```diff
@@ -1,395 +1,160 @@
...
+                                     rocblas_int       lda,
+                                     T*                c,
+                                     rocblas_int       ldc)
     {
         if(!handle)
             return rocblas_status_invalid_handle;
-        if(!alpha)
-            return rocblas_status_invalid_pointer;
 
         auto layer_mode = handle->layer_mode;
```
leekillough: We disable logging when `handle->is_device_memory_size_query()`.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+#include "../blas1/rocblas_scal.hpp"
+#include "../blas2/rocblas_gemv.hpp"
+#include "../blas3/Tensile/gemm.hpp"
+#include "dcld.hpp"
+#include "handle.h"
+#include "rocblas.h"
+#include "utility.h"
+
+namespace
+{
```
leekillough: Nnnamed namespaces should not appear in header files.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                        //
+                        //                  T2 := A', the transpose of a upper unit or non-unit
+                        //                  triangular diagonal block of A is copied to the
+                        //                  lower triangular part of T2.
+                        //
+                        for(int i = ii + offd; i <= ii + isec - 1; i++)
+                        {
+                            CHECK_INTERNAL_ROCBLAS_ERROR(
+                                (rocblas_copy_template<NB, T>(handle,
+                                                              i - ii + 1 - offd,
```
leekillough: Parentheses are only needed around `rocblas_copy_template<NB, T>`, not the whole call. Makes `clang-format` work better sometimes.

Also, `T` does not need be passed as a template argument, since it can be automatically deduced; hence, `<NB, T>` can be replaced with `<NB>`, and there is no need for extra parentheses.


`CHECK_INTERNAL_ROCBLAS_ERROR()` is a new macro?  We already have `RETURN_IF_ROCBLAS_ERROR()`.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                            //                      T1 := C', the transpose of a rectangular block
+                            //                      of C is copied to T1.
+                            //
+                            if(cldc)
+                            {
+                                for(int j = jj; j <= jj + jsec - 1; j++)
+                                {
+                                    CHECK_INTERNAL_ROCBLAS_ERROR(
+                                        (rocblas_copy_template<NB, T>(handle,
+                                                                      isec,
```
leekillough: Same as above

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                            {
+                                if(diag == rocblas_diagonal_non_unit)
+                                {
+                                    delta                  = *alpha;
+                                    bool pointer_mode_host = false;
+                                    if(handle->pointer_mode == rocblas_pointer_mode_host)
+                                    {
+                                        pointer_mode_host = true;
+                                        CHECK_INTERNAL_ROCBLAS_ERROR((rocblas_set_pointer_mode(
+                                            handle, rocblas_pointer_mode_device)));
```
leekillough: We should call `handle->push_pointer_mode()` instead of `rocblas_set_pointer_mode` if possible. It changes the current mode and restores the object returned by `push_pointer_mode()` is destroyed. It is automatically exception-safe.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                                1)));
+                        }
+                    }
+                }
+                else
+                {
+                    //
+                    //             Form  C := alpha*A'*C. Left, Upper, Transpose.
+                    //
+                    delta     = *alpha;
```
leekillough: Is `alpha` guaranteed on the host? 

The usual way to handle scalars is explained as #3 [on this page](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.github/CONTRIBUTING.md)

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                                                                      0,
+                                                                      1,
+                                                                      0,
+                                                                      1)));
+                                }
+                                else
+                                {
+                                    CHECK_INTERNAL_ROCBLAS_ERROR(
+                                        (rocblas_gemv_template<T>(handle,
+                                                                  rocblas_operation_none,
```
leekillough: `<T>` should not be necessary.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1488 @@
...
+                                            handle, rocblas_pointer_mode_host)));
+                                    }
+                                }
+                                tsec = ii + isec - 1 - i;
+                                tsec = i - ii;
+                                if(0 == tsec)
+                                {
+                                    CHECK_INTERNAL_ROCBLAS_ERROR(
+                                        (rocblas_scal_template<NB, T>(handle,
+                                                                      jsec,
```
leekillough: `<NB, T>` should be able to be replaced with `<NB>`, and then the parentheses around the call are unnecessary.

library/src/include/utility.h
leekillough: See PR #772

library/src/include/utility.h
```diff
@@ -233,6 +233,40 @@ constexpr size_t rocblas_sizeof_datatype(rocblas_datatype type)
...
+                    rocblas_get_status_string(error),       \
+                    error,                                  \
+                    __FILE__,                               \
+                    __LINE__);                              \
+            return rocblas_status_internal_error;           \
+        }                                                   \
+    } while(0)
+#endif
 
 // return rocblas_datatype from type
```
leekillough: We already have `RETURN_IF_ROCBLAS_ERROR()` and `PRINT_IF_ROCBLAS_ERROR()`.

`PRINT_IF_ROCBLAS_ERROR()` is more descriptive than `CHECK_INTERNAL_ROCBLAS_ERROR()`, since we have `rocblas_status_internal_error` and the macro name suggests it only checks for `rocblas_status_internal_error`.

Macro definitions should go into `definitions.h` instead of `utility.h`.

library/src/include/utility.h
amcamd: Change to using macro PRINT_AND_RETURN_IF_ROCBLAS_ERROR from src/include/definitions.h , and it uses new external API function rocblas_status_to_string.

clients/include/flops.hpp
amcamd: Since at least 28-August-1989 with file sblas2time.f from http://www.netlib.org/blas/old-index.html BLAS has used flops/sec to measure BLAS performance. GEMM usually gets close to peak performance. It may easier to understand a flops/sec plot with the peak performance plotted as a line rather than a time result.

clients/include/testing_trmm.hpp
amcamd: Thanks for catching this, added back device_mode tests.

library/src/blas3/dcld.hpp
amcamd: Thanks, changed to inline.

library/src/blas3/rocblas_trmm.hpp
amcamd: Removed unnamed namespace.

library/src/blas3/rocblas_trmm.hpp
amcamd: Changed to use handle->push_pointer_mode()

library/src/blas3/rocblas_trmm.hpp
amcamd: Removed use of delta.

library/src/include/utility.h
amcamd: Added macro PRINT_AND_RETURN_IF_ROCBLAS_ERROR .

library/src/blas3/rocblas_trmm.hpp
amcamd: Changed to parenthesis around rocblas_copy_template<NB, T> . The T is needed, it cannot be deduced.

amcamd: Removed CHECK_INTERNAL_ROCBLAS_ERROR macro. Added macro PRINT_AND_RETURN_IF_ROCBLAS_ERROR.

library/src/blas3/rocblas_trmm.hpp
amcamd: T is required, moved parenthesis to function name and template parameters

library/src/blas3/rocblas_trmm.hpp
amcamd: T is required, cannot be deduced.

library/src/blas3/rocblas_trmm.hpp
amcamd: T is required, cannot be deduced, moved parenthesis to be around function name and templates.

library/src/blas3/rocblas_trmm.cpp
amcamd: Disabled logging when is_device_memory_size_query

clients/include/testing_trmm.hpp
```diff
@@ -27,138 +28,177 @@ void testing_trmm(const Arguments& arg)
...
+        std::cout << "M,N,alpha,lda,ldb,side,uplo,transA,diag,rocblas-Gflops,us";
+
+        if(arg.unit_check || arg.norm_check)
+            std::cout << ",CPU-Gflops,us,norm-error";
+
+        std::cout << std::endl;
+
+        std::cout << M << ',' << N << ',' << arg.get_alpha<T>() << ',' << lda << ',' << ldb << ','
+                  << char_side << ',' << char_uplo << ',' << char_transA << ',' << char_diag << ','
+                  << rocblas_gflops << "," << gpu_time_used;
```
mahmoodw: You must divide gpu_time_used by number of hot calls for accurate logging.

amcamd: Thanks for catching this, fixed in commit 80858ac

clients/gtest/trmm_gtest.cpp
```diff
@@ -0,0 +1,75 @@
...
+    {
+    };
+
+    // When the condition in the second argument is satisfied, the type combination
+    // is valid. When the condition is false, this specialization does not apply.
+    template <typename T>
+    struct trmm_testing<
+        T,
+        typename std::enable_if<std::is_same<T, float>{} || std::is_same<T, double>{}>::type>
+    {
```
leekillough: The `trmm_testing` class needs to derive from `rocblas_test_valid` for the valid type combinations, and `rocblas_test_invalid` for the invalid type combinations.

clients/gtest/trmm_gtest.cpp
```diff
@@ -0,0 +1,75 @@
...
+        explicit operator bool()
+        {
+            return true;
+        }
```
leekillough: We no longer need to define `explicit operator bool()` in test classes, since it is derived from both `rocblas_status_valid` and `rocblas_status_invalid`.

library/src/blas3/rocblas_trmm.hpp
```diff
@@ -0,0 +1,1453 @@
...
+                    }
+                }
+            }
+            else
+            {
+                //
+                //             Form  C := alpha*A'*C. Left, Upper, Transpose.
+                //
+                bool cldc = dcld(ldc);
+                for(int ii = m - ((m - 1) % cb); ii >= 1; ii -= cb)
```
leekillough: It's not important now, but for consistency we should use `rocblas_int` for loop control variables limited by `rocblas_int` variables.

If the loop limit is calculated with multiplication, and thus can exceed 32-bits, at least one operand should be promoted with `size_t(n)` and the product and any loops which iterate over it should be `size_t`.

clients/gtest/trmm_gtest.cpp
amcamd: fixed in commit 560bdd2

PR #767, opened 2019-10-30
introduce rocblas_reduction_template and rocblas_reduction_impl

YvanMokwinski:
and generalize its usage.

Summary of proposed changes:

============================================

Generalize the 'template' and the 'impl' to be used by nrm2/asum/amax/amin

- 3/ NEW: library/src/blas1/rocblas_reduction_template.hpp
- 4/ NEW: library/src/blas1/rocblas_reduction_impl.hpp

============================================

- 5/ MODIFIED library/src/blas1/rocblas_asum.cpp
- 6/ MODIFIED library/src/blas1/rocblas_asum.hpp

- 7/ MODIFIED library/src/blas1/rocblas_asum_batched.cpp
- 8/ MODIFIED library/src/blas1/rocblas_asum_batched.hpp

- 9/ MODIFIED library/src/blas1/rocblas_asum_strided_batched.cpp
- 10/ MODIFIED library/src/blas1/rocblas_asum_strided_batched.hpp

============================================

- 11/ MODIFIED library/src/blas1/rocblas_nrm2.cpp
- 12/ MODIFIED library/src/blas1/rocblas_nrm2.hpp

- 13/ MODIFIED library/src/blas1/rocblas_nrm2_batched.cpp
- 14/ MODIFIED library/src/blas1/rocblas_nrm2_batched.hpp

- 15/ MODIFIED library/src/blas1/rocblas_nrm2_strided_batched.cpp
- 16/ MODIFIED library/src/blas1/rocblas_nrm2_strided_batched.hpp

============================================

- 1/ MODIFIED library/src/blas1/reduction_strided_batched.h
- 2/ MODIFIED library/src/blas1/rocblas_amax_amin.h

### DIFF
--- a/library/src/blas1/rocblas_amax_amin.h
+++ b/library/src/blas1/rocblas_amax_amin.h
@@ -1,12 +1,8 @@
 /* ************************************************************************
  * Copyright 2018-2019 Advanced Micro Devices, Inc.
  * ************************************************************************ */
-#include "fetch_template.h"
-#include "handle.h"
-#include "logging.h"
-#include "reduction.h"
-#include "rocblas.h"
-#include "utility.h"
+
+#include "rocblas_reduction_impl.hpp"
 
 #ifndef MAX_MIN
 #define MAX_MIN max
@@ -33,7 +29,7 @@ struct index_value_t
 
 // Specialization of default_value for index_value_t<T>
 template <typename T>
-struct default_value<index_value_t<T>>
+struct rocblas_default_value<index_value_t<T>>
 {
     __forceinline__ __host__ __device__ constexpr auto operator()() const
     {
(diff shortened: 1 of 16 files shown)


library/src/blas1/rocblas_asum.hpp
```diff
@@ -28,18 +25,14 @@ rocblas_status rocblas_asum_template(rocblas_handle handle,
                                      To*            workspace,
                                      To*            result)
 {
```
daineAMD: We aren't calling these templates anymore are we? Are we planning on keeping them around only for "internal API" purposes?

library/src/blas1/rocblas_asum_batched.hpp
```diff
@@ -19,23 +16,12 @@ rocblas_status rocblas_asum_batched_template(rocblas_handle handle,
...
-            {
-                results[i] = 0;
-            }
-        }
-        return rocblas_status_success;
-    }
-
-    return rocblas_reduction_strided_batched_kernel<NB, rocblas_fetch_asum<To>>(
-        handle, n, x, shiftx, incx, 0, batch_count, workspace, results);
+    static constexpr bool           isbatched = true;
```
daineAMD: Same question about the templates here.

library/src/blas1/rocblas_reduction_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+              name,
+              "-r",
+              rocblas_precision_string<Ti>,
+              "-n",
+              n,
+              "--incx",
+              incx,
+              "--batch",
+              batch_count);
+}
```
daineAMD: Alternatively, we could add _another_ template parameter to differentiate between batched & strided_batched.

library/src/blas1/rocblas_reduction_impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+void reduction_log_profile(rocblas_handle   handle,
+                           rocblas_int      n,
+                           const Ti* const* x,
+                           rocblas_int      incx,
+                           rocblas_stride   stridex,
+                           rocblas_int      batch_count,
+                           const char*      name)
+{
+    log_profile(handle, name, "N", n, "incx", incx, "batch", batch_count);
+}
```
daineAMD: Same goes for here.

library/src/blas1/rocblas_asum.hpp
YvanMokwinski: Good point. 
We could keep for internal purpose :) What do you think?

daineAMD: I would vote "yes" I think. We don't need them for anything yet afaik, but we might in the future, and its good for consistency. We could potentially just use the `reduction_template` if we ever need to, but I think that just introduces some confusion. I could be easily swayed, though, since removing them would get rid of a decent chunk of code.

YvanMokwinski: I tried to stay consistent with the C-API === IMPL === TEMPLATE

daineAMD: Yeah, sounds good to me. In that case we don't need the batched and strided_batched templates, and the regular templates, `rocblas_asum_template` for example, should be generic enough to handle batched/ strided_batched cases.

YvanMokwinski: that's a possibility ! :) let's see what other REVIEWERS! have to say :p

library/src/blas1/rocblas_amax_amin.h
```diff
@@ -176,34 +142,25 @@ static rocblas_status rocblas_iamaxmin(
...
-}
+#define IMPL(name_, typei_, typew_)                                   \
+    rocblas_status name_(rocblas_handle handle,                       \
+                         rocblas_int    n,                            \
+                         const typei_*  x,                            \
+                         rocblas_int    incx,                         \
+                         rocblas_int*   results)                      \
+    {                                                                 \
+        return rocblas_iamaxmin<typew_>(handle, n, x, incx, results); \
+    }
```
TorreZuk: Opinion: I prefer to see more code rather than less code if it is more straight forward, i.e. duplicated code that I can read I think is more maintainable than macros I can not.

library/src/blas1/rocblas_asum.hpp
TorreZuk: If we want to export compiled templates then there is some value in making the explicit templates that have a clear mapping to functionality.

PR #768, opened 2019-10-30
Split MI25 and V340L in Jenkinsfile

eidenyoshida:
To check V340L functionality

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ rocBLASCI:
     rocblas.paths.build_command = './install.sh -lasm_ci -c'
 
     // Define test architectures, optional rocm version argument is available
-    def nodes = new dockerNodes(['ubuntu && gfx900', 'centos7 && gfx900', 'centos7 && gfx906', 'sles && gfx906'], rocblas)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -32,7 +32,7 @@ rocBLASCI:
     rocblas.paths.build_command = './install.sh -lasm_ci -c'
 
     // Define test architectures, optional rocm version argument is available
-    def nodes = new dockerNodes(['ubuntu && gfx900', 'centos7 && gfx900', 'centos7 && gfx906', 'sles && gfx906'], rocblas)
+    def nodes = new dockerNodes(['mi25 && ubuntu && gfx900', 'mi25 && centos7 && gfx900', 'centos7 && gfx906', 'sles && gfx906', 'v340l && ubuntu'], rocblas)
```
amdkila: I could be wrong about this (please correct me if I am @saadrahim) but shouldn't we remove the gfx900 tag?

saadrahim: No, the gfx900 can stay in case projects do not care about the specific GPU type. That is the current behaviour in Jenkins. Not sure what our long term approach to the testing will be.

saadrahim: Maybe I am wrong. We can discuss in person.

amdkila: That makes sense. It should be fine for this PR as is

PR #771, opened 2019-10-31
Adding complex flops calcs. and updating other flop calcs.

daineAMD:
Correcting complex flops calculations for L1 functions that do flops calculations:
- caxpy/zaxpy
- cdotu/zdotu & cdotc/zdotc
- scscal/dzscal & cscal/zscal

We should probably add flops calculations to all of the other functions some time in the future, and add them to logging.

### DIFF
--- a/clients/include/flops.hpp
+++ b/clients/include/flops.hpp
@@ -10,7 +10,8 @@
 
 /*!\file
  * \brief provides Floating point counts of Basic Linear Algebra Subprograms (BLAS) of Level 1, 2,
- * 3.
+ * 3. Where possible we are using the values of NOP from the legacy BLAS files [sdcz]blas[23]time.f
+ * for flop count.
  */
 
 /*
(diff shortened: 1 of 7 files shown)


clients/include/flops.hpp
```diff
@@ -23,16 +23,68 @@ constexpr double axpy_gflop_count(rocblas_int n)
...
+template <>
+constexpr double axpy_gflop_count<rocblas_double_complex>(rocblas_int n)
+{
+    return (8.0 * n) / 1e9;
+}
+
+template <bool CONJ, typename T>
 constexpr double dot_gflop_count(rocblas_int n)
 {
     return (2.0 * n) / 1e9;
```
daineAMD: Technically this should be `n + (n - 1)` since there only has to be `n - 1` adds, right? Although I'm guessing this is nitpicking and we don't care haha.

mahmoodw: Agreed... better to be more accurate though I suppose?

mahmoodw:
LGTM - would like to see what other people think. There is some urgency to getting this in since it is used for performance recording.

clients/include/flops.hpp
leekillough: Most algorithms initialize `sum =  0` and add to it `n` times, which can handle the case when `n==0`. So it's `n` adds, not `n-1`.

PR #772, opened 2019-11-01
Add rocblas_status_to_string() to public API, fix compiler warnings, rename _ex functions properly

leekillough:
Changes:
- Add `rocblas_status_to_string()` function to the public C API. Move it from `clients/include/rocblas_test.hpp` to `library/src/rocblas_auxiliary.cpp`, and make it completely handle every possible value of `rocblas_status`.
- Rename `*_ex_batched` and `*_ex_strided_batched` functions and test names to `*_batched_ex` and `*_strided_batched_ex`. **`_ex` must appear last, and not in the middle, of names.**
- Do not use `default:` in `switch` statements which are supposed to list all values of an `enum`, so that the compiler will make sure that all `enum` cases are handled in the `switch`. If you use `default:`, then a subset of `enum` values can be listed with no warning from the compiler, because the unlisted values will be considered by the compiler to be handled by the `default:` case. In case the `enum` value is invalid and not one of the defined cases, we fall-through the `switch` statement to take the default action. Enabling these warnings by removing `default:` found one case where we did not list all of the possible `enum` values.
- Fix [SWDEV-205156](http://ontrack-internal.amd.com/browse/SWDEV-205156) by removing duplicate declarations in `rocblas-functions.h`. There were some multiple declarations in `rocblas-functions.h`. It looks like someone copied-and-pasted the declarations and Doxygen code for some functions like `rocblas_gemm_strided_batched_ex`, to start writing new functions, but they did not make any changes to the names or declarations, causing duplicate declarations _after_ the macro definitions of the same name(s). This duplication of function declarations after the same-named macro definition, caused the warning in [SWDEV-205156](http://ontrack-internal.amd.com/browse/SWDEV-205156) and in `examples` builds.
- Improve the `pre-commit` script to remove trailing whitespace at the ends of lines, and add missing newlines at ends of files, for `*.txt` and `*.yaml` files. Run the formatting on the current code base. **Note:** One thing this change does, is automatically convert MSDOS CR-LF formatted files, into Unix files with only LF at the ends of lines. For those files which have large changes which appear to be the whole file, those were just converted from CR-LF to LF formatting.
- Replace `trans_b` with `transB` in `rocblas-functions.h`. We were already using `transA`, so `transA` followed by `trans_b` looks oddly asymmetric.


leekillough:
To see the important changes when viewing changes, choose "Settings" and  "Hide whitespace changes". Most changed files are only whitespace changes from the new `pre-commit` script.

library/src/rocblas_auxiliary.cpp
```diff
@@ -755,3 +755,29 @@ catch(...) // catch all exceptions
...
+        CASE(rocblas_status_perf_degraded);
+        CASE(rocblas_status_size_query_mismatch);
+        CASE(rocblas_status_size_increased);
+        CASE(rocblas_status_size_unchanged);
+    }
+#undef CASE
+    // We don't use default: so that the compiler warns us if any valid enums are missing
+    // from our switch. If the value is not a valid rocblas_status, we return this string.
+    return "<undefined rocblas_status value>";
+}
```
amcamd: Should this be part of the external rocBLAS API and added to rocBLAS/library/include/rocblas_auxiliary.h ? something like:

ROCBLAS_EXPORT const char* rocblas_status_to_string(rocblas_status status);

TorreZuk: Yes I want to call it externally.

leekillough: > Should this be part of the external rocBLAS API and added to rocBLAS/library/include/rocblas_auxiliary.h ? something like:
> 
> ROCBLAS_EXPORT const char* rocblas_status_to_string(rocblas_status status);

It is in rocblas-functions.h, which is where exported functions are declared. Rocblas does not declare exported functions in other headers.

amcamd: Thanks, I see it in line 5432 of rocblas-functions.h

leekillough:
> Approved but must add proto-type
> extern "C" const char* rocblas_status_to_string(rocblas_status status)
> to header file as Andrew mentioned.

It is in rocblas-functions.h, which is where all exported functions go.

Note that rocBLAS does not declare .h or .hpp files "symmetrically" with .cpp files. It declares functions in one header, while it defines them elsewhere, without requiring the declaration and definition to share a common file name.

Instead, headers and .cpp files are arranged by functional area, rather than by name. 

Thus rocblas_auxiliry.cpp defines all auxiliary functions, while rocblas-functions.h declares all external functions.

You can see that all the CI tests passed. That's because this function was added to the list of exported functions in rocblas-functions.h.

It is truncated by GitHub which may be why you missed it.

PR #774, opened 2019-11-02
Update master-hipclang branch 

amdkila:
If merged in, this won't break the commit hash in the manifest.xml file.
As of October 31st, all of rocblas-test passed on hip-clang. Need to run full rocblas-test before merging.


aaronenyeshi:
The CI shows that this is failing on gfx908 for hip-clang

amdkila:
Looks like tests timed out after running for 5 hours on gfx908. Not sure why tests are timing out, as 900/906 took 3.5 hours. Also the log is full of assertion failures and isn't helpful. I'll build rocBLAS and run rocblas-test locally on gfx908 on the latest Tensile develop with hip-clang and see if that passes.

amdkila:
I ran rocblas-test on rack 46 and while it didn't complete (not sure why, the log isn't helpful) it passed 483075/483075 tests, stopping at some gemm-ex blas3 tests. Now, that's only a little over half the tests, so have to take that with a grain of salt. The log also shows that the failures on gfx908 are machine-specific (unable to connect, laggy file system). This should be safe to merge.

PR #776, opened 2019-11-04
batched and strided batched version of iamax iamin

YvanMokwinski:
resolves SWDEV-202341

Summary of proposed changes:
-  implemented batched and strided batched version of iamax / iamin with the use of rocblas_reduction_template and rocblas_reduction_impl.
-  created testing reduction template (that could be also used for asum / nrm2, maybe next refactoring)

### DIFF
--- a/clients/include/device_strided_batch_vector.hpp
+++ b/clients/include/device_strided_batch_vector.hpp
@@ -201,7 +201,7 @@ public:
     hipError_t transfer_from(const host_strided_batch_vector<T>& that)
     {
         return hipMemcpy(
-            this->m_data, that.data(), sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
+            this->data(), that.data(), sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
     }
 
     //!
--- a/clients/include/device_strided_batch_vector.hpp
+++ b/clients/include/device_strided_batch_vector.hpp
@@ -96,6 +96,22 @@ public:
     }
 
     //!
+    //! @brief Returns the data pointer.
+    //!
+    T* data()
+    {
+        return this->m_data;
+    }
+
+    //!
+    //! @brief Returns the data pointer.
+    //!
+    const T* data() const
+    {
+        return this->m_data;
+    }
+
+    //!
     //! @brief Returns the length.
     //!
     rocblas_int n() const
@@ -184,7 +200,8 @@ public:
     //!
     hipError_t transfer_from(const host_strided_batch_vector<T>& that)
     {
-        return hipMemcpy((*this)[0], that[0], sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
+        return hipMemcpy(
+            this->m_data, that.data(), sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
     }
 
     //!
--- a/clients/include/host_strided_batch_vector.hpp
+++ b/clients/include/host_strided_batch_vector.hpp
@@ -208,7 +208,7 @@ public:
         if(that.n() == this->m_n && that.inc() == this->m_inc && that.stride() == this->m_stride
            && that.batch_count() == this->m_batch_count)
         {
-            memcpy(this->m_data, that.data(), sizeof(T) * this->m_nmemb);
+            memcpy(this->data(), that.data(), sizeof(T) * this->m_nmemb);
             return true;
         }
         else
--- a/clients/include/host_strided_batch_vector.hpp
+++ b/clients/include/host_strided_batch_vector.hpp
@@ -99,6 +99,22 @@ public:
     }
 
     //!
+    //! @brief Returns the data pointer.
+    //!
+    T* data()
+    {
+        return this->m_data;
+    }
+
+    //!
+    //! @brief Returns the data pointer.
+    //!
+    const T* data() const
+    {
+        return this->m_data;
+    }
+
+    //!
     //! @brief Returns the length.
     //!
     rocblas_int n() const
@@ -192,7 +208,7 @@ public:
         if(that.n() == this->m_n && that.inc() == this->m_inc && that.stride() == this->m_stride
            && that.batch_count() == this->m_batch_count)
         {
-            memcpy((*this)[0], that[0], sizeof(T) * this->m_nmemb);
+            memcpy(this->m_data, that.data(), sizeof(T) * this->m_nmemb);
             return true;
         }
         else
@@ -209,13 +225,8 @@ public:
     template <size_t PAD, typename U>
     hipError_t transfer_from(const device_strided_batch_vector<T, PAD, U>& that)
     {
-        auto hip_err
-            = hipMemcpy((*this)[0], that[0], sizeof(T) * this->m_nmemb, hipMemcpyDeviceToHost);
-        if(hipSuccess != hip_err)
-        {
-            return hip_err;
-        }
-        return hipSuccess;
+        return hipMemcpy(
+            this->m_data, that.data(), sizeof(T) * this->m_nmemb, hipMemcpyDeviceToHost);
     }
 
     //!
--- a/clients/include/rocblas_cblas.hpp
+++ b/clients/include/rocblas_cblas.hpp
@@ -1,0 +1,88 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+#pragma once
+
+//
+//  CBLAS does not have an  function iamin
+//  neither a rocblas_half based function iamax,
+//  so we write our own version of it
+//
+namespace rocblas_cblas
+{
+    template <typename T>
+    T asum(T x)
+    {
+        return x < 0 ? -x : x;
+    }
+
+    rocblas_half asum(rocblas_half x)
+    {
+        return x & 0x7fff;
+    }
+
+    template <typename T>
+    bool lessthan(T x, T y)
+    {
+        return x < y;
+    }
+
+    bool lessthan(rocblas_half x, rocblas_half y)
+    {
+        return half_to_float(x) < half_to_float(y);
+    }
+
+    template <typename T>
+    bool greatherthan(T x, T y)
+    {
+        return x > y;
+    }
+
+    bool greatherthan(rocblas_half x, rocblas_half y)
+    {
+        return half_to_float(x) > half_to_float(y);
+    }
+
+    template <typename T>
+    void iamin(rocblas_int N, const T* X, rocblas_int incx, rocblas_int* result)
+    {
+        rocblas_int minpos = -1;
+        if(N > 0 && incx > 0)
+        {
+            auto min = asum(X[0]);
+            minpos   = 0;
+            for(size_t i = 1; i < N; ++i)
+            {
+                auto a = asum(X[i * incx]);
+                if(lessthan(a, min))
+                {
+                    min    = a;
+                    minpos = i;
+                }
+            }
+        }
+        *result = minpos;
+    }
+
+    template <typename T>
+    void iamax(rocblas_int N, const T* X, rocblas_int incx, rocblas_int* result)
+    {
+        rocblas_int maxpos = -1;
+        if(N > 0 && incx > 0)
+        {
+            auto max = asum(X[0]);
+            maxpos   = 0;
+            for(size_t i = 1; i < N; ++i)
+            {
+                auto a = asum(X[i * incx]);
+                if(greatherthan(a, max))
+                {
+                    max    = a;
+                    maxpos = i;
+                }
+            }
+        }
+        *result = maxpos;
+    }
+
+} // namespace rocblas_cblas
--- a/clients/include/testing_reduction_batched.hpp
+++ b/clients/include/testing_reduction_batched.hpp
@@ -66,7 +66,7 @@ void template_testing_reduction_batched(const Arguments&                  arg,
         CHECK_HIP_ERROR(dr.memcheck());
 
         CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
-        EXPECT_ROCBLAS_STATUS((func(handle, N, dx.ptr_on_device(), incx, batch_count, dr)),
+        EXPECT_ROCBLAS_STATUS(func(handle, N, dx.ptr_on_device(), incx, batch_count, dr),
                               rocblas_status_invalid_size);
         return;
     }
@@ -80,7 +80,7 @@ void template_testing_reduction_batched(const Arguments&                  arg,
         device_vector<R> dr(std::max(10, batch_count));
         CHECK_HIP_ERROR(dr.memcheck());
         CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_device));
-        CHECK_ROCBLAS_ERROR((func(handle, N, dx.ptr_on_device(), incx, batch_count, dr)));
+        CHECK_ROCBLAS_ERROR(func(handle, N, dx.ptr_on_device(), incx, batch_count, dr));
         return;
     }
 
@@ -116,7 +116,7 @@ void template_testing_reduction_batched(const Arguments&                  arg,
         //
         {
             CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
-            CHECK_ROCBLAS_ERROR((func(handle, N, dx.ptr_on_device(), incx, batch_count, hr1)));
+            CHECK_ROCBLAS_ERROR(func(handle, N, dx.ptr_on_device(), incx, batch_count, hr1));
         }
 
         //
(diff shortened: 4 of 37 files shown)


clients/include/testing_reduction_batched.hpp
```diff
@@ -0,0 +1,201 @@
...
+                rocblas_error_2 = std::max(rocblas_error_2, std::abs((c - a2) / c));
+            }
+        }
+    }
+
+    if(arg.timing)
+    {
+        int number_cold_calls = 2;
+        int number_hot_calls  = 100;
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
```
TorreZuk: We should check to see if we are timing for fetching host results always, or would the device mode be better?

library/include/rocblas-functions.h
```diff
@@ -1406,6 +1406,122 @@ ROCBLAS_EXPORT rocblas_status rocblas_izamax(rocblas_handle                handl
...
+                                                     const rocblas_float_complex* const x[],
+                                                     rocblas_int                        incx,
+                                                     rocblas_int                        batch_count,
+                                                     rocblas_int*                       result);
+
+ROCBLAS_EXPORT rocblas_status rocblas_izamax_batched(rocblas_handle                      handle,
+                                                     rocblas_int                         n,
+                                                     const rocblas_double_complex* const x[],
+                                                     rocblas_int                         incx,
+                                                     rocblas_int  batch_count,
```
TorreZuk: clang format?

library/include/rocblas-functions.h
```diff
@@ -1406,6 +1406,122 @@ ROCBLAS_EXPORT rocblas_status rocblas_izamax(rocblas_handle                handl
...
+    @param[in]
+    n         rocblas_int
+              number of elements in each vector x_i
+    @param[in]
+    x         pointer to the first vector x_i on the GPU.
+    @param[in]
+    incx      rocblas_int
+              specifies the increment for the elements of each x_i. incx must be > 0.
+    @param[in]
+    stridex   rocbla_stride
```
TorreZuk: rocbla_stride type

library/include/rocblas-functions.h
```diff
@@ -1406,6 +1406,122 @@ ROCBLAS_EXPORT rocblas_status rocblas_izamax(rocblas_handle                handl
...
+                                                     const rocblas_double_complex* const x[],
+                                                     rocblas_int                         incx,
+                                                     rocblas_int  batch_count,
+                                                     rocblas_int* result);
+
+/*! \brief BLAS Level 1 API
+
+    \details
+     amax_strided_batched finds over a strided batch of vectors the first index of the element of maximum magnitude of real vector x
+         or the sum of magnitude of the real and imaginary parts of elements if x is a complex
```
TorreZuk: Conjunction is ambiguous.  Maybe a new sentence as I presume it never finds a sum.

library/include/rocblas-functions.h
YvanMokwinski: good catch

library/include/rocblas-functions.h
YvanMokwinski: good point

library/include/rocblas-functions.h
YvanMokwinski: mystery

clients/include/testing_reduction_batched.hpp
YvanMokwinski: To see with Wasiq

library/include/rocblas-functions.h
leekillough: > 
> 
> clang format?

> mystery

Does the `.githooks/pre-commit` script get run when you commit your changes? It requires the latest ROCm to be installed, for it to get `clang-format`.

Are you using a MS-DOS or Mac editor? The `.githooks/pre-commit` script converts CR-LF to LF for line endings, and removes trailing spaces on the lines. I did not anticipate LF-CR, so I did not plan for the script to standardize it to LF.

You should make sure your environment runs `.githooks/pre-commit` correctly. On Linux, it "just works".

clients/include/testing_reduction_batched.hpp
leekillough: > 
> 
> We should check to see if we are timing for fetching host results always, or would the device mode be better?

Host mode is usually better, if any scalar arguments have to be used on the host. We can pass scalars from the host to the device in the kernel call after loading them on the host, but we cannot pass scalars from the device to host without using a synchronizing memory copy.

Pointer mode only has to do with scalars. It does not affect vectors, matrices, tensors, etc., which area always on the device memory.

clients/include/device_strided_batch_vector.hpp
```diff
@@ -184,7 +200,8 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
     //!
     hipError_t transfer_from(const host_strided_batch_vector<T>& that)
     {
-        return hipMemcpy((*this)[0], that[0], sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
+        return hipMemcpy(
+            this->m_data, that.data(), sizeof(T) * this->nmemb(), hipMemcpyHostToDevice);
```
leekillough: For symmetry, I'd use `this->data()` and `that.data()`, or `this->m_data` and `that.m_data`.

clients/include/host_strided_batch_vector.hpp
```diff
@@ -209,13 +225,8 @@ class host_strided_batch_vector
...
-        auto hip_err
-            = hipMemcpy((*this)[0], that[0], sizeof(T) * this->m_nmemb, hipMemcpyDeviceToHost);
-        if(hipSuccess != hip_err)
-        {
-            return hip_err;
-        }
-        return hipSuccess;
+        return hipMemcpy(
+            this->m_data, that.data(), sizeof(T) * this->m_nmemb, hipMemcpyDeviceToHost);
     }
```
leekillough: Same as above.

clients/include/rocblas_cblas.hpp
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    max    = a;
+                    maxpos = i;
+                }
+            }
+        }
+        *result = maxpos;
+    }
+
+} // namespace rocblas_cblas
```
leekillough: Why was this moved to `rocblas_cblas.hpp` ? It was very specific to the IAMAX/IAMIN tests, which is why it was located there. I do not anticipate using them outside of that file. This increases the size of `rocblas_cblas.hpp` for something which is only used in one place.

clients/include/testing_reduction_batched.hpp
```diff
@@ -0,0 +1,201 @@
...
+                                        rocblas_reduction_batched_t<T, R> func,
+                                        void (*CBLAS_FUNC)(rocblas_int, const T*, rocblas_int, R*))
+{
+    rocblas_int          N = arg.N, incx = arg.incx, batch_count = arg.batch_count;
+    rocblas_stride       stride_x = arg.stride_x;
+    double               rocblas_error_1, rocblas_error_2;
+    rocblas_local_handle handle;
+
+    if(batch_count < 0)
+    {
```
leekillough: **This is being refactored as part of the argument error check changes.** I have already made changes to {IAMAX,IAMIN,ASUM,NRM2} batched tests; similar changes will be needed here now or later.

Basically:

if `N<=0 || incx <= 0 || batch_count == 0`, then the result is the `default_value` of the reduction, i.e., the reduction when applied to a vector of 0 elements. There is no error condition (return `rocblas_status_success`).

If `N > 0 && incx > 0 && batch_count < 0`, it is a `rocblas_status_invalid_size`.

So the normal flow is like:
```
if (N <= 0 || incx <=0 || batch_count <= 0)
{
   // Allocate safe size
   // ..
   EXPECT_ROCBLAS_STATUS(func(handle, N, dx.ptr_on_device(), incx, batch_count, dr).
                              N > 0 && incx > 0 && batch_count < 0 ? rocblas_status_invalid_size :
                              rocblas_status_success);
}
```

clients/include/testing_reduction_batched.hpp
```diff
@@ -0,0 +1,201 @@
...
+        device_batch_vector<T> dx(10, 1, 10);
+        CHECK_HIP_ERROR(dx.memcheck());
+        device_vector<R> dr(10);
+        CHECK_HIP_ERROR(dr.memcheck());
+
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
+        EXPECT_ROCBLAS_STATUS((func(handle, N, dx.ptr_on_device(), incx, batch_count, dr)),
```
leekillough: Parentheses are not needed around `(func(...))`.

Extra parentheses are only required when the outer expression in a macro argument contains commas not surrounded by parentheses, braces or backets, such as `func<T**,**U>(arg1, arg2)`. That needs to be parenthesized at least as `(func<T,U>)(arg1, arg2)` but you can use `(func<T,U>(arg1, arg2))` if you are more comfortable with it. When only one template argument is needed, such as `func<T>`, then no parentheses are necessary because there is no comma unsurrounded by parentheses, brackets or braces in the top-level expression.

clients/include/testing_reduction_batched.hpp
```diff
@@ -0,0 +1,201 @@
...
+                              rocblas_status_invalid_size);
+        return;
+    }
+
+    // check to prevent undefined memory allocation error
+
+    if(N <= 0 || incx <= 0 || batch_count == 0)
+    {
+        device_batch_vector<T> dx(10, 1, 2);
```
leekillough: See above. It's usually better to combine the early-exit `N <=0` and error `batch_count < 0` cases into one combined OR of them, and then the only difference required is the expect return status, which is either `rocblas_status_success` or `rocblas_status_invalid_size`. All the rest of the code is the same.

As said above, this refactoring is being done across all of rocBLAS now, to make sure early exits and argument errors are handled correctly.

clients/include/testing_reduction_strided_batched.hpp
```diff
@@ -0,0 +1,212 @@
...
+    //
+    // allocate memory on device
+    //
+    device_vector<T> dx(batch_count);
+    CHECK_HIP_ERROR(dx.memcheck());
+
+    R h_rocblas_result;
+
+    EXPECT_ROCBLAS_STATUS(
+        (func(handle, N, nullptr, incx, incx * N, batch_count, &h_rocblas_result)),
```
leekillough: Unnecessary parentheses around `(func(...))`

clients/include/testing_reduction_strided_batched.hpp
```diff
@@ -0,0 +1,212 @@
...
+    rocblas_reduction_strided_batched_t<T, R> func,
+    void (*CBLAS_FUNC)(rocblas_int, const T*, rocblas_int, R*))
+{
+    rocblas_int    N = arg.N, incx = arg.incx, batch_count = arg.batch_count;
+    rocblas_stride stridex = arg.stride_x;
+
+    double               rocblas_error_1, rocblas_error_2;
+    rocblas_local_handle handle;
+    if(batch_count < 0)
+    {
```
leekillough: See above.

clients/include/testing_reduction_strided_batched.hpp
```diff
@@ -0,0 +1,212 @@
...
+        CHECK_HIP_ERROR(dr.memcheck());
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_device));
+        EXPECT_ROCBLAS_STATUS(func(handle, N, dx, incx, stridex, batch_count, dr),
+                              rocblas_status_invalid_size);
+        return;
+    }
+
+    // check to prevent undefined memory allocation error
+    if(N <= 0 || incx <= 0 || batch_count == 0)
+    {
```
leekillough: See above. The two cases should be merged together and then only the expected return status needs to be differentiated. This change is being made globally in rocBLAS as part of the early exit and argument checking refactor.

library/src/blas1/rocblas_amax_amin.h
```diff
@@ -39,9 +30,11 @@ struct rocblas_default_value<index_value_t<T>>
     }
 };
 
-// Fetch absolute value
+//!
+//! @brief Struct-operator to fetch absolute value
+//!
 template <typename To>
-struct rocblas_fetch_amax_amin
+struct rocblas_fetch_index_value
```
leekillough: `amax_amin` seems more intuitive than `index_value`, if the `struct` you are passing is supposed to designate "how to fetch the values for <insert_function_here>", i.e., "How to fetch the values for `amax`, `amin`," as opposed to `asum` or `nrm2`, makes more sense than"how to fetch the values for `index_value`" (whatever that means). `index_value` is an internal implementation detail which does not need to be exposed when expressing `amax`, `amin`.

library/src/blas1/rocblas_amax_amin.h
```diff
@@ -50,117 +43,14 @@ struct rocblas_fetch_amax_amin
...
-        }
-    }
-};
-
-struct rocblas_finalize_amax_amin
+//!
+//! @brief Struct-operator to finalize the data.
+//!
+struct rocblas_finalize_index_value
 {
```
leekillough: Same comment as above.

library/src/blas1/rocblas_iamax.cpp
```diff
@@ -0,0 +1,71 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamax.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Unnamed namespace is not required, since all of its contents are `static`.

Either use `static`, or an unnamed namespace. Using both is redundant and longer than necessary.

library/src/blas1/rocblas_iamax_batched.cpp
```diff
@@ -0,0 +1,84 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamax_batched.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Same as above. Use either unnamed namespace or `static`, not both.

library/src/blas1/rocblas_iamax_strided_batched.cpp
```diff
@@ -0,0 +1,88 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamax_strided_batched.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Again

library/src/blas1/rocblas_iamin.cpp
```diff
@@ -0,0 +1,71 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamin.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Again

library/src/blas1/rocblas_iamin_batched.cpp
```diff
@@ -0,0 +1,83 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamin_batched.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Again

library/src/blas1/rocblas_iamin_strided_batched.cpp
```diff
@@ -0,0 +1,88 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_iamin_strided_batched.hpp"
+#include "rocblas_reduction_impl.hpp"
+
+namespace
+{
```
leekillough: Again

clients/include/rocblas_cblas.hpp
YvanMokwinski: It helps sometimes to structure the code with one file => one class/struct.

clients/include/device_strided_batch_vector.hpp
YvanMokwinski: Good point

clients/include/testing_reduction_batched.hpp
YvanMokwinski: mistake

library/src/blas1/rocblas_iamax.cpp
YvanMokwinski: okay

library/include/rocblas-functions.h
YvanMokwinski: Using rack72 with a linux OS ... but the clang-format checking on the CI passed!..

YvanMokwinski: Thanks, I don't think the hook is running ... I already had clang format failures with the CI. Hence, I have my own command line to loop over and applying the clang format. 
But again, the CI passed on the format.

clients/include/testing_reduction_batched.hpp
YvanMokwinski: Fixed.

clients/include/testing_reduction_strided_batched.hpp
YvanMokwinski: Fixed

clients/include/testing_reduction_batched.hpp
YvanMokwinski: Fixed

clients/include/rocblas_cblas.hpp
YvanMokwinski: Renamed rocblas_cblas with rocblas_iamax_iamin_ref and added routine returning Fortran indices.

PR #779, opened 2019-11-05
New Tensile client plus extensive rocBLAS refactoring changes

leekillough:
1. Refactor the new Tensile client, combining `gemm` and `gemm_strided_batched` into one  `RocblasContractionProblem` construction (with `gemm` using a `batch_count` of 1).
2. Add logic to determine the directory of `librocblas.so` at runtime, and to search for Tensile host files in paths relative to `librocblas.so`'s directory location. Exact details to be determined later.
3. Redefine `rocblas_half` to `_Float16`, and replace all `_Float16` instances in rocBLAS with `rocblas_half`, except for the definition of `rocblas_half`. Simplify code to not have to use `float_to_half()` and `half_to_float()` functions. `half_to_float()` is replaced with an implicit conversion to `float` (no loss), while `float_to_half()` is replaced with an explicit conversion to `rocblas_half` (it should be explicit since it involves loss of precision, although I don't think that the compiler enforces this).
4. Remove `stride_alpha` and `stride_beta` from Level-3 BLAS routines. As discussed in email, Tensile is unlikely to support vectored `alpha` and `beta` anytime soon, and Tensile constructs problems with alpha and beta _by value_, which is hard to do with batched `alpha` and `beta`. Also, register pressure would be too high if `stride_alpha` and `stride_beta` were kept in registers.
5. Fix `std::vector` initialization in Tensile client code. In the past, we were using initializations like `{2}` which initialized a `std::vector` embedded object with a single `std::initializer_list` element, instead of initializing the `std::vector` with 2 elements. The syntax was changed to use `(2)` instead of `{2}`.
6. Remove the `*kernel_name` functions, replacing them with functions which return `rocblas_status_not_implemented`.
7. Prepare for GEMM_EX to use new Tensile client. (The foundations have been set, but it hasn't been implemented yet.)
    - New functions for logging `alpha` and `beta`, whether real or complex, and handle `nullptr`.
    - Remove unnecessary and unused code.
    - Simplify `device_strided_batched_matrix_copy` to be templated and use ordinary pointer arithmetic, instead of passing an element size and doing the calculations at the byte level.
8. Change many (**not all, yet**) rocBLAS functions and tests to abide by Legacy BLAS rules regarding quick returns and invalid arguments. In particular, `K == 0` is not an early exit for `GEMM` functions, since `C` still needs to be multiplied by `beta`. Put comments to this effect in the code, since it's very easy to mistake `K == 0` as an early exit condition. On the other hand, `GEMV` does not multiply `Y` by `beta` when `M == 0` or `N == 0`. It's a strange historical bug which must be preserved. (I can send references.)
9. The batched and strided batched tests used to have two "exceptional" cases at their beginning, one when the arguments are invalid, and one when the arguments have early-exit conditions. These two cases have been merged into one, since 75% of the code is similar, and the only thing which is different, is the expected return status.
10. Remove redundant parentheses from some routines. When you call a macro with an argument with embedded commas, these commas need to be guarded around parentheses, brackets or braces. Template arguments like `<T, U>` do not count as bracketing expressions. But when you only have `<T>`, then no comma appears, and no extra parentheses are required. I simplified a few cases where redundant parentheses were used.
11. Refactor `rocblas-bench` to use `std::map` tables instead of long `if(!strcmp(...)){ } else if(!strcmp(...)){ } else...` chains. The code may be slightly slower, but it is much more compact and maintainable. The `rocblas-bench` code's time spent dispatching functions by name is much less important than the maintainability of the code.
12. Add `rocblas_test_valid` as a corresponding base class to `rocblas_test_invalid`, and define it to return `true` when converted to `bool`, and to have an abstract interface which must be defined in a derived class. This simplifies the code everywhere, by not requiring you to define `operator bool()` as returning `true`, and forcing you to define `operator()()`.
13. Replace the `batch` command-line-arguments and YAML names with `batch_count`, since `batch_count` is less ambiguous, and is used almost everywhere else. For `rocblas-bench`, add a function which changes `--batch` to `--batch_count` while displaying a warning once.
14. Improve `.githooks/pre-checkin` to handle `.py` and `.sh` files.

**2360 fewer lines of code!!!**


leekillough:
Oh, and batched TRSM had some changes with respect to memory allocation. The code is simpler nonetheless.

leekillough:
I don't understand the CI failure with `tensile_host.cpp`. I just ran `clang-format` on it, and it was unchanged.

leekillough:
I wonder if it is a chicken-and-egg problem, since I had changes to `pre-checkin`.

leekillough:
It passed the new client. I may need to fix it for the old client.

library/src/blas1/rocblas_rot_batched.cpp
```diff
@@ -57,7 +57,7 @@ namespace
                       incx,
                       "--incy",
                       incy,
-                      "--batch",
+                      "--batch_count",
```
TorreZuk: NOT for this PR but should have created constants for all these standardized parameter strings.

library/src/blas3/Tensile/gemm.cpp
```diff
@@ -28,20 +23,20 @@ namespace
...
-                                     rocblas_int       k,
-                                     const T*          alpha,
-                                     const T*          A,
-                                     rocblas_int       ld_a,
-                                     const T*          B,
-                                     rocblas_int       ld_b,
-                                     const T*          beta,
-                                     T*                C,
-                                     rocblas_int       ld_c)
+    auto rocblas_gemm_impl(rocblas_handle    handle,
```
TorreZuk: Overuse of auto may reduce readability.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -743,63 +667,32 @@ rocblas_status gemm_ex_handle_transpose(rocblas_handle    handle,
     }
     else
     {
-        device_strided_batched_matrix_copy(
-            c, ldc, stride_c, d, ldd, stride_d, m, n, batch_count, sizeof(To));
+        device_strided_batched_matrix_copy(c, ldc, stride_c, d, ldd, stride_d, m, n, batch_count);
         c_in     = d;
         ldi      = ldd;
         stride_i = stride_d;
```
TorreZuk: Potential data loss moving to smaller type?    Looks like tensile doesn't support the 64bit int for stride either?

library/include/rocblas-types.h
```diff
@@ -38,9 +38,18 @@ typedef int64_t rocblas_stride;
...
+
+// Clang supports _Float16 as an extension in C11 and C++11 and later
+#if __cplusplus >= 201103L || __STDC_VERSION__ >= 201112L
+typedef _Float16 rocblas_half;
+#else
+typedef struct
+{
+    uint16_t data;
+} rocblas_half;
+#endif
```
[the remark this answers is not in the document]
leekillough: > typedef struct
> {
 >   uint16_t data;
> } rocblas_half;
> #endif
>
> Will this case even be handled by the code now?

That is needed when someone compiles with a C/C++ compiler which doesn't recognize `_Float16` but only needs to pass pointers to 16-bit values to rocBLAS. It is similar to the old `rocblas_half`, except that it has a wrapper `struct` to distinguish it from `uint16_t`, which, under the old scheme, could not be disambiguated from `rocblas_half` since `rocblas_half` was simply a `typedef uint16_t rocblas_half;`

Clang supports `_Float16` in C11 and C++11 and later, hence the version checks around this declaration. GCC has wavered, sometimes including it in C++ mode but sometimes saying it shouldn't be, because it's only mentioned in the C11 standard and not in C++11.  There is no feature flag that I know of for detecting `_Float16` support, which isn't necessarily defined as a macro, although maybe with SFINAE tricks you could figure it out in C++ (something we cannot do in this header file, which must compile in C as well as C++).

If we need to add GCC-specific checks for `_Float16` support, they can be added.

We were already using `_Float16` when calling the old Tensile client; this just standardizes rocBLAS to use `_Float16` for `rocblas_half` and to avoid using any other type names in the code, so that we have one definition for the whole library.

Eventually, all C++ compilers will support `_Float16`, `__fp16`, or some variant.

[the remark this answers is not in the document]
TorreZuk: Okay, I just wasn't sure all your other changes to implicit conversions to float will still work for these old compilers that use this struct.

leekillough:
> This PR is definitely too large: it makes reviewing too time consuming, reviewers please don't comment outside a review as we don't want that many emails! Additionally it makes cherry picking, reversion, and other code maintenance difficult. This sprint retrospective should include setting guidelines on PRs that should be followed. E.g. code for separate issues must be in separate PRs unless very tighlty coupled.

The PR is large because it is hard to do it in increments, especially with all the other changes going on right now. It took more than a day just to merge it with previously merged PRs. It took another day to get all the tests to pass by fixing the "expected behavior" for many routines.

The refactoring to use the new Tensile client is a big change, and needs refactoring GEMM to be organized a little differently, if we are going to keep the separation of the C APIs, the `*_impl` routines and the `*_template` templates, while also minimizing duplication.

The removal of `stride_alpha` and `stride_beta` was discussed before; it affects the GEMM routines and their callers the most, which is why it's in this PR.

The fixing of the argument checks to return success before detecting argument errors when the sizes or increments are negative (in the case of BLAS-1) or zero, and the changing of the tests to expect the new behavior, is a large part of this change, and will continue in several more iterations. A full sweep of all of rocBLAS needs to be done to make sure each and every routine is doing error checking and early-exits correctly, and tests need to be added to detect regressions from this expected behavior. That will be another large PR (currently in draft) affecting dozens of files.

Some of the refactoring, e.g., the function tables in `rocblas-bench`, or the redefinition of `rocblas_half` to `_Float16`, could be done in separate PRs, but splitting these changes up would take extra time, and we are very short on time. Also, part of why `_Float16` changes are in this PR, is because `_Float16` was tied to the old Tensile client code (it cast `rocblas_half*` to `_Float16*`).

This large PR is part of the reason I have not had time to review other PRs.

> I think the tensile host globbing of co files design should also be revisited as would think we should pre-process this before runtime. Outside of this PR of course.

There was a concall last week on how to organize the Tensile code object files; more discussions are likely forthcoming. We've considered adding them to lightweight archives like `ar` and extracting them into memory on demand at runtime.

It was said in the meeting that there were 3 different clients (customers) of Tensile, including rocBLAS, and that differences between their needs is why rocBLAS has to load the `.co` files itself. Ideally, this is something that should be done automatically inside of Tensile, but since Tensile is not a separate product and is packaged as part of rocBLAS, we have to use rocBLAS packaging conventions and so the onus is on rocBLAS to come up with and follow those conventions, and as long as other clients want to do things differently, we need to do it at the Tensile client level instead of inside of Tensile.

The current glob idea is more of a proof-of-concept than production code, since there might be thousands of `.co` files, and having that many files can put a load on the filesystem and OS. This PR is just one step of the way towards moving to the new Tensile client. There will be other iterations of this move to the new Tensile client, including when we eventually get rid of the old client entirely, which won't be for another few releases.

library/include/rocblas-types.h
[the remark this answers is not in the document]
leekillough: > Okay, I just wasn't sure all your other changes to implicit conversions to float will still work for these old compilers that use this struct.

We won't be using the old compilers to build the code which does the conversions to `float`. We will be using compilers which support `_Float16`, which can be implicitly converted to `float` and `double`.

For old compilers, we just need an opaque datatype which can be `memcpy`'ed from FP16 data, and pointers to `rocblas_half` can be converted to/from pointers to FP16 data.

We may need to add more checks to handle GCC, since, unlike Clang, GCC has not expressed universal support for `_Float16` in C++.

If you are using or allowing GCC to be used for the `examples`, then we might have to account for it, and make sure that they build and run correctly.

library/src/blas_ex/rocblas_gemm_ex.hpp
leekillough: > Potential data loss moving to smaller type? Looks like tensile doesn't support the 64bit int for stride either?

We may need to file a Tensile bug then, if it does not support 64-bit stride.

There are many places in rocBLAS where `int` types and sizes are used carelessly, such as using `int` instead of `rocblas_int` for a loop variable whose limits are `rocblas_int`, or not converting to `size_t` before doing a size calculation with multiplication. We need to go through and sweep rocBLAS looking for size calculations which may truncate or overflow.

The change referred to here was only `clang-format` changes after the removal of a `sizeof(To)` argument in a call, and instead making `To` a template parameter automatically deduced from the `c` and `d` pointers.

Formerly:
```
void func(..., size_t elemsize)
{
    size_t size = elemsize;
}

func(..., c, ..., sizeof(To)):
```
New:
```
template <typename To>
void func(..., To *c, ...)
{
   size_t  size =  sizeof(To);
}

func(..., c, ...);
```
So the change does not introduce any new truncation from 64-bit to 32-bit. That problem already existed and will remain until Tensile supports 64-bits, which is beyond the scope of this PR.

library/src/blas1/rocblas_rot_batched.cpp
leekillough: > NOT for this PR but should have created constants for all these standardized parameter strings.

That would require a large refactoring, and may reduce readability if variables are used everywhere that string literals are used now. It may also require compile-time or run-time concatenation of names and option dashes "--". It would also have to span multiple languages (C++, Python, YAML). Right now the strings seem to be sufficient.

library/src/blas_ex/rocblas_gemm_ex.hpp
sdquiring: Tensile uses size_t for all strides right now.  Not sure where you got that it doesn't support 64-bit strides?

We may be moving to int64_t at some point to allow specifying -1 as an "auto" stride.

library/src/blas3/Tensile/gemm.cpp
leekillough: > Overuse of auto may reduce readability.

I don't know how that got in there -- when the return type of a function is a simple name like `rocblas_status`, it should be used. It's only when the type is very long or would require `decltype()` that `auto` may simplify things.

library/src/blas_ex/rocblas_gemm_ex.hpp
leekillough: > 
> 
> Tensile uses size_t for all strides right now. Not sure where you got that it doesn't support 64-bit strides?
> 
> We may be moving to int64_t at some point to allow specifying -1 as an "auto" stride.

That's what I thought; right now `rocblas_stride` is defined as `int64_t` but it could just as well be `size_t` -- both are the same size, and conversion between them is a no-op in 2's complement.

TorreZuk: Sorry my comment on stride related to the "tensile" calls in code context right above the lines of code I highlighted (not to the tensile library in general).  E.g. code #define pasted below which I think has already been changed.  32bit integers were being used for strides.  We were discussing negative strides to allow reverse ordering in rocblas so I think this may still be a consideration we need to iron out with tensile, otherwise I had no concern to raise with tensile itself.      

#define TENSILE_IN_ARGS(Ti, To, Tc)                                                         \
    To *dataD, const To *dataC, const Ti *dataA, const Ti *dataB, Tc alpha, Tc beta,             \
    To *dataD, const To *dataC, const Ti *dataA, const Ti *dataB, Tc alpha, Tc beta,        \
        unsigned int strideD1J, unsigned int strideD2K, unsigned int strideC1J,                  \
        unsigned strideD1J, unsigned strideD2K, unsigned strideC1J, unsigned strideC2K,     \
        unsigned int strideC2K, unsigned int strideA1L, unsigned int strideA2K,                  \
        unsigned strideA1L, unsigned strideA2K, unsigned strideB1J, unsigned strideB2K,     \
        unsigned int strideB1J, unsigned int strideB2K, unsigned int sizeI, unsigned int sizeJ,  \
        unsigned sizeI, unsigned sizeJ, unsigned sizeK, unsigned sizeL, hipStream_t stream, \
        unsigned int sizeK, unsigned int sizeL, hipStream_t stream, unsigned int numInputEvents, \
        unsigned numInputEvents, void *dummy1, void *dummy2

leekillough: > We were discussing negative strides to allow reverse ordering in rocblas so I think this may still be a consideration we need to iron out with tensile, otherwise I had no concern to raise with tensile itself.

We already decided that negative strides should be handled just like any other in the kernel code, and that they should not be handled like `incx` and `incxy` are handled in `AXPY`, `SWAP`, etc.

Negative strides in the test code simply involves handling a mismatch between the stated problem and the pointers returned by allocated memory: The stated problem is one with a starting address and negative strides, while the pointers returned for allocated memory always start at the lower address. Hence we need to adjust the pointers to the end of allocated memory, if the problem has negative strides. But the rocBLAS and Tensile kernels do not need to do anything special -- they simply take a starting address and a stride, which can be positive or negative.

The macros you show above are from the **old** Tensile client code, which was probably never fully ported to use the new `rocblas_stride` type. The **new** Tensile client code always represents strides using `rocblas_stride`, which is defined as `int64_t` right now.

leekillough:
The CI failures seem to be timeouts.

PR #781, opened 2019-11-05
Swdev 209194 async get set vec matrix

TorreZuk:
Adds async get/set vector/matrix API.
* host_pinned_vector added for testing
* pinned_memory_allocator added for STL use in testing


YvanMokwinski (approved):
Approved.
One comment:
- the host_pinned_vector could be re-written as a host_base_vector taking the allocator type as a parameter. In this way, we could have

template<typename T>
using host_pinned_vector = host_base_vector<T, pinned_memory_allocator < T > >;

but also 

template<typename T>
using host_vector = host_base_vector<T, std::allocator < T > >;

That would get rid of the code duplication between host_vector and host_pinned_vector.

TorreZuk:
Thanks Yvan, I will look at that refactor and see how it looks.   I don't want to bloat the 99% uses of host vectors with passing the std::allocator just for my 2 test cases so will see how it looks.   I could also delete some of the code not used in the host_pinned_vector class.

leekillough:
> 
> 
> Thanks Yvan, I will look at that refactor and see how it looks. I don't want to bloat the 99% uses of host vectors with passing the std::allocator just for my 2 test cases so will see how it looks. I could also delete some of the code not used in the host_pinned_vector class.

You can make `std::allocator` a default template argument, like `std::vector` and other containers do.

clients/include/host_pinned_vector.hpp
```diff
@@ -0,0 +1,99 @@
+//
```
leekillough: Do we need a separate class for pinned memory? Or can we just pin memory on a `host_vector`, using `mlock()`?

Pinning memory seems an orthogonal concern to data layout. We do not need custom allocators or data structures to pin the memory. We can just add a function to our other classes to "pin the memory owned by this class".

clients/include/host_pinned_vector.hpp
```diff
@@ -0,0 +1,99 @@
...
+#pragma once
+#include "pinned_memory_allocator.hpp"
+#include <memory>
+
+//!
+//! @brief  Pseudo-vector subclass which uses host pinned memory.
+//!
+template <typename T>
+struct host_pinned_vector : std::vector<T, pinned_memory_allocator<T>>
```
leekillough: You do not need to pin memory at allocation time -- it can be pinned after allocation with `mlock()`.

So you don't need custom allocators or separate classes for pinned memory -- it's just an OS call which pins the memory already owned by a data structure.

leekillough (changes requested):
I would suggest reusing the existing `Vector` classes, and adding member functions to them to `pin()` or `unpin()`.

```
host_vector<T>  myvector(n);
myvector.pin();
```

The member function just needs to call `mlock()` on the address range of the vector object.

clients/include/host_pinned_vector.hpp
TorreZuk: I think we want consistency with the hip API into which we are passing the memory.  Memory flags may be used by hip to determine what needs to be done, maybe it shouldn't be called pinned memory but hipHostAllocator.   I don't think we want to directly use low level memory bindings such as mlock which will be brittle if hip changes what they do behind the scenes, or they add windows support, or want to manage over allocations.  The use case is just for hipMemcpyAsync.   I can rename the class but now thinking I'll just merge what I have that works and can later see if other memory constraints arise in the next release and refactor when we get additional requirements.   For now this is only to be used in testing.

clients/include/host_pinned_vector.hpp
TorreZuk: See my previous comment.   What you propose would work but would likely break if hip changes they way they manage memory so is less maintainable.   This simply allocates the memory in the way required, so one could argue is closer to the resource allocation is initialization pattern.

TorreZuk:
Lee see my earlier comments, I still propose merging without changes for now and not making the suggested changes.

clients/include/host_pinned_vector.hpp
leekillough: `mlock()` is universal in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mlock.html). HIP is unlikely to change this.

If HIP is this fragile, then it needs its own APIs for pinning / unpinning memory, and we use those APIs instead of using POSIX APIs.

Creating new classes or allocators is unnecessarily complicated. Pinning is an operation which is independent from data layout or our own data structures.

leekillough: Also, you can add an extra constructor argument to the existing constructors, if you want to call a different memory allocator.

Something like:
```
 host_vector<T>(HIP_Pinned_Memory, x, ...);
```
and then define an alternative constructor which calls the pinned HIP memory allocator.

There is no need to change the type of the `host_vector` or add extra template parameters. We already have too many types.

If you insist on using a different type name, you can use a derived class:
```
template <typename T>
class host_vector_pinned<T> : public host_vector<T>
{
public:
    template< typename... Ts >
    host_vector_pinned(Ts&&... args) : host_vector<T>( HIP_Pinned_Memory, std::forward<Ts>(args)...) {}
};
```
}

leekillough:
> 
> 
> Lee see my earlier comments, I still propose merging without changes for now and not making the suggested changes.

I would try to find if there is a portable HIP way to pin memory which is separate from allocation, and use that to decouple the allocation and construction of `host_vector` objects from their pinning. Then pinning just becomes an extra step, and we do not need extra type names or template parameters.

HIP documentation is not very good; I briefly looked but could not find functions to pin arbitrary memory.

dcyca:
[hipHostRegister](https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674) and and hipHostUnregister could be used to implement @leekillough's suggested pin() interface.

TorreZuk:
Thanks Dan.  Yes I saw the hipHostRegister interface. Our training slides on it also comment it only works on some systems.  I wanted to use exactly the same memory allocation in my tests as I will be advising clients to use.   I assume we will advise they use hip calls so hipMalloc and hipHostMalloc.

Derived STL classes and allocators don't play well together, I have tried multiple variations.   I just want to put in my simple design that should work for this release and then I or someone else can make a more "elegant" variation when we get further requirements or use cases.    We can add pin and unpin to the standard host_vector after testing more variations and down the road delete the special case class.

leekillough:
> [hipHostRegister](https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674) and and hipHostUnregister could be used to implement @leekillough's suggested pin() interface.

Thanks, I was hoping there was such an API.

leekillough (approved):

PR #783, opened 2019-11-05
bf16 unit tests

AlexBrownAMD:
Edit: The new tests can be run by themselves with the following command+filter:

./rocblas-test --gtest_filter=\*gemm\*bf16\*TN\*512\*:\*gemm\*bf16\*TN\*1024\*:\*gemm\*bf16\*TN\*2048\*:\*gemm\*bf16\*TN\*4096\*


library/include/rocblas_bfloat16.h
```diff
@@ -57,8 +57,8 @@ struct rocblas_bfloat16
     __host__ __device__ rocblas_bfloat16() = default;
 
     // round upper 16 bits of IEEE float to convert to bfloat16
-    explicit constexpr __host__ __device__ rocblas_bfloat16(float f)
-        : data(float_to_bfloat16(f))
+    explicit constexpr __host__ __device__ rocblas_bfloat16(float f, bool round = true)
+        : data(round ? float_to_bfloat16(f) : truncate(f))
```
yoichiyoshida: It looks like the first function takes in round, is this ternary still needed and the truncate function needed?

library/include/rocblas_bfloat16.h
```diff
@@ -74,45 +74,58 @@ struct rocblas_bfloat16
     }
 
 private:
-    static constexpr __host__ __device__ uint16_t float_to_bfloat16(float f)
+    static constexpr __host__ __device__ uint16_t float_to_bfloat16(float f, bool round = true)
```
leekillough: It would be better if this truncation function were implemented somewhere in `clients/*` instead of in `library/include/rocblas_bfloat16.h`. Do you need it anywhere outside of the tests?

There is an associated `tensile_bfloat16.h`. Right now only the names, as the use of C++14 `constexpr`, is different between `rocblas_bfloat16.h` and `tensile_bfloat16.h`. This change would be making them diverge further.

library/include/rocblas_bfloat16.h
```diff
@@ -74,45 +74,58 @@ struct rocblas_bfloat16
...
     {
         union
         {
             float    fp32;
             uint32_t int32;
         } u = {f};
-        if(~u.int32 & 0x7f800000)
+        if(round)
```
leekillough: This logic is incorrect. We do not want it to apply if `~u.int32 & 0x7f800000` returns 0, which means the value is either Inf or NaN.

Why are you modifying `rocblas_bfloat16.h`? It seems like this is better handled in the testing code.

clients/include/testing_gemm_ex.hpp
```diff
@@ -241,6 +241,61 @@ void testing_gemm_ex_bad_arg(const Arguments& arg)
...
+}
+
+namespace
+{
+    bool is_replacement_kernel(rocblas_int m, rocblas_int n)
+    {
+        if((m == 512 && n == 512) || (m == 1024 && n == 1024) || (m == 2048 && n == 2048)
+           || (m == 4096 && n == 4096) || (m == 960 && n == 1024) || (m == 3840 && n == 4096))
+            return true;
+        return false;
```
leekillough: Why are certain sizes singled out as far as expected answers?

Our tests should usually be "black-box", i.e., without knowledge of internal implementation details. So we should not single out certain sizes as behaving differently, but we should apply the same standard to all sizes.

For some precisions, this might require the use of `near` instead of `unit` testing functions.

clients/include/testing_gemm_ex.hpp
```diff
@@ -241,6 +241,61 @@ void testing_gemm_ex_bad_arg(const Arguments& arg)
...
+    host_vector<float> C_float(size_C);
+    for(int i = 0; i < size_C; ++i)
+        C_float[i] = C[i];
+    cblas_gemm<rocblas_bfloat16, float, float>(
+        transA, transB, m, n, k, alpha, A, lda, B, ldb, beta, C_float, ldc);
+    bool round = !is_replacement_kernel(m, n);
+    for(int i = 0; i < size_C; ++i)
+        C[i] = rocblas_bfloat16(C_float[i], round);
+}
+
```
leekillough: This `reference_gemm` needs to support all combinations of `rocblas_gemm_ex`, `rocblas_batched_ex`, and `rocblas_strided_batched_ex`.

Is there any reason `cblas_gemm` is not adequate, even if we convert to `float` first?

leekillough:
Also, see [PR 779](https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/779) which will interact with these changes. GEMM is undergoing extensive refactoring.

clients/include/testing_gemm_ex.hpp
AlexBrownAMD: Those particular sizes are currently implemented using replacement kernels that we want to test. The replacement kernel truncates the answer rather than rounds so a different test behaviour is needed. The test is already using a near check rather than a unit check.

clients/include/testing_gemm_ex.hpp
AlexBrownAMD: cblas_gemm<bfloat16, bfloat16, float> rounds the answer. cblas_gemm<bfloat16, float, float> works, but calling it is a special case since To doesn't match the test case To. The alternative is to write a specialized version of testing_gemm_ex which would result in more duplicated code. The reference_gemm as a wrapper for cblas_gemm has less duplicated code, and should support all combinations of rocblas_gemm_ex through templates.

library/include/rocblas_bfloat16.h
AlexBrownAMD: The check on ~u.int32 & 0x7f800000 is still there, we just have the option of either rounding or truncating.

I think it makes more sense to have an option to truncate built into the type rather than in some parts of the test code. The operation is specific to this type, and rocblas was recently refactored to remove all independent specialized type conversion code in favor of simpler types and casting.

library/include/rocblas_bfloat16.h
AlexBrownAMD: Aha, didn't notice we had a tensile_bfloat16 as well, I can add this change to that too. As a side note, would it make sense to have a common shared library for things like this? The only difference looks to be the name (tensile_bfloat16 vs rocblas_bfloat16).

library/include/rocblas_bfloat16.h
AlexBrownAMD: Oops, you're right. I tried both versions, seeing which looked cleaner and forgot I left both in. Probably tidiest to have the rounding option in float_to_bfloat16 rather than the separate truncate function I think?

library/include/rocblas_bfloat16.h
leekillough: I still don't think we should be messing with `rocblas_bfloat16.h` (or `tensile_bfloat16.h`).

This truncation instead of rounding seems to be a corner case for specific kernels -- why are they truncating instead of rounding properly? Since it's a corner case for specific kernels and we just need to test them, the truncation function belongs in the tests and not in the global, public header files `rocblas_bfloat16.h` and `tensile_bfloat16.h`.

library/include/rocblas_bfloat16.h
leekillough: But this rounding is specific to certain kernels, and the only reason we're doing all of this is so that tests for certain sizes pass. As such, the truncation should go into the test code. We already have to modify the test code to account for the different sizes, and take special steps for BF16. We should write a truncation function in `clients/include/rocblas_math.hpp` or even in the same file as the test code, and call it.

We do not have functions for truncating instead of rounding for other types, so we should not introduce them for BF16, since then we have to introduce it for all types if we are to be consistent. Since this is a corner test case where a different kernel behavior is expected, the different rounding/truncation behavior belongs in the test code.

AlexBrownAMD: Moved the code to float_to_bfloat16_truncate function

clients/include/rocblas_math.hpp
```diff
@@ -24,6 +24,21 @@ inline __host__ float half_to_float(rocblas_half val)
...
+    union
+    {
+        float    fp32;
+        uint32_t int32;
+    } u = {val};
+    rocblas_bfloat16 ret;
+    ret.data = uint16_t(u.int32 >> 16);
+    return ret;
+}
+
```
leekillough:
```
inline __host__ rocblas_bfloat16 float_to_bfloat16_truncate(float val)
{
    union
    {
        float    fp32;
        uint32_t int32;
    } u = {val};
    rocblas_bfloat16 ret;
    ret.data = uint16_t(u.int32 >> 16);
    return ret;
}
```


You need to consider the lower bits in the `float` value. The above logic is not exactly correct, although it won't affect the test results at hand. This is correct (also adding `__device__`, to be more universal):
```
inline __host__ __device__ rocblas_bfloat16 float_to_bfloat16_truncate(float val)
{
    union
    {
        float    fp32;
        uint32_t int32;
    } u = {val};
    rocblas_bfloat16 ret;
    ret.data = uint16_t(u.int32 >> 16);
    if(~u.int32 & 0x7f800000 && u.int32 & 0xffff)
         ret.data |= 1; // Preserve signaling NaN
    return ret;
}
```

leekillough: Edit: The last comment had the reverse of the correct condition.

leekillough: You could also use:
```
inline __host__ __device__ rocblas_bfloat16 float_to_bfloat16_truncate(float val)
{
    union
    {
        float    fp32;
        uint32_t int32;
    } u = {val};
    rocblas_bfloat16 ret;
    ret.data = uint16_t(u.int32 >> 16);
    if((u.int32 & 0x7fff0000) == 0x7f800000 && u.int32 & 0xffff)
         ret.data |= 1; // Preserve signaling NaN
    return ret;
}
```
The main thing is that if the exponent bits are all `1`, and the mantissa has any `1`'s in it, then the value is `NaN`, and we need to preserve it. We do so by setting the lowest bit in the result mantissa if any of the 16 lower bits of the `float` were nonzero. We can be conservative, and only do this when the higher bits of the `float` mantissa are all `0`.

If the MSB of the `float` mantissa is `1`, then it's a quiet `NaN` and the lower bits are just payload and don't matter. If the MSB of the `float` mantissa is `0`, then if any of the lower bits are `1`, it indicates signaling `NaN`, and we have to preserve that by setting at least one bit below the MSB in the result mantissa.

PR #785, opened 2019-11-06
Disable header compilation tests when not testing

pfultz2:
This disables the header compilation tests since they dont work when we are building rocBLAS in our CI. This will only run when testing is enabled, which is set to on as default to keep the current behavior. We always build with `BUILD_TESTING=Off` since we dont want to build or run any tests.


daniellowell:
This fix would help us test newer commits of rocBLAS in our CI. Currently, our CI gets stuck building rocBLAS on newer releases. We would like to start incorporating more advanced features into MIOpen and this specific fix would help us do that.

PR #789, opened 2019-11-08
Revert "bf16 unit tests"

leekillough:
Reverts ROCmSoftwarePlatform/rocBLAS#783

This allows other PRs to be tested and merged.

PR #790, opened 2019-11-08
More specific restrictions on when to truncate in bf16 unit tests

AlexBrownAMD:
Fix broken bf16 unit tests.

There is one test case still failing:
pre_checkin/gemm_strided_batched_ex.blas3/bf16_rbf16_rbf16_rbf16_rf32_r_CN_512_512_512_1_512_512_0_512_512_3_262144_262144_262144
I think this is not related to #789 since it is testing strided batched


leekillough:
> I think this is not related to #789 since it is testing strided batched

All tests, including this still-failing BF16 strided_batched_ex, passed before I merged #779. 

This code seems very fragile, depending on rounding mode changing depending on size. Why was an engineering/management decision made to make these sized "replacement" kernels behave differently? It seems easier to make them round correctly, rather than to expect different rounding behavior at certain sizes.

AlexBrownAMD:
Not sure, but they do for now. My point was the truncate behaviour can only happen for gemm_ex, not for gemm_strided_batched_ex so the failing test case should still be rounding.

PR #792, opened 2019-11-08
Update tensile tag

amdkila:
This picks up the hotfix to disable assertion messages in the debug output.


leekillough:
SLES:
```
C++AMP runtime load error: libmcwamp_cpu.so.2.8: cannot open shared object file: No such file or directory
Can't load CPU runtime!
script returned exit code 255
```

bragadeesh (approved):

amdkila:
There's a problem with the sles base image that's unrelated to this change. I'm in the middle of trying to fix that.

PR #794, opened 2019-11-08
Added back dot tests, commented out failing dot host tests.

daineAMD:
Looks like batched/strided_batched dot tests were removed accidentally in 4589ef5e403e6fa9657277a649603b4d5d809e48.

Added back these tests. dot(c)_batched and dot(c)_strided_batched tests are failing on the host. They seem to only fail with smaller N sizes (they passed the initial PR with N ~= 33000) and batch_count > ~16 or so. Will create a ticket for this bug; I don't have time to look into it right now, hopefully someone else does. Commented out host-side tests for now.


daineAMD:
Thanks Torre, I'm guessing the issue is somewhere in the copying of results in rocblas_dot.hpp ~line 120 - 135, but not positive!

PR #797, opened 2019-11-09
Added additional device ID for Vega 10

amdkila:


sunway513:
Thank you @amdkila for the quick fix!

PR #807, opened 2019-11-11
packaging for tensile host library

wbgilmartin:
Added elements for packaging the tensile host "new client" libraries.

Updated CMakeFiles to enable packaging.
Updated tensile_host.cpp to find default install location.

### DIFF
--- a/library/src/tensile_host.cpp
+++ b/library/src/tensile_host.cpp
@@ -8,6 +8,7 @@
 
 #include "tensile_host.hpp"
 #include "rocblas.h"
+//#include <Tensile/AMDGPU.hpp>
 #include <Tensile/Contractions.hpp>
 #include <Tensile/EmbeddedLibrary.hpp>
 #include <Tensile/MasterSolutionLibrary.hpp>
@@ -23,6 +24,7 @@
 #include <memory>
 #include <string>
 #include <unistd.h>
+#include <sys/stat.h>
 
 // Return the value category for a value as a double precision value, such as whether it's 0, 1,
 // or some other value. Tensile uses a double precision value to express the category of beta.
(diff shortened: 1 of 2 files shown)


library/src/tensile_host.cpp
```diff
@@ -211,6 +213,31 @@ static auto GetTensileInputs(const RocblasContractionProblem<T, U, V>& problem)
     return inputs;
 }
 
+bool TestPath(std::string path)
+{
+  struct stat st;
+  if (stat(path.c_str(), &st) == 0)
```
leekillough: `access()` is probably better than `stat()`. Also, should use `const std::string& path`, like:

```
inline bool TestPath(const std::string& path)
{
    return access(path.c_str(), R_OK) == 0;
}
```

leekillough:
There are some `clang-format` errors. To fix them, run `clang-format -i <file>` on your `.cpp` file, or run `./install.sh -h` which will install `pre-commit` in `.git/hooks` and will automatically format all file types before commit.

I already merged your branch to my working branch, since I'm working rapidly on getting the new client working with GEMM_EX functions, which require runtime type dispatch.

library/src/tensile_host.cpp
```diff
@@ -235,12 +257,18 @@ struct TensileHostImpl : TensileHost
...
                 path.resize(strlen(path.c_str()));
-                path += "/../../Tensile/library";
             }
             else
             {
-                path = "/opt/rocm"; // TODO: Need to set default
+                path = "/opt/rocm/rocblas/lib"; 
             }
+	    // Find the location of the libraries
+            if (TestPath(path + "/../../Tensile/library"))
```
saadrahim: This assumes library is built in the same folder as the source. This is not necessarily true.

leekillough:
I already merged your changes with mine, which has another PR open, so I'm closing this.

PR #808, opened 2019-11-11
change to rocm-smi-lib64 package name

amcamd:
- Debian package names do not allow underscore '_' . Change to use new package name for rocm-smi.


amcamd:
rocBLAS PR#811 contains the same changes as in this PR.

PR #811, opened 2019-11-12
Update install.sh

pramenku:
rocm_smi64 is repalced with rocm-smi-lib64
Also, we don't need it as rocm-dkms install will take care already.
This way we can stop modifying install.sh whenever any change in rocm stack component.

resolves #___

Summary of proposed changes:
-  
-  
-

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -135,16 +135,16 @@ install_packages( )
   local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "pkg-config"
                                       "python2.7" "python3" "python-yaml" "python3-yaml"
                                       "llvm-6.0-dev" "libomp-dev"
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
@amdkila @amcamd can this be merged into master-hipclang branch? It is failing in HIP-Clang CI.

amdkila:
@aaronenyeshi yeah, I think hip-clang CI installs rock-dkms (i.e. w/o the kernel as it runs in docker) and thus needs rocm-smi-lib64 installed at some point. Maybe can install the rocm-smi-lib64 package in the dockerfile itself? In our Math Libs CI, we include it in the dockerfile rather than installing dependencies with the -d flag

PR #814, opened 2019-11-13
Update Tensile tag to pick up HPA fix

leekillough:
The new Tensile client code requires an Tensile fix which has gone back and forth over a few commits. This is the latest Tensile commit.

------

Notes from Tensile commit:

I just verified that if we keep the `TypedContractionInputs<Half, Half, Half, Half, float, float>` and this PR is used, then rocBLAS links okay, and the HPA tests pass, even with the new client.

I went through the trouble of making a workaround, namely, to use `TypedContractionInputs<Half, Half, Half, Half, Half, Half>` with a `Tensile::ContractionProblem` of types `Half` and `Half` for input and output. I converted the `float` `alpha` and `beta` values to `Half` before storing them in `TypedContractionInputs`. The result is a SegFault.

There does not seem to be any way of specifying the compute type explicitly -- I wonder if it's using heuristics. I guess the only way it tells HGEMM from HPA is the types of alpha and beta.

If we do not include this change, then the current rocBLAS will get a linker error when building for the new client.

This should not affect any code which is not using the new client, since all this does (besides call_once->static) is add an extra instantiation of `TypedContractionInputs`.


sdquiring:
HPA is specified with ContractionProblem's [`setHighPrecisionAccumulate()`](https://github.com/ROCmSoftwarePlatform/Tensile/blob/a89467f8e7627782acf5c5e24e82095cea39aa9d/Tensile/Source/lib/include/Tensile/ContractionProblem.hpp#L186) function.

I'm surprised that your test doesn't fail or crash since there's no case for `Half, Half, Half, Half, float, float` in `solve()`: https://github.com/ROCmSoftwarePlatform/Tensile/blob/a89467f8e7627782acf5c5e24e82095cea39aa9d/Tensile/Source/lib/source/ContractionSolution.cpp#L408

PR #815, opened 2019-11-13
Change pre-commit script and LICENSE.md to allow parsing of copyright dates

leekillough:
Simplify the pattern used to find files to reformat in `pre-commit`, making use of only one instance of `\.in$` instead of `\.cpp\.in$`, `\.hpp\.in$`, etc, since in this part of the script we're only removing whitespace at the end of lines and filling a missing newline at EOF (`clang-format` is done later with a different pattern).

Add `\.pl$` pattern for Perl scripts, and remove the Unicode copyright character from LICENSE.md, since it doesn't appear in any other files, and it makes the copyright dates harder to parse. This change also allows the `pre-commit` script to automatically adjust the copyright date range for `LICENSE.md`.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -27,7 +27,7 @@ for file in $(git diff-index --cached --name-only $against); do
 done
 
 # do the formatting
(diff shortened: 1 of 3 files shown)


amdkila:
This is failing due to the Tensile version update. It's looking for Tensile version 4.14 instead of 4.13, and then errors out.

leekillough:
Is there a way to see the test results?  I am running them independently, but they haven't finished yet (nightlies and all), and I don't have a GFX908 machine to run (I have to use `-lasm_ci`).

I realize people are on short schedule, and may need to merge PRs for others, but I have been on top of this all week.

amcamd:
I merged after the pre-checkin tests pass, so I am not sure if the nightly pass. I only have access to a gfx906 machine. I am running the nightly on the gfx906 machine.

PR #817, opened 2019-11-13
merge master into develop for ROCm 3.0

amcamd:


bump_develop_version.sh
```diff
@@ -5,14 +5,14 @@
...
+NEW_ROCBLAS_VERSION="2.13.0"
 
-OLD_TENSILE_VERSION="Tensile 4.12.0"
-NEW_TENSILE_VERSION="Tensile 4.13.0"
+OLD_TENSILE_VERSION="Tensile 4.14.0"
+NEW_TENSILE_VERSION="Tensile 4.14.0"
 
-OLD_TENSILE_HASH="fe4f721886d07eef6251cea4225e027181022aa5"
-NEW_TENSILE_HASH="a9379f4e42efb754c9a618047bfbf292d74dfd0f"
+OLD_TENSILE_HASH="78e72fbb35c3fba688deab3895f70f2f39f537af"
```
amdkila: Should old_tensile_hash and new_tensile_hash both be the same?

amcamd: I think this is good, as we want the develop branch and master branch to be the same, and they both have 78e72fbb35c3fba688deab3895f70f2f39f537af 
This is the tip of Tensile master branch

PR #818, opened 2019-11-14
rocBLAS changes to support reloctable ROCM installation

pruthvistony:
- Hard coded reference to /opt/rocm is parameterized
  to CMAKE_INSTALL_PREFIX
- Package is generated to install into ROCM_INSTALL_PATH if
  defined in the env otherwise default into /opt/rocm


TorreZuk:
I will assist with this PR, will be discussing with Pruthvi

header_compilation_tests.sh
```diff
@@ -38,13 +38,18 @@ out_uptodate()
     done
 }
 
-HCC=/opt/rocm/hcc/bin/hcc
+rocm_path=/opt/rocm
```
saadrahim: Why are we hard coding here? What if we install to /usr/local instead? Why are we assuming /opt/rocm?

TorreZuk: That is a default, you override with ROCM_PATH.   Not my PR but that is fine by me.  Same is done in install.sh

pruthvistony:
A new PR is raised on master-rocm-2.10.1 branch.
https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/835
So closing this PR.

PR #819, opened 2019-11-14
Remove rocm_smi64 from install.sh

amdkila:
Also reverts the commit that breaks the hip-clang build on CentOS


install.sh
```diff
@@ -122,6 +122,22 @@ install_packages( )
   local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "python2.7" "python3" "python-yaml" "python3-yaml" "hip_hcc" "pkg-config" )
   local library_dependencies_centos=( "epel-release" "make" "cmake3" "python34" "PyYAML" "python3*-PyYAML" "hip_hcc" "gcc-c++" "rpm-build" )
   local library_dependencies_fedora=( "make" "cmake" "python34" "PyYAML" "python3*-PyYAML" "hip_hcc" "gcc-c++" "libcxx-devel" "rpm-build" )
+  
+  local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "pkg-config"
```
aaronenyeshi: Why is there two of this same variable?

amdkila: Yeah, didn't deal with the merge conflict properly. This should be fixed now.

PR #820, opened 2019-11-14
Fix for SWDEV-210362

daineAMD:
Fixes SWDEV-210362.

- Avoid unnecessary hipMemCpys for strided_batched/batched versions of gemm_ex
- Remove change causing #batch_count calls to Tensile rather than a single strided_batched call for gemm_strided_batched_ex

These issues were introduced in d300cc1c3ecac1dae352829142851425b4849b9e and solved later in b87c84d6db25f123af8be38c7f48826db1be4298 where batched scalars were removed. This PR is a quick fix for removing batched scalars.

### DIFF
--- a/library/src/blas_ex/rocblas_gemm_ex.hpp
+++ b/library/src/blas_ex/rocblas_gemm_ex.hpp
@@ -915,8 +915,8 @@ rocblas_status gemm_ex_typecasting(rocblas_handle    handle,
                                    rocblas_stride    stride_d,
                                    rocblas_int       batch_count)
 {
-    Tc h_alpha[batch_count];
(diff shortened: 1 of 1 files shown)


amdkila:
The sles failure in CI is a known issue for rocm 2.10, and isn't unique to this PR. It has been fixed in the develop and master branches.

daineAMD:
Moving back to c725f1c since it seems to be improving performance and running QA benchmarks fine, while cc78913 is segfaulting with their benchmarks apparently.

daineAMD:
QA has informed me that everything looks good on their side. Was told to merge after QA verification so doing this now.

PR #824, opened 2019-11-19
blis generic configuration

TorreZuk:
Test to fix SWDEV-212579 making blis configuration generic so should run on all recent CPUs


TorreZuk:
Not going to merge for 2.10 branch as advised by Subodh

PR #829, opened 2019-11-20
Switch to prebuilt AMD blis 2.0

eidenyoshida:
Previously building blis 1.3


eidenyoshida:
This PR will allow tests on develop to run faster on both Intel and AMD CPUs. Potentially resolves ticket SWDEV-212579

PR #847, opened 2019-12-02
Limit Tensile_LOGIC by architecture

amdkila:
- This allows users to pass in the architecture they want to use when building rocBLAS
---> e.g. If on vega10, use gfx900 flag (-a gfx900) to only generate Vega 10 kernels
---> See this Tensile PR: https://github.com/ROCmSoftwarePlatform/Tensile/pull/742, it **needs to be merged in first**
- Also allows user to specify a build directory using --build-dir flag
- This works with both old and new Tensile clients
- This works in combination with the Tensile_LOGIC (-l flag)


CMakeLists.txt
```diff
@@ -172,7 +173,9 @@ if( BUILD_WITH_TENSILE )
   option( Tensile_PRINT_DEBUG "Tensile to print runtime debug info?" OFF )
 
   set( Tensile_TEST_LOCAL_PATH "" CACHE PATH "Use local Tensile directory instead of fetching a GitHub branch" )
-  set_property( CACHE Tensile_LOGIC PROPERTY STRINGS asm_full asm_lite hip_lite other )
+  set( CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories" )
+  set_property( CACHE Tensile_ARCHITECTURE PROPERTY STRINGS arcturus fiji vega10 vega20 _)
```
amcamd: Should this be gfx803, gfx900, gfx906, gfx908 in place of arcturus fiji vega10 vega20 ?

CMakeLists.txt
```diff
@@ -172,7 +173,9 @@ if( BUILD_WITH_TENSILE )
   option( Tensile_PRINT_DEBUG "Tensile to print runtime debug info?" OFF )
 
   set( Tensile_TEST_LOCAL_PATH "" CACHE PATH "Use local Tensile directory instead of fetching a GitHub branch" )
-  set_property( CACHE Tensile_LOGIC PROPERTY STRINGS asm_full asm_lite hip_lite other )
+  set( CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories" )
```
TorreZuk: I removed this as it is not tensile specific and this prefix is handled elsewhere for relocatable installations.

CMakeLists.txt
amdkila: Yes it should be, thanks for catching that. I've fixed it in my latest commit

CMakeLists.txt
amdkila: Thanks, I've removed it now. I created this branch before you made that change.

PR #848, opened 2019-12-02
SWDEV-213288 Potential Fix

daineAMD:
This is a potential fix for SWDEV-213288.

- In gemm, this allows for A = nullptr and B = nullptr when k == 0. In this case, we don't reference A or B so I don't believe there's a reason that A or B _need_ to be checked.
- We do not allow alpha to be nullptr in this case yet. This could be addressed in the future.
- Added a test ~~& changed bad_arg test name.~~

This should be able to fix SWDEV-213288 if we want to support this. Alternatively we could not change anything and just say for the future that nothing can be a nullptr in gemm when we are **not** quick-returning.

### DIFF
--- a/clients/gtest/gemm_gtest.cpp
+++ b/clients/gtest/gemm_gtest.cpp
@@ -87,6 +87,12 @@ namespace
         static std::string name_suffix(const Arguments& arg)
         {
             RocBLAS_TestName<gemm_test_template> name;
+
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
+            }
+
             name << rocblas_datatype2string(arg.a_type);
             bool isEx = (GEMM_TYPE == GEMM_EX || GEMM_TYPE == GEMM_BATCHED_EX
                          || GEMM_TYPE == GEMM_STRIDED_BATCHED_EX);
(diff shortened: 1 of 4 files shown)


clients/gtest/gemm_gtest.cpp
```diff
@@ -87,6 +87,12 @@ namespace
         static std::string name_suffix(const Arguments& arg)
         {
             RocBLAS_TestName<gemm_test_template> name;
+
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
```
YvanMokwinski: if (strstr(arg.function, "_bad_arg"))

daineAMD: Yeah good point. This also doesn't print quite as I intended (pre_checkin/gemm.blas3/_bad_argf16_r_NN_n1_n1_n1_1_n1_n1_0_n1). Just added it to run the added bad_arg check. Will fix or delay for a more fitting PR.

YvanMokwinski:
If k = 0, pointer validity of A and B  should not matter.
rocBLAS only does a quick returns if beta=1, otherwise it calls Tensile.

1/ perform C = beta * C in rocBLAS

2/ otherwise we assume Tensile performs this operation as expected and this should be tested for all the kernels.

The simplest would probably be 1/, but 2/ seems to already be the current behavior so it should be tested. That would prevent new Tensile generated kernels to diverge from this behavior.

PR #850, opened 2019-12-02
Adding more SUSE support

amdkila:
This PR ensures that if you're using opensuse, the install script will work. Right now, an error message pops up if the user attempts to build rocBLAS on OpenSUSE Leap distros, because the ID is opensuse-leap instead of sles.

See rocm-cmake PR: https://github.com/RadeonOpenCompute/rocm-cmake/pull/28


cmake/os-detection.cmake
```diff
@@ -7,6 +7,9 @@ function (get_os_id OS_ID)
   if (EXISTS "/etc/os-release")
     read_key("ID" _os_id)
   endif()
+  if (_os_id STREQUAL "opensuse-leap")
+    set(_os_id "sles")
```
saadrahim: We need some support added to rocm_cmake.

amdkila: https://github.com/RadeonOpenCompute/rocm-cmake/pull/28 Forgot to link this in the PR

amdkila:
For SLES, packaging will fail until rocm-cmake bug is fixed. I'll merge this in for now.

PR #851, opened 2019-12-03
Perf scripts

mahmoodw:
Addition of performance scripts to benchmark user defined problem sizes, using rocblas-bench.

### DIFF
--- a/scripts/performance/blas/alltime.py
+++ b/scripts/performance/blas/alltime.py
@@ -10,7 +10,6 @@ import re # regexp package
 import shutil
 import tempfile
 import rocblas_gentest as gt
-from loadData import compareDataSets
 import getspecs
 
 usage = '''A timing script for rocBLAS the generates plots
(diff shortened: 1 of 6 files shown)


scripts/performance/blas/alltime.py
```diff
@@ -0,0 +1,786 @@
+#!/usr/bin/python3
+
+import sys, getopt
+sys.path.append('../../../clients/common/')
```
mahmoodw: Note this path is hardcoded to reuse yaml scripts that are used by the client. If anyone has a suggestion for a better way to do this please let me know.

TorreZuk: I would think you could just keep a relative path and append to the current working directory to prefix when fetching yaml by file name.   Adding to the system path could have side effects down the road, but probably fine for now.   May need an environment variable for the full path if this python is invoked from different locations in CI and standalone.

PR #853, opened 2019-12-03
Removing unnecessary template parameter in rocblas_copy_template

YvanMokwinski:
Removing  a useless template parameter in rocblas_copy_template


YvanMokwinski:
written as failed but still testing ... 
and some failures are not related to my commit ...

PR #854, opened 2019-12-04
Tbmv fix

daineAMD:


YvanMokwinski:
Can u wait #853 to be merged first? Otherwise I will have again to pass all the test again because of this update.

daineAMD:
@YvanMokwinski Yup, I'll wait. Alternatively you could add this to your commit. I can add comments as Torre suggested if not.

YvanMokwinski:
I am just waiting some tests succeeding to merge. And it is so long ... and need to ignore sles.
Will be done in a couple of hours. Thanks.

PR #855, opened 2019-12-04
Adding system requirements to install documentation

daineAMD:
Currently there doesn't seem to be any documentation on system memory requirements to build rocBLAS. This PR may not be 100% accurate so discussion is welcome on what the requirements _actually_ are.

Andrew mentioned that he was able to successfully build on a Vega20 with 24GB.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -19,6 +19,13 @@ rocBLAS Debian packages can also be downloaded from the `rocBLAS releases tag <h
 Building from Source
 ********************
 
+Requirements
+````````````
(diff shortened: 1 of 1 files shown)


TorreZuk:
@saadrahim wondering if we can in the future flag some PRs as not requiring build or test stages?

saadrahim:
image
We can exclude the docs folder. Does anyone object?

TorreZuk:
Unless there is a documentation pipeline that gets run....

daineAMD:
Changed to 64GB, changed the wording to mention the --logic flag for ./install.sh. Will merge here soon if no one has any other recommendations.

PR #857, opened 2019-12-06
BLAS level 2 trmv routines (non batched, batched, strided batched)

YvanMokwinski:
SWDEV-211980

CLIENT SIDE /////////////////

-1/ modified:   clients/benchmarks/client.cpp
adding trmv headers and add references to trmv routines

-2/ modified:   clients/common/rocblas_gentest.py
adding default values of strides for the trmv strided batched routine

-3 /  modified:   clients/gtest/CMakeLists.txt
Reference the trmv_gtest.yaml file

-4 / modified:   clients/gtest/rocblas_gtest.yaml
Include the rocblas_trmv.yaml file

-5/  modified:   clients/include/cblas_interface.hpp
Add complex types for the cblas trmv routine.

-6 /  modified:   clients/include/flops.hpp
Add flop counts for the non-batched trmv routine.

-7/  modified:   clients/include/rocblas.hpp
add headers of the trmv routines.
 
-8 / modified:   clients/include/rocblas_template.yaml
Add trmv routines
   
-9/ modified:   clients/include/testing_logging.hpp
Add non batched trmv, why referencing batched routines with gemm only ?...

-10/   modified:   library/src/CMakeLists.txt
Adding the 3 cpp files to compile.
   
-12/     new: clients/gtest/trmv_gtest.cpp
Add trmv template for tests

-13/     new: clients/gtest/trmv_gtest.yaml
Definition of the tests

-14/  new: clients/include/testing_trmv.hpp
Testing the non batched  rocblas_trmv

-15/  new: clients/include/testing_trmv_batched.hpp
Testing the non batched  rocblas_trmv_batched

-16/ new: clients/include/testing_trmv_strided_batched.hpp
Testing the non batched  rocblas_trmv_strided_batched

 LIBRARY SIDE /////////

-1/ modified:   library/include/rocblas-functions.h
Headers and documentation of trmv routines.

-2/  new: library/src/blas2/rocblas_trmv.cpp
implementation C API + rocblas_trmv_impl
        
-3/ new: library/src/blas2/rocblas_trmv.hpp
definition of rocblas_trmv_template
        
-4/ new: library/src/blas2/rocblas_trmv_batched.cpp
implementation C API + rocblas_trmv_batched_impl
        
-5/ new: library/src/blas2/rocblas_trmv_batched.hpp
defintion of rocblas_trmv_batched_template     
    
-6/ new: library/src/blas2/rocblas_trmv_strided_batched.cpp
implementation C API +  rocblas_trmv_batched_impl

-7/  new: library/src/blas2/rocblas_trmv_strided_batched.hpp
defintion of rocblas_trmv_batched_strided_template     

KERNELS //////////

-8/ new: library/src/blas2/trmv_device.hpp
Kernel definitions, there is room for improvements.
How to remove the extra-memory? Would need a lot of synchronization.
The functionality is implemented, but far to be optimal.

-9/ new: library/src/blas2/trmv_template.hpp
Template definition for calling kernels defined in trmv_device.hpp.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -49,6 +49,9 @@
 #include "testing_ger.hpp"
 #include "testing_syr.hpp"
 #include "testing_tbmv.hpp"
+#include "testing_trmv.hpp"
+#include "testing_trmv_batched.hpp"
+#include "testing_trmv_strided_batched.hpp"
 #include "type_dispatch.hpp"
 #include "utility.hpp"
 #include <algorithm>
@@ -203,6 +206,9 @@ struct perf_blas<
                 {"gemv", testing_gemv<T>},
                 {"gemv_batched", testing_gemv_batched<T>},
                 {"gemv_strided_batched", testing_gemv_strided_batched<T>},
+                {"trmv", testing_trmv<T>},
+                {"trmv_batched", testing_trmv_batched<T>},
+                {"trmv_strided_batched", testing_trmv_strided_batched<T>},
                 {"ger", testing_ger<T>},
                 {"syr", testing_syr<T>},
                 {"tbmv", testing_tbmv<T>},
@@ -290,6 +296,7 @@ struct perf_blas<T,
                 {"iamax", testing_iamax<T>},
                 {"iamin", testing_iamin<T>},
                 {"gemv", testing_gemv<T>},
+                {"trmv", testing_trmv<T>},
                 {"tbmv", testing_tbmv<T>},
 #if BUILD_WITH_TENSILE
                 {"gemm", testing_gemm<T>},
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -297,6 +297,8 @@ struct perf_blas<T,
                 {"iamin", testing_iamin<T>},
                 {"gemv", testing_gemv<T>},
                 {"trmv", testing_trmv<T>},
+                {"trmv_batched", testing_trmv_batched<T>},
+                {"trmv_strided_batched", testing_trmv_strided_batched<T>},
                 {"tbmv", testing_tbmv<T>},
 #if BUILD_WITH_TENSILE
                 {"gemm", testing_gemm<T>},
--- a/clients/gtest/trmv_gtest.yaml
+++ b/clients/gtest/trmv_gtest.yaml
@@ -1,0 +1,156 @@
+---
+include: rocblas_common.yaml
+include: known_bugs.yaml
+
+Definitions:
+  - &small_matrix_size_range
+    - { M:    -1, lda:    1, stride_a:       1 }
+    - { M:     1, lda:    0, stride_a:       1 }
+    - { M:     0, lda:    1, stride_a:       1 }
+    - { M:     1, lda:    -1, stride_a:       1 }
+    - { M:    10, lda:   10, stride_a:       100 }
+    - { M:    1, lda:   -1, stride_a:       1 }
+    - { M:    10, lda:    2, stride_a:      20 }
+    - { M:   100, lda:  200, stride_a:    40000 }
+
+  - &medium_matrix_size_range
+    - { M:   300, lda:  400, stride_a:   160000 }
+    - { M:   600, lda:  601, stride_a:   400000 }
+
+  - &large_matrix_size_range
+    - { M:  1000, lda: 1000, stride_a:  1000000 }
+    - { M:  2000, lda: 2000, stride_a:  4000000 }
+    - { M:  4011, lda: 4011, stride_a: 16088200 }
+    - { M:  8000, lda: 8000, stride_a: 64000000 }
+
+  - &incx_range
+    - [2, -2,1,-1,0,3,10]
+
+
+Tests:
+- name: trmv_bad_arg
+  category: pre_checkin
+  function: trmv_bad_arg
+  precision: *single_double_precisions
+  uplo: [L, U]
+  transA: [N, T, C]
+  diag: [N, U]
+
+
+- name: trmv_small
+  category: quick
+  function: trmv
+  precision: *single_double_precisions_complex_real
+  uplo: [L, U]
+  transA: [N, T, C]
+  diag: [N, U]
+  matrix_size: *small_matrix_size_range
+  incx: *incx_range
+
+- name: trmv_medium
+  category: pre_checkin
+  function: trmv
+  precision: *single_double_precisions_complex_real
+  uplo: [L, U]
+  transA: [N, T, C]
+  diag: [N, U]
+  matrix_size: *medium_matrix_size_range
+  incx: *incx_range
+
+- name: trmv_large
+  category: nightly
+  function: trmv
+  precision: *single_double_precisions_complex_real
+  uplo: [L, U]
+  transA: [  N, T, C ]
+  diag: [N, U]
+  matrix_size: *large_matrix_size_range
+  incx: *incx_range
+
+- name: trmv_batched_bad_arg
+  category: pre_checkin
+  function: trmv_batched_bad_arg
+  precision: *single_double_precisions
+  uplo: [L, U]
+  transA: [  N, T, C ]
+  diag: [N, U]
+
+- name: trmv_batched_small
+  category: quick
+  function: trmv_batched
+  precision: *single_double_precisions
+  uplo: [L, U]
+  transA: [N, T, C]
+  diag: [N, U]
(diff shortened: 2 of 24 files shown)


clients/benchmarks/client.cpp
```diff
@@ -290,6 +296,7 @@ struct perf_blas<T,
                 {"iamax", testing_iamax<T>},
                 {"iamin", testing_iamin<T>},
                 {"gemv", testing_gemv<T>},
+                {"trmv", testing_trmv<T>},
```
daineAMD: Complex batched/strided_batched?

clients/include/flops.hpp
```diff
@@ -93,6 +93,25 @@ constexpr double scal_gflop_count<rocblas_double_complex, double>(rocblas_int n)
...
+template <typename T>
+constexpr double trmv_gflop_count(rocblas_int m)
+{
+    return ((m * (m + 1))) / 1e9;
+}
+
+template <>
+constexpr double trmv_gflop_count<rocblas_float_complex>(rocblas_int m)
+{
+    return (2.0 * (m * (m + 1))) / 1e9;
```
daineAMD: According to http://www.netlib.org/blas/cblas2time.f this should be 4 x m x m

library/src/blas2/trmv_template.hpp
```diff
@@ -0,0 +1,123 @@
+/* ************************************************************************
```
daineAMD: Hm, why are we splitting this into another file? Previously we've had rocblas_func.hpp define the rocblas_func_template() function which had the generic strided_batched/batched/non-batched implementation.

library/src/blas2/trmv_device.hpp
```diff
@@ -0,0 +1,205 @@
...
+__device__ void trmvn_kernel_calc(rocblas_fill     uplo,
+                                  rocblas_diagonal diag,
+                                  rocblas_int      m,
+                                  const T*         A,
+                                  rocblas_int      lda,
+                                  T*               x,
+                                  rocblas_int      incx,
+                                  T*               w)
+{
+    ptrdiff_t   tid = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
```
daineAMD: Just like with all the triangular kernels I've been writing, performance can definitely be improved, but that can be saved for a later date I'm sure.

clients/gtest/trmv_gtest.yaml
```diff
@@ -0,0 +1,156 @@
...
+  matrix_size: *large_matrix_size_range
+  incx: *incx_range
+  batch_count: [ 3 ]
+
+- name: trmv_strided_batched_bad_arg
+  category: pre_checkin
+  function: trmv_strided_batched_bad_arg
+  precision: *single_double_precisions
+  transA: N
+  uplo: [L]
```
TorreZuk: Can you add a bad uplo in here?   The documentation should say if 'rocblas_fill_full' does anything and for our testing we convert a character so should also catch a typo in the bad arg check.

clients/gtest/trmv_gtest.yaml
```diff
@@ -0,0 +1,156 @@
...
+  transA: [  T ]
+  diag: [N]
```
TorreZuk: Same for diag, a bad arg to test your checks for valid parameter.

library/include/rocblas-functions.h
```diff
@@ -3026,6 +3026,308 @@ ROCBLAS_EXPORT rocblas_status rocblas_zgemv_strided_batched(rocblas_handle
...
+/*! \brief BLAS Level 2 API
+
+    \details
+    trmv performs one of the matrix-vector operations
+
+         x = A*x or x = A**T*x,
+
+    where x is an n element vector and A is an n by n unit, or non-unit, upper or lower triangular matrix.
+
+    The vector x is overwritten on b.
```
TorreZuk: 'b' is referred to in multiple places but not defined.

clients/benchmarks/client.cpp
YvanMokwinski: gemv batched / strided batched?

clients/include/flops.hpp
YvanMokwinski: okay ... but if I count the operation (out of diagonal non unit or unit) should be m(m+1)/2 instead of m*m.
As a result, 2 * m * (m+1).

library/src/blas2/trmv_template.hpp
YvanMokwinski: That's what I have done, rocblas_trmv.hpp rocblas_trmv_stride_batched.hpp etc.
trmv_template is only a intermediate routine.

library/src/blas2/trmv_device.hpp
YvanMokwinski: Fully agree

clients/gtest/trmv_gtest.yaml
YvanMokwinski: Good point, will do

clients/gtest/trmv_gtest.yaml
YvanMokwinski: Same

library/include/rocblas-functions.h
YvanMokwinski: !

clients/benchmarks/client.cpp
daineAMD: Sure, those too. Just should add them so we can do benchmark tests for complex values for all functions that support it.

clients/include/flops.hpp
daineAMD: There are m(m+1)/2 complex multiplications, plus the same amount of complex additions. Each complex multiplication is 6 operations (4 multiplies and 2 adds) and each complex addition is 2 operations. This would be:
3m(m+1) + m(m+1) = 4m(m+1). I might be overlooking something though.

On netlib, _real_ gemv is calculated to have 2m*n + 2m operations, this is because of the multiply and the add, although I'm not sure where the extra 2m is coming from.

library/src/blas2/trmv_template.hpp
daineAMD: I see. I'm just not sure where we ended up with the conversation that was had a month or two ago about the "C++ interface", since most other routines don't have a C++ rocblas_batched_template/rocblas_strided_batched_template function I believe.
Not a big deal, just think consistency between routines is important. I know gemv is an outlier right now for having a gemv_device.hpp file. Iirc we started moving away from this, although I know there was some contention.

clients/include/flops.hpp
YvanMokwinski: Extra 2m is coming from operations on y.

Okay, so (m(m+1)/2 multiplications + (m(m+1))/2 additions, means that for the float double we are good. (Actually, it should be (m(m+1)/2) multiplications  + ((m-1)*m)/2 which would lead to m*m operations)

For the complex case, you are right the constant must be 4.
(or 4*(m(m+1)/2)+2*(m(m-1))/2) + 2*(m(m-1))/2  = 2m(m-1) + 2m(m+1) = 2m(2m) = 4*m*m.

So m*m and 4*m*m for complex types.
Thank you Dayne, it is clear now.

library/include/rocblas-functions.h
YvanMokwinski: See commit

library/src/blas2/trmv_device.hpp
YvanMokwinski: Should we create JIRA ticket for kernel optimizations? To keep track.

daineAMD: Yeah that's a good idea. I see there's some specific optimization tickets already (see http://ontrack-internal.amd.com/browse/SWDEV-172731).

clients/gtest/trmv_gtest.yaml
YvanMokwinski: Let's postpone it as none of these enum checks are performed. Will create a JIRA ticket for that.

clients/gtest/trmv_gtest.yaml
YvanMokwinski: Let's postpone it as none of these enum checks are performed. Will create a JIRA ticket for that.

clients/benchmarks/client.cpp
YvanMokwinski: Added

library/src/blas2/trmv_template.hpp
YvanMokwinski: Postponed discussion since it is a valid point.

PR #860, opened 2019-12-06
Tbmv batched

daineAMD:
- Adding tbmv_batched and tbmv_strided_batched.
- Fixed logging for regular tbmv so it logs whether on device_mode or host_mode

### DIFF
--- a/clients/include/testing_tbmv_batched.hpp
+++ b/clients/include/testing_tbmv_batched.hpp
@@ -226,11 +226,11 @@ void testing_tbmv_batched(const Arguments& arg)
                                     batch_count);
         }
 
-        gpu_time_used  = (get_time_us() - gpu_time_used) / number_hot_calls;
-        rocblas_gflops = batch_count * tbmv_gflop_count<T>(M, K) / gpu_time_used * 1e6;
-        rocblas_int k1 = K < M ? K : M;
-        rocblas_bandwidth
-            = (M * k1 - ((k1 * (k1 + 1)) / 2.0) + M) * sizeof(T) / gpu_time_used / 1e3;
+        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops    = batch_count * tbmv_gflop_count<T>(M, K) / gpu_time_used * 1e6;
+        rocblas_int k1    = K < M ? K : M;
+        rocblas_bandwidth = batch_count * (M * k1 - ((k1 * (k1 + 1)) / 2.0) + 2 * M) * sizeof(T)
+                            / gpu_time_used / 1e3;
 
         // only norm_check return an norm error, unit check won't return anything
         std::cout << "M,K,lda,incx,batch_count,rocblas-Gflops,rocblas-GB/s,us,";
--- a/clients/include/testing_tbmv_batched.hpp
+++ b/clients/include/testing_tbmv_batched.hpp
@@ -63,6 +63,11 @@ void testing_tbmv_batched_bad_arg(const Arguments& arg)
                                                   incx,
                                                   batch_count),
                           rocblas_status_invalid_handle);
+
+    // Adding test to check that if batch_count == 0 we can pass in nullptrs and get a success.
+    EXPECT_ROCBLAS_STATUS(
+        rocblas_tbmv_batched<T>(handle, uplo, transA, diag, M, K, nullptr, lda, nullptr, incx, 0),
+        rocblas_status_success);
 }
 
 template <typename T>
@@ -188,25 +193,44 @@ void testing_tbmv_batched(const Arguments& arg)
     {
         int number_cold_calls = 2;
         int number_hot_calls  = arg.iters;
-        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
 
         for(int iter = 0; iter < number_cold_calls; iter++)
         {
-            rocblas_tbmv_batched<T>(
-                handle, uplo, transA, diag, M, K, dA, lda, dx, incx, batch_count);
+            rocblas_tbmv_batched<T>(handle,
+                                    uplo,
+                                    transA,
+                                    diag,
+                                    M,
+                                    K,
+                                    dA.ptr_on_device(),
+                                    lda,
+                                    dx.ptr_on_device(),
+                                    incx,
+                                    batch_count);
         }
 
         gpu_time_used = get_time_us(); // in microseconds
 
         for(int iter = 0; iter < number_hot_calls; iter++)
         {
-            rocblas_tbmv_batched<T>(
-                handle, uplo, transA, diag, M, K, dA, lda, dx, incx, batch_count);
+            rocblas_tbmv_batched<T>(handle,
+                                    uplo,
+                                    transA,
+                                    diag,
+                                    M,
+                                    K,
+                                    dA.ptr_on_device(),
+                                    lda,
+                                    dx.ptr_on_device(),
+                                    incx,
+                                    batch_count);
         }
 
-        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
-        rocblas_gflops    = batch_count * tbmv_gflop_count<T>(M, K) / gpu_time_used * 1e6;
-        rocblas_bandwidth = batch_count * (1.0 * M * M) * sizeof(T) / gpu_time_used / 1e3;
+        gpu_time_used  = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops = batch_count * tbmv_gflop_count<T>(M, K) / gpu_time_used * 1e6;
+        rocblas_int k1 = K < M ? K : M;
+        rocblas_bandwidth
+            = (M * k1 - ((k1 * (k1 + 1)) / 2.0) + M) * sizeof(T) / gpu_time_used / 1e3;
 
         // only norm_check return an norm error, unit check won't return anything
         std::cout << "M,K,lda,incx,batch_count,rocblas-Gflops,rocblas-GB/s,us,";
--- a/clients/include/testing_tbmv_batched.hpp
+++ b/clients/include/testing_tbmv_batched.hpp
@@ -1,0 +1,245 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "cblas_interface.hpp"
+#include "flops.hpp"
+#include "near.hpp"
+#include "norm.hpp"
+#include "rocblas.hpp"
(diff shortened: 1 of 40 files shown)


clients/include/testing_tbmv_batched.hpp
```diff
@@ -0,0 +1,236 @@
...
+    rocblas_diagonal  diag        = char2rocblas_diagonal(char_diag);
+    rocblas_int       batch_count = arg.batch_count;
+
+    rocblas_local_handle handle;
+
+    // argument sanity check before allocating invalid memory
+    if(M < 0 || K < 0 || lda < M || lda < 1 || !incx || K >= lda || batch_count <= 0)
+    {
+        static const size_t     safe_size = 100; // arbitrarily set to 100
+        device_vector<T*, 0, T> dA1(safe_size);
```
TorreZuk: Consider device_batch_vector ?

clients/include/testing_tbmv_batched.hpp
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    size_t size_A = lda * size_t(M);
+    size_t size_x, abs_incx;
+
+    abs_incx = incx >= 0 ? incx : -incx;
+    size_x   = M * abs_incx;
+
+    // Naming: dK is in GPU (device) memory. hK is in CPU (host) memory
+    host_vector<T> hA[batch_count];
```
TorreZuk: ?  host_batch_vector

library/src/blas2/rocblas_tbmv.hpp
```diff
@@ -258,52 +258,46 @@ __global__ void tbmvx_kernel(rocblas_operation transA,
...
-                       1,
-                       m);
+    rocblas_status status = rocblas_copy_template<256>(
+        handle, m, x, offsetx, incx, stridex, x_copy, 0, 1, m, batch_count);
+
+    if(status != rocblas_status_success)
+        return status;
+
+    // in case of negative inc shift pointer to end of data for negative indexing tid*inc
+    offsetx = incx < 0 ? offsetx - ptrdiff_t(incx) * (m - 1) : offsetx;
```
TorreZuk: Sorry if this is a duplicate, but my comment got lost:  offsetx is a rocblas_int which may too small for the result of ptr_diff_t math.

clients/include/testing_tbmv_batched.hpp
daineAMD: Ah right, I forgot about the device_vector refactoring. Changed based off of testing_axpy_batched.hpp.

library/src/blas2/rocblas_tbmv.hpp
daineAMD: Good catch, fixed.

clients/include/testing_tbmv_batched.hpp
```diff
@@ -0,0 +1,245 @@
...
+            rocblas_error_1
+                = norm_check_general<T>('F', 1, M, abs_incx, batch_count, hx_gold, hx_1);
+        }
+    }
+
+    if(arg.timing)
+    {
+        int number_cold_calls = 2;
+        int number_hot_calls  = arg.iters;
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
```
YvanMokwinski: unnecessary

YvanMokwinski: there is no need for selecting pointer mode here

YvanMokwinski: Would be nice to use host_batch_vector and device_batch_vector only when testing batch routines.

clients/include/testing_tbmv_batched.hpp
```diff
@@ -0,0 +1,245 @@
...
+    host_vector<T> hx_gold[batch_count];
+
+    device_batch_vector<T> bA(batch_count, size_A);
+    device_batch_vector<T> bx(batch_count, size_x);
+
+    int last = batch_count - 1;
+    if((!bA[last] && size_A) || (!bx[last] && size_x))
+    {
+        CHECK_HIP_ERROR(hipErrorOutOfMemory);
+        return;
```
YvanMokwinski: No.
use CHECK_HIP_ERROR(bA.memcheck()) and CHECK_HIP_ERROR(bx.memcheck())

clients/include/testing_tbmv_strided_batched.hpp
```diff
@@ -0,0 +1,270 @@
+/* ************************************************************************
```
YvanMokwinski: batch_count == 0 ?
we should call the routine with null pointers and having success.

clients/include/testing_tbmv_strided_batched.hpp
```diff
@@ -0,0 +1,270 @@
...
+                                            lda,
+                                            dx,
+                                            incx,
+                                            stride_x,
+                                            batch_count);
+        }
+
+        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops    = batch_count * tbmv_gflop_count<T>(M, K) / gpu_time_used * 1e6;
+        rocblas_bandwidth = batch_count * (1.0 * M * M) * sizeof(T) / gpu_time_used / 1e3;
```
YvanMokwinski: :) 1.0

library/include/rocblas-functions.h
```diff
@@ -3143,6 +3143,274 @@ ROCBLAS_EXPORT rocblas_status rocblas_ztbmv(rocblas_handle                handle
...
+              the number of rows and columns of the matrix represented by each A_i.
+    @param[in]
+    k         [rocblas_int]
+              if uplo == rocblas_fill_upper, k specifies the number of super-diagonals
+              of each matrix A_i.
+              if uplo == rocblas_fill_lower, k specifies the number of sub-diagonals
+              of each matrix A_i.
+              k must satisfy k > 0 && k < lda.
+    @param[in]
+    A         device array to the first matrix A_i of the batch. Stores each banded triangular matrix A_i.
```
YvanMokwinski: nice

clients/include/testing_tbmv_batched.hpp
daineAMD: I think you're looking at an old commit, correct me if I'm wrong but I think this was fixed in 
33bf010.

YvanMokwinski: you are right Daine, my mistake.

YvanMokwinski:
- Can you remove the uncessary T template parameter?
This block:

    const T* A      = load_ptr_batch(Aa, hipBlockIdx_y, shifta, strideA);
    const T* x_copy = load_ptr_batch(xa_copy, hipBlockIdx_y, 0, m);
    T*       x      = load_ptr_batch(xa, hipBlockIdx_y, shiftx, stridex);

could be re-written with 

    const auto * A      = load_ptr_batch(Aa, hipBlockIdx_y, shifta, strideA);
    const auto* x_copy = load_ptr_batch(xa_copy, hipBlockIdx_y, 0, m);
    auto*       x      = load_ptr_batch(xa, hipBlockIdx_y, shiftx, stridex);

And normally, all the type deductions of T should be done by the compiler, no need to specify it explicitly. 

- Also, are x and x_copy of the same type ?
They should not have the same V abstract type.

clients/include/testing_tbmv_batched.hpp
daineAMD: Removed unnecessary pointer mode selection. testing_tbmv_batched should only reference host_batch_vectors and device_batch_vectors, not regular vectors now too.

clients/include/testing_tbmv_strided_batched.hpp
daineAMD: I added a test in testing_tbmv_(batched/strided_batched)bad_arg to check that we return success when batch_count == 0 and A and x are nullptrs. This was the easiest way to add them, although I'm not sure they completely fit in the "bad_arg" tests.

clients/include/testing_tbmv_strided_batched.hpp
daineAMD: This calc was wrong to begin with, should be fixed now if my math is right.

daineAMD:
Removed the T parameter, thanks.

x and x_copy in rocblas_tbmv_template _are_ of the same type, namely:
- T*, or
- T* const*
depending on batched version. Both these are modified so they cannot be const.

In tbmvx_kernel, however, x_copy can be a const, thus it is either:
- const T*, or
- const T* const*
While x remains the same type as previously.

clients/include/testing_tbmv_strided_batched.hpp
TorreZuk: That is good as I think we are under using bad_arg test, everything possible should be moved in there to reduce the test count.   Otherwise all these bad args cases are multiplied by all the other numeric parameter scans in the yaml.   You get the same code coverage without duplicating test runs by having them in bad_arg.
I don't suggest you do it but many of the invalid_size parameter values could likely be put in bad_arg parameters instead to reduce our test time.

YvanMokwinski: batch_count = 0 is not a bad argument. The routine returns success.

daineAMD: I don't think it makes any sense to add these tests to the "regular" test set, these are outliers and don't follow the pattern used to generate tests.
If we don't like adding successful tests to the bad_arg set, we could create a new test set, aside from the regular/bad_arg ones. Say, "quick_return" tests, or "nullptr_tests" or something similar? Or even "edge_cases" and add invalid sizes, N == 0 tests, nullptr tests, etc?

YvanMokwinski: Make more sense than putting it in bad args :D

daineAMD: I will leave it in here for now, adding this should be another PR. I'll start a thread so we can discuss adding another test set and what should be added to it.

clients/include/testing_tbmv_strided_batched.hpp
daineAMD: Updated after discussion with @YvanMokwinski and @mahmoodw.

Bandwidth = loads from A + loads from x
loads from x = m
loads from A = sigma(m-i) from i = 0 : k1
loads from A = m * k1 - ((k1 * (k1 + 1)) / 2) + m

therefore bandwidth = (m * k1 - ((k1 * (k1 + 1)) / 2) + 2*m
Where k1 is min(k, m) since when k > m we aren't using the extra values that exceed the size of the matrix.

daineAMD: Edit: I forgot the store to x, so add another m to get: therefore bandwidth = ( (m * k1 - ((k1 * (k1 + 1)) / 2) + 3*m ) * sizeof(T) / time

PR #862, opened 2019-12-09
Initial fix for logging multuthreaded corruption

leekillough:
Fixes SWDEV-215335

Right now it does not work, because the compiler/linker/runtime is inexplicably not calling `rocblas_create_handle()` from `rocblas_local_handle::rocblas_local_handle()`.

### DIFF
--- a/library/src/rocblas_auxiliary.cpp
+++ b/library/src/rocblas_auxiliary.cpp
@@ -32,6 +32,7 @@ extern "C" rocblas_pointer_mode rocblas_pointer_to_mode(void* ptr)
  ******************************************************************************/
 extern "C" rocblas_status rocblas_get_pointer_mode(rocblas_handle        handle,
                                                    rocblas_pointer_mode* mode)
+try
 {
     if(!handle)
         return rocblas_status_invalid_handle;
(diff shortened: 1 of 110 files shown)


library/src/rocblas_auxiliary.cpp
```diff
@@ -60,56 +58,52 @@ extern "C" rocblas_status rocblas_set_pointer_mode(rocblas_handle handle, rocbla
...
-    }
-    catch(...)
-    {
-        return rocblas_status_internal_error;
-    }
     return rocblas_status_success;
 }
+catch(...)
+{
+    return rocblas_status_memory_error;
```
YvanMokwinski: Are you sure that that it should be a rocblas_status_memory_error rather than rocblas_status_internal_error?

leekillough: It almost always catches a `std::bad_alloc` from the `new` when it receives an exception, so the memory error is more descriptive than internal error. I could catch        `std::bad_alloc` and flag it as memory error, while making other exceptions return internal error.

TorreZuk: Normally I think that would be good enough but for this case I think  two catch blocks might be better to discern the two failure modes.  Or two try catches.   I think we may want another error enum but not for this case, discuss later today.

leekillough: See #863 for a general solution to the exception issue. If you need to catch different exceptions and make them return a different `rocblas_status`, then change `library/src/include/utility.h`, which has a universal exception -> `rocblas_status` conversion function.

leekillough:
Will have to reopen in 2020

PR #863, opened 2019-12-09
In C routines, catch all C++ exceptions and convert them to rocblas_status

leekillough:
We currently allow C++ exceptions to escape into the rocBLAS C API, while we should only be returning `rocblas_status`. A C program has no way of handling the exceptions, which will always be fatal.

This change goes through every `extern "C"` routine returning `rocblas_status` and adds a _function_ ` try / catch(...)` block (_function_ meaning it's inserted outside the body of the function, to minimize merge conflicts), and calls an `exception_to_rocblas_status()` function to convert the current exception to `rocblas_status`.

C++11 added `std::exception_ptr`, `std::current_exception` and `std::rethrow_exception`, which allow us to catch any exception with `catch(...)`, and then process it inside another function with `std::current_exception()`, which can `std::rethrow_exception` the current exception inside another `try/catch` block, handling the mappings from C++ exceptions to `rocblas_status` all in one place.

In this change, if there is no current exception, `exception_to_rocblas_status()` returns `rocblas_status_success`. If there is a `std::bad_alloc` exception, it returns `rocblas_status_memory_error`. Otherwise, it returns `rocblas_status_internal_error`, because otherwise the exception would leak into the C API, which it is not prepared for.

Even though not all `extern "C"` routines currently throw exceptions, I have tried to add this `try / catch(...)` around all of the ones which return `rocblas_status`, so that exception leaks are not likely to occur. In the future, all `extern "C"` routines which return `rocblas_status` should use this paradigm.

There could be more clever ways of "wrapping" the C APIs with some kind of universal exception catcher and converter, but it's simpler and more readable simply to put `try / catch(...)` outside and around each separate `extern "C"` function.

### DIFF
--- a/library/src/include/utility.h
+++ b/library/src/include/utility.h
@@ -358,7 +358,11 @@ try
         std::rethrow_exception(e);
     return rocblas_status_success;
 }
-catch(std::bad_alloc)
+catch(const rocblas_status& status)
+{
+    return status;
+}
+catch(const std::bad_alloc&)
 {
     return rocblas_status_memory_error;
 }
--- a/library/src/include/utility.h
+++ b/library/src/include/utility.h
@@ -351,10 +351,9 @@ inline std::ostream& operator<<(std::ostream& os, rocblas_half x)
 // Convert the current C++ exception to rocblas_status
 // This allows extern "C" functions to return this function in a catch(...) block
 // while converting all C++ exceptions to an equivalent rocblas_status here
-inline rocblas_status exception_to_rocblas_status()
+inline rocblas_status exception_to_rocblas_status(std::exception_ptr e = std::current_exception())
 try
 {
-    auto e = std::current_exception();
     if(e)
         std::rethrow_exception(e);
     return rocblas_status_success;
--- a/library/src/include/utility.h
+++ b/library/src/include/utility.h
@@ -8,6 +8,7 @@
 #include "rocblas.h"
 #include <cmath>
 #include <complex>
+#include <exception>
 #include <hip/hip_runtime.h>
 #include <type_traits>
 
@@ -346,4 +347,25 @@ inline std::ostream& operator<<(std::ostream& os, rocblas_half x)
 {
     return os << float(x);
 }
+
+// Convert the current C++ exception to rocblas_status
+// This allows extern "C" functions to return this function in a catch(...) block
+// while converting all C++ exceptions to an equivalent rocblas_status here
+inline rocblas_status exception_to_rocblas_status()
+try
+{
+    auto e = std::current_exception();
+    if(e)
+        std::rethrow_exception(e);
+    return rocblas_status_success;
+}
+catch(std::bad_alloc)
+{
+    return rocblas_status_memory_error;
+}
+catch(...)
+{
+    return rocblas_status_internal_error;
+}
+
 #endif
--- a/library/src/blas1/rocblas_asum.cpp
+++ b/library/src/blas1/rocblas_asum.cpp
@@ -53,9 +53,14 @@ extern "C" {
 #define IMPL(name_, typei_, typeo_)                                                               \
     rocblas_status name_(                                                                         \
         rocblas_handle handle, rocblas_int n, const typei_* x, rocblas_int incx, typeo_* results) \
+    try                                                                                           \
     {                                                                                             \
         constexpr rocblas_int NB = 512;                                                           \
         return rocblas_asum_impl<NB>(handle, n, x, incx, results);                                \
+    }                                                                                             \
+    catch(...)                                                                                    \
+    {                                                                                             \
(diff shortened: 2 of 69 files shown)


library/src/include/utility.h
```diff
@@ -346,4 +347,25 @@ inline std::ostream& operator<<(std::ostream& os, rocblas_half x)
...
+        std::rethrow_exception(e);
+    return rocblas_status_success;
+}
+catch(std::bad_alloc)
+{
+    return rocblas_status_memory_error;
+}
+catch(...)
+{
+    return rocblas_status_internal_error;
```
YvanMokwinski: Is it really necessary to remove curly brackets?

YvanMokwinski:
Please include trmv

YvanMokwinski:
I would prefer 2 versions of exception_to_rocblas_status.

inline rocblas_status exception_to_rocblas_status(Exception&e or std::exception_ptr) { ... bla bla bla};
inline rocblas_status exception_to_rocblas_status() 
{ return exception_to_rocblas_status(std::current_exception); }

I should be able to use exception_to_rocblas_status without having to use the std::current_exception.
I also have a concern, is std::current_exception thread safe?

library/src/include/utility.h
leekillough: We are not _removing_ curly brackets.

If you're asking whether they are needed around a 1-line `catch` handler, [yes, they are](https://en.cppreference.com/w/cpp/language/try_catch). A `try` / `catch` block requires _compound statements_ for the `try` and `catch` clauses, which are brace-enclosed.

leekillough:
> Please include trmv

I can merge with latest.

library/src/include/utility.h
YvanMokwinski: You wrote:
void foo() try {} catch(..) {}

I am saying why not writing it like:
void foo() { try {} catch(..) {} }

leekillough:
> 
> 
> I would prefer 2 versions of exception_to_rocblas_status.
> 
> inline rocblas_status exception_to_rocblas_status(Exception&e or std::exception_ptr) { ... bla bla bla};
> inline rocblas_status exception_to_rocblas_status()
> { return exception_to_rocblas_status(std::current_exception); }
> 
> I should be able to use exception_to_rocblas_status without having to use the std::current_exception.
> I also have a concern, is std::current_exception thread safe?

Yes, `std::current_exception` is thread-safe.

Your request can be done simply with one function and default argument:

```
inline rocblas_status exception_to_rocblas_status(std::exception_ptr e = std::current_exception())
{
  ...
}
```
This preserves the old behavior while allowing you to pass `std::exception_ptr` directly if you wish. `std::exception_ptr` is a lightweight smart-pointer which can be copied and passed around.

YvanMokwinski:
> > I would prefer 2 versions of exception_to_rocblas_status.
> > inline rocblas_status exception_to_rocblas_status(Exception&e or std::exception_ptr) { ... bla bla bla};
> > inline rocblas_status exception_to_rocblas_status()
> > { return exception_to_rocblas_status(std::current_exception); }
> > I should be able to use exception_to_rocblas_status without having to use the std::current_exception.
> > I also have a concern, is std::current_exception thread safe?
> 
> Yes, `std::current_exception` is thread-safe.
> 
> Your request can be done simply with one function and default argument:
> 
> ```
> inline rocblas_status exception_to_rocblas_status(std::exception_ptr e = std::current_exception())
> {
>   ...
> }
> ```
> 
> This preserves the old behavior while allowing you to pass `std::exception_ptr` directly if you wish. `std::exception_ptr` is a lightweight smart-pointer which can be copied and passed around.

I like this

leekillough:
Please excuse the whitespace-only changes; there were caused by pre-commit script being run before commit. It removes trailing whitespace at the ends of lines.

When you run install.sh, the pre-commit script gets symlinked in your .git directory so that it will run before your commits.

library/src/include/utility.h
leekillough: > 
> 
> You wrote:
> void foo() try {} catch(..) {}
> 
> I am saying why not writing it like:
> void foo() { try {} catch(..) {} }

Because the former is _function_ try-block, and thus it does not interfere with the indentation of the code -- all we're saying is _"try this function, exactly as it was already written, and if it raises any exceptions, handle them in blocks after the function. You do not need to reindent your code, and the extra code for handling exceptions will be at the end of the function and thus not a distraction from the main part of the code."_

[Function try blocks](https://en.cppreference.com/w/cpp/language/function-try-block) are mainly for constructors, since there's no other way to catch exceptions thrown during base or member initialization lists (by the time you're in the main body of the constructor, any exceptions raised during initialization would have escaped). They are also good for this use-case, where we want to stop any exceptions from escaping our C API, but we don't write our code thinking mainly about the exceptions, or organizing our code indentation around them.

PR #868, opened 2019-12-12
Catch signals as failures in rocblas-test, print out Tensile client version, fix HPA in new Tensile client

leekillough:


leekillough:
Do not merge until I am finished running tests offline (not CI)

leekillough:
The old client tests pass.
```
[==========] 877803 tests from 227 test cases ran. (18452040 ms total)
[  PASSED  ] 877803 tests.
```
The new client tests fail in integer8 and some others, but are much better than before.

PR #872, opened 2019-12-12
Disabling build testing

saadrahim:
hip-clang has issues with the BUILD_TESTING section. It is disabled with this PR.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -205,11 +205,16 @@ rocm_export_targets(
 
 rocm_install_symlink_subdir( rocblas )
 
+#Control for enabling or disabling header testing
+option (RUN_HEADER_TESTING ON)
+
+if(BUILD_TESTING AND RUN_HEADER_TESTING)
 # Compilation tests to ensure that header files work independently,
 # and that public header files work across several languages
 add_custom_command(
   TARGET rocblas
   POST_BUILD
   COMMAND ${CMAKE_HOME_DIRECTORY}/header_compilation_tests.sh
-  WORKING_DIRECTORY ${CMAKE_HOME_DIRECTORY}
+  WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
  )
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,8 @@ option( BUILD_VERBOSE "Output additional build information" OFF )
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build rocBLAS as a shared library" ON )
 
+option( BUILD_TESTING "Build tests for rocBLAS" ON )
+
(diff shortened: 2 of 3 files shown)


leekillough:
Looks like CI is broken by this change -- CI needs to explicitly turn on BUILD_TESTING:
```
+ cd 1vSe3BbYj/rocblas/build/release/clients/staging
+ LD_LIBRARY_PATH=/opt/rocm/lib
+ sudo ./example-sscal
N        rocblas(us)     
10240    10092.00
SSCAL TEST PASSES
+ LD_LIBRARY_PATH=/opt/rocm/lib
+ GTEST_LISTENER=NO_PASS_LINE_IN_LOG
+ sudo ./rocblas-test --gtest_output=xml --gtest_color=yes '--gtest_filter=*quick*:*pre_checkin*-*known_bug*'
sudo: ./rocblas-test: command not found
script returned exit code 1
```

leekillough:
This change seems to be too big. The flag it disables disables all rocBLAS tests, not just header_compilation_tests.sh. 

Another possible change:

In install.sh (if necessary):

export build_hip_clang

At the top of header_compilation_tests.sh:

  [[ "${build_hip_clang}" == true ]] && exit

However, this introduces a dependency on install.sh.

Better would be to use a CMake variable exported to the environment, for detecting hip-clang builds in header_compilation_tests.sh.

Or you can, in CMake, not call header_compilation_tests.sh if CMake knows it's a hip-clang build. Look for header_compilation_tests.sh in CMakeLists.txt and change the logic there.

Disabling all testing code from hip-clang builds in install.sh is overkill.

I'm not in front of a computer (this is over my phone in a hospital).

saadrahim:
I made the changes and will fix test. Not committing until tested.

library/src/CMakeLists.txt
```diff
@@ -205,11 +205,16 @@ rocm_export_targets(
 
 rocm_install_symlink_subdir( rocblas )
 
+#Control for enabling or disabling header testing
+option (RUN_HEADER_TESTING ON)
```
pfultz2: I believe `option` should have the documentation first and then the initial value:

https://cmake.org/cmake/help/latest/command/option.html

If the documentation is missing, then this is treated as `Off`.

library/src/CMakeLists.txt
```diff
@@ -205,11 +205,16 @@ rocm_export_targets(
...
+
+if(BUILD_TESTING AND RUN_HEADER_TESTING)
```
pfultz2: Why not just set `RUN_HEADER_TESTING` default based on `BUILD_TESTING`: `option (RUN_HEADER_TESTING "" ${BUILD_TESTING})`? Then you can just check `RUN_HEADER_TESTING` here instead of both.

leekillough:
As soon as Paul's suggestions are addressed, I'll approve this ASAP.

leekillough:
Please see PR https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/885

PR #873, opened 2019-12-16
[2.10-hipclang] replace hip_hcc with rocm-dev

aaronenyeshi:
2.10-hipclang requires dependency on rocm-dev instead of hip_hcc.


aaronenyeshi:
@saadrahim I will merge once I confirm that our next build will have these package naming changes

PR #874, opened 2019-12-16
Replace hip_hcc with rocm-dev for HIP-Clang VDI

aaronenyeshi:
Switch to using rocm-dev instead of hip_hcc package dependency.


aaronenyeshi:
This is safe to merge anytime.

PR #878, opened 2019-12-18
Fix Usage of _Float16 to clang 9 or later

leekillough:
May need to be cherry-picked and applied to master, to fix SWDEV-216454

### DIFF
--- a/library/include/rocblas-types.h
+++ b/library/include/rocblas-types.h
@@ -48,7 +48,7 @@ typedef double rocblas_double;
 // Clang supports _Float16 on C11 and C++11
 // GCC does not currently support _Float16 on amd64
 /*! \brief Represents a 16 bit floating point number. */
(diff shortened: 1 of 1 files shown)


leekillough:
> 
> 
> Could be (defined(**clang_major**) && **clang_major** >= 9) && ... ?
> The later && tests may be redundant now as well.

The C preprocessor automatically defaults undefined macros as `0` in `#if` directives. I.e., if a macro is undefined, it will be treated as `0`.

The `&&` clause is because the clang compiler only supports `_Float16` in language levels C++11 and C11 and later.

TorreZuk:
Okay that is fine, just quoting Starfleet regulation / MISRA C:2004, 19.11  "All macro identifiers in preprocessor directives shall be defined before use, except in #ifdef and #ifndef preprocessor directives and the defined() operator. "  :-)

PR #883, opened 2019-12-20
New client for build

wbgilmartin:
This will decouple the old client components from the new client builds when BUILD_WITH_TENSILE_HOST=ON. 

The builds that I ran (without the gtests) took 20 minutes to complete. (9 minutes for the tensile part for the full asm_full logic) and 11 minutes for the rocBLAS build. 

The gmake test will take an extra 5 minutes or so.


library/src/handle.cpp
```diff
@@ -18,12 +19,13 @@
 _rocblas_handle::_rocblas_handle()
 {
 #if BUILD_WITH_TENSILE
-    static int dummy = (tensileInitialize(), 0);
```
leekillough: `tensileInitialize()` still needs to be called in new client, right?

leekillough:
I will attempt a local build. I also have some changes to `tensile_host.`*.

library/src/handle.cpp
leekillough: I will move `tensileInitialize()` to `tensile_host.cpp` so that all references to Tensile are in `tensile_host.cpp` in the new Tensile client.

PR #893, opened 2019-12-23
New tensile client 

leekillough:
This is more development and cosmetic cleanup of the new Tensile client in rocBLAS; we still need to figure out the Int8 failures and get other things settled before this can be merged, but this has been started because of the January 8 deadline.


leekillough:
The large diffs are because pending PRs have already been merged into this branch.

library/src/tensile_host.cpp
```diff
@@ -178,63 +172,70 @@ namespace
...
+     *************************************************************************/
+    template <typename Ti, typename To = Ti, typename Tc = To>
+    struct AlphaBeta
     {
-        using type = typename rocblas_to_tensile_type<Tc>::type;
-        static void copy(type* dst, const Tc* src)
+        using tensile_type = typename rocblas_to_tensile_type<Tc>::tensile_type;
+        static void copy(tensile_type* dst, const Tc* src)
         {
             static_assert(sizeof(*src) == sizeof(*dst),
```
TorreZuk: Using the types instead of *pointers it would be clearer.

library/src/tensile_host.cpp
```diff
@@ -405,30 +409,49 @@ TensileHost* createTensileHost()
...
-    auto  inputs          = GetTensileInputs(problem);
-    auto  solution        = host->library->findBestSolution(tensile_problem, *host->hardware);
-    if(!solution)
+    std::shared_ptr<Tensile::ContractionSolution> solution;
+    rocblas_status                                status = rocblas_status_internal_error;
+
+    try
+    {
+        // We know that the TensileHost instance is a TensileHostImpl, so we can downcast to it
+        auto* host            = static_cast<TensileHostImpl*>(this);
```
TorreZuk: The base class is TensileHost so this upcast should use dynamic_cast which would correctly fail if you change the classes.

library/src/tensile_host.cpp
leekillough: > Using the types instead of *pointers it would be clearer.

I disagree, because if you use types, you have to name the type in two places, while if you use the pointers, you're always sure you're using the type that it was declared as.

[Reference](https://wiki.sei.cmu.edu/confluence/display/c/EXP09-C.+Use+sizeof+to+determine+the+size+of+a+type+or+variable?focusedCommentId=88016344#comment-88016344)

library/src/tensile_host.cpp
leekillough: We do not need to use dynamic_cast.

These are not polymorphic classes.

These are PIMPL classes, which can use static_cast for better performance and smaller code.

Note that you are reviewing code that's already checked into develop; this PR just refined some of it.

This will not be changed in future releases. This design is intentional.

leekillough:
This is being folded into another PR

PR #894, opened 2019-12-23
Make tensile_helper explicit instantiations inline 

leekillough:
The `tensile_helper()` function has explicit instantiations, which means you'll get multiple definitions if you put them in a header file that's included in multiple places.

The reason it doesn't occur in rocBLAS is that rocBLAS includes the header only once, and thus only gets one definition.

If rocSOLVER uses rocBLAS headers, it will create multiple definitions.

We simply declare the instantiations `inline` to get around the One Definition Rule.

`tensile_helper()` will be removed when we move to the new Tensile client.


leekillough:
This is causing rocSOLVER build failures and needs to be fixed ASAP

leekillough:
> 
> 
> We could have these instantiations in a cpp file and thus not being worried about multiple instances.

The functions will be removed with the new Tensile client; this is just a temporary fix for the old client.

PR #897, opened 2020-01-01
Catch signals and exceptions as failures

leekillough:
Add the rest of the rocBLAS function tests to the tests which catch signals as failures

Add uncaught exceptions as failures, just like signals, since rocBLAS is a C API which shouldn't throw C++ exceptions.

Implement wrapper as a function taking lambda expressions, instead of a large macro.

Update test documentation on the new functionality.


clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -114,53 +118,108 @@ class ConfigurableEventListener : public TestEventListener
...
-    sigaction(SIGSEGV, &act, nullptr);
-    sigaction(SIGSYS, &act, nullptr);
-    sigaction(SIGUSR1, &act, nullptr);
-    sigaction(SIGUSR2, &act, nullptr);
+    sigfillset(&act.sa_mask);
+    act.sa_flags   = 0;
+    act.sa_handler = rocblas_test_signal_handler;
+
+    for(int sig :
+        {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGPIPE, SIGQUIT, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
```
TorreZuk: Exclude SIGQUIT just like SIGINT as user driven?

leekillough: > Exclude SIGQUIT just like SIGINT as user driven?

Done.

PR #904, opened 2020-01-06
Adding an option to skip creating ld.so.conf entry

pruthvistony:
This option will be used when it is build by the jenkins.
This change will enable users to select proper rocblas lib when multiple versions are installed.


pruthvistony:
I have updated the code with GETOPT_PARSE, also corrected tensile-host parameter.
Please review

PR #907, opened 2020-01-07
Allow known_bugs to be platform-specific

leekillough:
This allows known_bugs to be made platform-specific.

It was motivated by vega 10 - only failures.

Basically, a `known_bug_platforms` key, as a space, comma, or colon separated list of platforms, needs to be added to each known bug, indicating the platforms which it applies to. If `known_bug_platforms` isn't defined, then the bug is assumed to apply to all platforms.


leekillough:
> The CI failures are due to disk space and an unrelated packaging issue. This should be safe to merge.

The Python code had to be massaged to get it to work right. I also added the Vega 10 known bugs to `known_bugs.yaml`, with a `known_bug_platforms: "gfx900"`  platform selector.

Going forward, we can specify platforms in `known_bugs.yaml`.

It is finally ready for full review and testing.

PR #917, opened 2020-01-10
Add additional device ID for arcturus

amdkila:


leekillough:
> 
> 
> No need to revert them now. When we next update them, we can revert as needed. You'll notice that since your last update, some logic files had been updated already, causing merge conflicts.

That's probably the reformatting changes. The pre-commit script should handle them automatically if you revert to the old files and then commit them. You don't have to merge conflicts if they were just caused by pre-commit.

amdkila:
> I changed at least 3 arcturus logic files a couple of days ago.  We cannot revert blindly.  It is unsafe.  If you absolutely want to revert, you need to take the current logic files and convert them back to the compact form.

Yep, you're right. I am in the process of doing that- after reverting, there are 4 conflicting files, and I'm dealing with the second file. At that point, there will be another PR made with those changes, then I can use this one to add the IDs. I'll be sure to add you as a reviewer to the 2nd PR.

amdkila:
Can close this now as PR #918 now has the changes.

PR #919, opened 2020-01-10
Lwpmlse 16 argument model

TorreZuk:
open for discussion and requirements review for Lee's logging refactoring

### DIFF
--- a/clients/common/rocblas_arguments.cpp
+++ b/clients/common/rocblas_arguments.cpp
@@ -55,7 +55,7 @@ void ArgumentModel::log_perf(std::stringstream& name_line,
         val_line << cblas_gflops << "," << cpu_us << ",";
         if(arg.norm_check)
         {
-            name_line << "norm1,norm2";
+            name_line << "norm1,norm2,";
             val_line << norm1 << "," << norm2 << ",";
         }
     }
--- a/clients/common/rocblas_arguments.cpp
+++ b/clients/common/rocblas_arguments.cpp
@@ -1,0 +1,62 @@
+/* ************************************************************************
+ * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include "rocblas_arguments.hpp"
+#include <cinttypes>
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <type_traits>
+
+//
+//
+
+ArgumentModel::ArgumentModel(const std::vector<arg_type>& params)
+    : m_args(params)
+{
+}
+
+bool ArgumentModel::hasParam(arg_type a)
+{
+    return std::find(m_args.begin(), m_args.end(), a) != m_args.end();
+}
+
+void ArgumentModel::log_perf(std::stringstream& name_line,
+                             std::stringstream& val_line,
+                             const Arguments&   arg,
+                             double             gpu_us,
+                             double             gflops,
+                             double             gpu_bytes,
+                             double             cpu_us,
+                             double             norm1,
+                             double             norm2)
+{
+    int batch_count = hasParam(e_batch_count) ? arg.batch_count : 1;
+    int hot_calls   = arg.iters < 1 ? 1 : arg.iters;
+
+    // per/us to per/sec *10^6
+    double rocblas_gflops = gflops * batch_count * hot_calls / gpu_us * 1e6;
+    double cblas_gflops   = gflops * batch_count / cpu_us * 1e6;
+
+    // bytes/us to GB/s = 10^6 * 10^-9 = /10^3
+    double rocblas_GBps = gpu_bytes * batch_count / gpu_us / 1e3;
+
+    name_line << "rocblas-Gflops,rocblas-GB/s,rocblas-us,";
+    val_line << rocblas_gflops << "," << rocblas_GBps << "," << gpu_us << ",";
+
+    if(arg.unit_check || arg.norm_check)
+    {
+        name_line << "CPU-Gflops,CPU-us,";
+        val_line << cblas_gflops << "," << cpu_us << ",";
+        if(arg.norm_check)
+        {
+            name_line << "norm1,norm2";
+            val_line << norm1 << "," << norm2 << ",";
+        }
+    }
+}
--- a/clients/common/rocblas_arguments.cpp
+++ b/clients/common/rocblas_arguments.cpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright 2018-2019 Advanced Micro Devices, Inc.
+ * Copyright 2018-2020 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
 #include "rocblas_arguments.hpp"
@@ -55,7 +55,7 @@ void ArgumentModel::log_perf(std::stringstream& name_line,
         val_line << cblas_gflops << "," << cpu_us << ",";
         if(arg.norm_check)
         {
-            name_line << "norm1,norm2,";
+            name_line << "norm_error_host_ptr,norm_error_device_ptr,";
             val_line << norm1 << "," << norm2 << ",";
         }
     }
--- a/clients/include/rocblas_arguments.hpp
+++ b/clients/include/rocblas_arguments.hpp
@@ -214,6 +214,12 @@ private:
         str << x;
     }
 
+    // Float16 output
+    static void print_value(std::ostream& str, rocblas_half x)
+    {
+        str << float(x);
+    }
+
     // Floating-point output
     static void print_value(std::ostream& str, double x)
     {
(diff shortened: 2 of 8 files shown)


YvanMokwinski (approved):

clients/include/rocblas_arguments.hpp
```diff
@@ -324,4 +328,216 @@ static_assert(std::is_trivial<Arguments>{},
...
+    std::stringstream value_list;
+    const char        delim = ',';
+
+    auto print = [&](const char* const name, auto x) mutable {
+        name_list << name << delim;
+        print_value(value_list, x);
+        value_list << delim;
+    };
+
+#define CASE(x)                  \
```
leekillough: These macros and the `print` lambda are unnecessarily complicated. Just define a variadic template function and pass names and values:
```
print_values("M", M, 
             "N", N,
             "K", K, 
             );
```

There are examples in `logging.h`. I plan to generalize them to support different formats.

clients/include/rocblas_arguments.hpp
```diff
@@ -324,4 +328,216 @@ static_assert(std::is_trivial<Arguments>{},
...
+                          const Arguments&   arg,
+                          double             gpu_us,
+                          double             gflops,
+                          double             gpu_bytes,
+                          double             cpu_us,
+                          double             norm1,
+                          double             norm2);
+
+protected:
+    std::vector<arg_type> m_args;
```
leekillough: This adds a lot of extra overhead, when it would be simpler to use composition/inheritance and CTRP with `Arguments` instead of a dynamically allocated `vector`.

leekillough: If you want to print out a specific argument set, it's best to list name, value pairs in an arbitrary-length print function call, making adjustments to names and values (such as alpha, beta) as necessary.

leekillough: A `std::tuple`, listing names and values/references, is just a container representation of it.

clients/include/rocblas_arguments.hpp
```diff
@@ -324,4 +328,216 @@ static_assert(std::is_trivial<Arguments>{},
...
+CTOKEN(stride_b);
+CTOKEN(stride_c);
+CTOKEN(stride_d);
+CTOKEN(stride_x);
+CTOKEN(stride_y);
+CTOKEN(algo);
+CTOKEN(solution_index);
+#undef CTOKEN
+
+class ArgumentModel
```
leekillough: We should not need to define another class for a "16-argument model". Then we have two `Arguments` classes to maintain, instead of one universal class. Right now so much of the rocBLAS testing and benchmarking code, both C++ and Python, is wound up in the `Arguments` structure, that introducing a new class would not be good.

Better to add custom functions/methods which operate on the universal `Arguments` class.

clients/common/rocblas_arguments.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+ArgumentModel::ArgumentModel(const std::vector<arg_type>& params)
+    : m_args(params)
+{
+}
+
+bool ArgumentModel::hasParam(arg_type a)
+{
+    return std::find(m_args.begin(), m_args.end(), a) != m_args.end();
+}
```
leekillough: Linear search should be avoided.

clients/include/rocblas_arguments.hpp
```diff
@@ -324,4 +328,216 @@ static_assert(std::is_trivial<Arguments>{},
               "Arguments is not a trivial type, and thus is "
               "incompatible with C.");
 
+enum arg_type
```
leekillough: An `enum` with another set of names to maintain, makes maintenance harder.

Better to use `std::tuple` to list name-value pairs, and pass the tuples to a function which prints the elements out. Then if a particular benchmark needs a particular set of `Arguments`, it just lists them by name, value pairs, adjusting the values as necessary, and passes them to a universal print routine. There's no need to maintain a `vector` of arguments, a list of `enums` of arguments, a list of macros for generate `CASE`, etc. Just pass name-value pairs to a function with formats the output you way you want. That's what we do in logging today.

leekillough: Or if you really want 100% dynamic name lookup, without any duplication of "name", name, then create a separate `Arguments` data structure which uses `std:map` or `std::unordered_map` to map names to fields in `Arguments`. The creation of the map, as well as other ordered operations on "name", value pairs, can be automated in an `Arguments` and can work to construct a `std::map` or to print out elements -- it simply creates a `std::tuple` of every "name", value pair in `Arguments`, which can be then be printed out or transformed into a `std::map`.

leekillough (changes requested):
Significantly deviates from rocBLAS `Arguments` model.

Will have to be refactored. I will do it.

clients/include/rocblas_arguments.hpp
TorreZuk: The list of arguments is not known at compile time, only run time, thus the switch.   The lambda is a simple functor much like the one pre-existing in the base class, to me a variadic template is more complicated.

clients/include/rocblas_arguments.hpp
TorreZuk: This is for test setup, so a single static instance can be created per function so almost zero relative overhead compared to other things.

clients/include/rocblas_arguments.hpp
TorreZuk: This "model" is a class for a different purpose, thus a different class.   I don't follow your argument we can discuss on Monday.

clients/common/rocblas_arguments.cpp
TorreZuk: maximum 16 integers, average 8.

clients/include/rocblas_arguments.hpp
TorreZuk: This class is to eliminate the typing of strings and argument ordering, using just the enums ensure no typos.   The logging is also a form of duplication which could be potentially eliminated and handled by a helper class(es).

clients/include/rocblas_arguments.hpp
leekillough: > This is for test setup, so a single static instance can be created per function so almost zero relative overhead compared to other things.

Then this should be in another tool, such as `rocblas_gentest.py`.

clients/include/rocblas_arguments.hpp
leekillough: This should not be used to create tools. They should be created outside of rocBLAS.

The code took 4 months to refactor to use `Arguments` and a YAML source model; this significantly deviates from that design.

Besides, it should not go into `rocblas_arguments.hpp`. I have already moved it out of that file, and am refactoring the logging.

Test generation, logging, and `Arguments` are 3 distinct issues, and should not be combined like this.

clients/include/rocblas_arguments.hpp
leekillough: The new logging code will have everything listed once in one place -- there is no need for this solution.

This is attempting a C++ solution for something better handled in Python.

clients/include/rocblas_arguments.hpp
leekillough: There are already tools to dump YAML source which can be turned into tests.

A Python script can be written to take the YAML output and turn into a format compatible with  `*_gtest.yaml`.

This should be done by taking the raw YAML output from the logging, and transform it using Python code, rather than adding an `ArgumentsModel` to the C++ code.

clients/include/rocblas_arguments.hpp
leekillough: > 
> 
> The list of arguments is not known at compile time, only run time, thus the switch. The lambda is a simple functor much like the one pre-existing in the base class, to me a variadic template is more complicated.

`constexpr` can be used to make the constants known at compile-time.

leekillough:
We need a group-wide meeting to design a modern C++ or Python solution for what this is trying to solve. **So far I have not heard a statement of the problem.** 

A simpler implementation may be possible, once I understand the problem. `constexpr`, `initializer_list` `offsetof`, and other C++11 features can be used to avoid runtime-lookup of names. This implementation is too complex and unmaintainable.

The `Arguments` class and the `rocblas_arguments.hpp` is not the place where this should go -- it needs to go into a separate file, if it is implemented in C++, but it might be easier implemented in Python.

Although the changes are currently small in terms of lines of code, they significantly change how tests are written, how `Arguments` are handled, and their impacts on the orthogonality and consistency of the code is large, and they heavily conflict with already-planned changes for logging. They might also interact with long-term open JIRA tickets for changing how `rocblas-test` collects and reports timing information, but that might be moot now that we have stress tests (so that JIRA ticket should be closed).

Test generation is done in Python, with YAML specifications for tests. `rocblas_gentest.py` can be improved if necessary.

`Arguments`, logging, and test generation are distinct concepts, and should not be muddied.

Logging can already occur at multiple levels, including a YAML dump of the `Arguments` used. This YAML output can be input to `rocblas-test` or `rocblas-bench`, and a script to generate, extract, or modify the YAML is the preferred way to go for generating tests.

PR #921, opened 2020-01-10
add python3-distutils as dependency on Ubuntu

scchan:
The setup for virtualenv would fail without this package on Ubuntu


saadrahim:
The conflict should be resolved first.

install.sh
```diff
@@ -138,17 +138,17 @@ install_packages( )
...
-                                      "python34" "PyYAML" "python3*-PyYAML"
+                                      "python34" "PyYAML" "python3*-PyYAML" "python3*-distutils"
                                       "gcc-c++" "llvm7.0-devel" "llvm7.0-static"
                                       "zlib-devel" "wget" )
   local library_dependencies_fedora=( "make" "cmake" "rpm-build"
-                                      "python34" "PyYAML" "python3*-PyYAML"
+                                      "python34" "PyYAML" "python3*-PyYAML" "python3*-distutils"
                                       "gcc-c++" "libcxx-devel" "zlib-devel" "wget" )
-  local library_dependencies_sles=(   "make" "cmake" "python3-PyYAM"
+  local library_dependencies_sles=(   "make" "cmake" "python3-PyYAM"
```
leekillough: The trailing space should be automatically removed by `pre-commit` -- you must run `install.sh` at least once to install `pre-commit`.

zaliu:
Ubuntu 16.04.6, which is still a supported environment (cf. https://github.com/RadeonOpenCompute/ROCm#Supported-Operating-Systems), does *not* have the package python3-distutils.  As a result, this PR broke rocBLAS build on Ubuntu 16.04.6.  We need to either fix the breakage or back the PR out.

PR #922, opened 2020-01-12
Refactor logging, tuple handling, and argument handling

leekillough:
This makes several changes
- Make logging entries atomic in the presence of multiple threads (currently they are interleaved)
- Splits `rocblas_arguments.hpp` into `rocblas_arguments.hpp` and `rocblas_arguments.cpp`
- Puts `Arguments` fields in one single place
- Move `std::tuple` helper functions out of `logging.h` and into `tupler_helper.hpp`
- Rewrite argument validation, dumping of new Tensile host and printing arguments, in terms of tuples and functions to apply to their pairs
- Change tuple helper functions to operate on their elements in-order rather than backwards

This is a work-in-progress; it does not build 100% right now; it is **not ready for review**, and it will be modified some more, with a final refactoring change at the end to move stuff around files, remove unnecessary `#include`s, etc.

It is being submitted to show that there would be **_many_** conflicts with another recent PR.

[Now It's ready for review]


leekillough:
There will likely be changes to use C++ file facilities instead of low-level POSIX; this is just an initial implementation to get things rolling. It is not finished or ready for review.

leekillough:
I have refactored Torre's `ArgumentModel` to work with the new logging and `Arguments` handling system.

It is much better to express `ArgumentModel` as a variadic template class with zero or more enums expressing arguments, than a plain class containing a `std::vector` of `enum`s expressing arguments:
```
// ArgumentModel template has a variadic list of argument enums
template <rocblas_argument... Args>
class ArgumentModel
{
};
```
Then you can simply loop around the arguments at compile-time:
```
        // Apply the arguments, calling print on each one
        // TODO: Replace with C++17 fold comma expression
        (void)(int[]){(apply<Args>{}()(print, arg, T{}), 0)...};
```
(The `Args` template parameter represents a scalar of `rocblas_argument` type in the expression, but the entire expression is repeated across all Args with the `…` expansion. The `(void)(int[])` cast is to perform ordered evaluation with a brace-init, before the availability of C++17 fold expressions.)

Whether a particular argument is a member of the `ArgumentModel` is a compile-time decision (`constexpr`):
```
    // Whether model has a particular parameter (compile-time)
    static constexpr bool has(rocblas_argument param)
    {
        // TODO: Replace with C++17 fold expression
        for(auto x : {Args...})
            if(x == param)
                return true;
        return false;
    }

constexpr bool has_batch_count = has(e_batch_count);     // constexpr!!! Can be computed at compile-time
```
There is no need for `std::vector` or for `static` variables anymore -- you can simply express it as a trivial class instance (the class has no data members):
```
        ArgumentModel<e_N, e_incx, e_incy>{}.log_args<T>(rocblas_cout,
                                                         arg,
                                                         gpu_time_used,
                                                         dot_gflop_count<CONJ, T>(N),
                                                         (2.0 * N) * sizeof(T),
                                                         cpu_time_used,
                                                         rocblas_error_1,
                                                         rocblas_error_2);
```
You create a `ArgumentModel<e_N, e_incx, e_incy>{}` temporary object, which encodes the arguments in the model, and then you call a `log_args` method. There is no data payload to the `ArgumentModel` object, so it just represents typing.

There is little to no code -- no `static` variable, and no `std::vector<>` class, for example. It's just type mapping done at compile-time.

The code is much simpler now; I hope I wasn't too strong in my comments about Torre's original PR. I just have a strong desire for simplicity,  efficiency, and consistency.

In the new code, there is:
-	One declaration list of the `Arguments` class members, in the class itself
-	One macro which iterates across the `Arguments` class members, applying another arbitrary macro to the member name and value
-	Automatic definition of enumerations for `Arguments` members -- there's no need for a separate list
-	Variadic template argument lists which use enumerations to specify a subset of arguments, such as in `ArgumentModel`
-	Functions to apply enumeration lists to functions using the listed arguments, which are fully resolved at compile-time

There is no other duplication.

clients/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp -pthread")
 endif()
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftemplate-depth=2048")
```
pfultz2: Why is this necessary? In general, it isn't good to go beyond the default template depth.

clients/include/argument_model.hpp
```diff
@@ -0,0 +1,152 @@
...
+    {
+        // TODO: Replace with C++17 fold expression
+        for(auto x : {Args...})
+            if(x == param)
+                return true;
+        return false;
+    }
+
+    template <rocblas_argument>
+    struct apply
```
pfultz2: This class should not be nested in here as it can ruin any memorizing by the compiler because each `apply` is dependent on the outer `Args...` even though they are not used by the `apply` class.

clients/include/rocblas_arguments.hpp
```diff
@@ -84,91 +87,89 @@ struct Arguments
...
+
+#define COMMA_SEPARATOR ,
+
+    // clang-format on
+
+    /**************************************************************
+    * Tuple of argument name, value pairs, preserving references *
+    **************************************************************/
+#define NAME_VALUE_PAIR(NAME) #NAME, NAME
+    auto as_tuple() const
```
pfultz2: If the point of the tuple is to just to pack and then unpack, it could be simpler(and probably faster to compile) to use a lambda instead that visits each pair:

```cpp
#define NAME_VALUE_PAIR(NAME) f(#NAME, NAME);
auto visit_each() {
    return [=](auto f) {
        FOR_EACH_ARGUMENT(NAME_VALUE_PAIR,)
    };
}
```

And then printing could be implemented like this:

```cpp
template <typename Visitor>
static rocblas_ostream& print_pairs(rocblas_ostream& os, const Visitor& v)
{
    // Turn YAML formatting on
    os << rocblas_ostream::yaml_on;

    const char* delim = "{ ";
    v([&](const char* name, const auto& value) {
        os << delim << name << ": " << value;
        delim = ", ";
    });

    // Closing brace and turn YAML formatting off
    return os << " }\n" << rocblas_ostream::yaml_off;
}
```

And there is no need for dealing with indices or tuples.

clients/CMakeLists.txt
leekillough: > 
> 
> Why is this necessary? In general, it isn't good to go beyond the default template depth.

It was necessary because libstd++ uses an inefficient recursive method for computing `std::is_nothrow_move_constructible`. I don't have references handy, but I heavily researched it. I thought about adding a `noexcept(false)` clause to one level of our `std::tuple` usage, on the chance that it might short-circuit the inefficient `std::is_nothrow_move_constuctible` algorithm, but it seemed like increasing the limit was simpler. We only had to double it, approximately.

We use `std::tuple` with 46 elements. We don't use recursion ourselves, but rather use `std::index_sequence` and the like, i.e., we use iterative rather than recursive algorithms. It's libstdc++ which uses an inefficient recursive algorithm for `std::is_nothrow_move_constructible`.

leekillough: c.f. https://code-examples.net/en/q/164ac69

clients/include/argument_model.hpp
```diff
@@ -0,0 +1,152 @@
...
+        // Output (name, value) pairs to name_list and value_list
+        auto print = [&, delim = ""](const char* name, auto&& value) mutable {
+            name_list << delim << name;
+            value_list << delim << value;
+            delim = ",";
+        };
+
+        // Apply the arguments, calling print on each one
+        // TODO: Replace with C++17 fold comma expression
+        (void)(int[]){(apply<Args>{}()(print, arg, T{}), 0)...};
```
pfultz2: This should be put into a separate function to apply a function to each args as there is a lot of sublte C++ quirks happening in a small amount of code. Ideally, something like this:

```cpp
struct swallow
{
    template <class... Ts>
    constexpr swallow(Ts&&...)
    {
    }
};
template <class F, class... Ts>
constexpr void each_args(F f, Ts&&... xs)
{
    swallow{(f(std::forward<Ts>(xs)), 0)...};
}
```

And then this can be written as:

```cpp
each_args([&](auto&& x) {
    apply<Args>{}()(print, x, T{});
}, xs...);
```

clients/CMakeLists.txt
pfultz2: Then maybe you should consider using a lambda instead of tuple. I present one possible solutions.

pfultz2: `std::tuple` is rather slow for compile-time and inefficient implementation(like in libstdc++) makes it worse.

leekillough: > 
> 
> `std::tuple` is rather slow for compile-time and inefficient implementation(like in libstdc++) makes it worse.

This is only for our tests, not for the main library, so performance and compile-time efficiency are not the most important concerns (code readability and reusability are more important).

leekillough: You'll notice that the increased limit is only in `clients/CMakeLists.txt`. It is not necessary that it go into `library/CMakeLists.txt`. That's what I mean by saying it's only in the tests, not the library.

clients/include/argument_model.hpp
leekillough: Some functions rely on the template arguments, which is why it is a nested class.

leekillough: Originally I tried separating it into `.hpp/.cpp` files, but then I found that the instantiations required would be too long, and would be complicated by the fact that lambdas (captures) have unique types which can't be expressed in another file, only in the same file with `decltype()`. So I had to put them in the header, and since they sometimes depend on template parameters, I had to make it a nested class.

clients/include/argument_model.hpp
leekillough: I think it would make the code harder to understand if we added additional functions/classes/lambdas. The `apply` function, as it stands now, does what we need it to for a specific purpose.

I'd prefer to use `std::apply`, but we're using `-std=c++14` now. I wonder if we should make the leap to `-std=c++17` yet. With C++17 we'd gain:
- `if constexpr`
- fold expressions
- `std::apply`
- `std::variant`
- automatic deduction of class template arguments from constructor arguments
- ...

clients/include/rocblas_arguments.hpp
leekillough: No, besides packing/unpacking, the point of using `std::tuple` is also to perform associative lookup (e.g., `std::unordered_map`). We need to track every argument regardless of type, i.e., to perform associative container operations in a type-generic way.

leekillough: `std::tuple` is a heterogeneous literal type which can be `hash`ed (although we write our own `hash` and `less` functions).

Almost all of the code you're talking about is in our testing code, not in our main library code. The `Arguments` class is specifically for our testing code; it's a universal way of expressing all arguments and external parameters of a particular problem to be tested. It is supposed to be cross-language (C/C++/Python/YAML), which is why it's so POD-ish.

We want the methods applied to `Arguments` to be simple and universal, to make it easier to write tests for dozens of different routines, rather than having the most efficient compile-time or run-time.

clients/include/argument_model.hpp
leekillough: It has been moved into an `ArgumentsHelper` class. I don't know why I thought it depended on the template parameters. It's much simpler now, since it uses `constexpr` lambda variable templates.

clients/include/rocblas_arguments.hpp
leekillough: For the printing of `(name, value)` pairs, a tuples implementation has been replaced with C macro iteration (`FOR_EACH_ARGUMENT`). But for the tracking and recording of `Arguments`, `std::tuple` is still necessary because it's stored in a `std::unordered_map` as a key.

The limit is not reached unless a large percentage of  `Arguments` are used to construct a `std::tuple`. With a subset of `Arguments` for a given problem, the limit is not reached. As stated earlier, this is due to a bad `libstdc++` implementation of `std::is_nothrow_move_constructible`.

clients/CMakeLists.txt
leekillough: We still need to use `std::tuple` in some places, but rarely with the full `Arguments` list. With the subsets we use, we no longer need this limit increase.

As I said before, we don't use recursion (we use `std::index_sequence` and the like). The overflow was caused by the inefficient recursive implementation of libstdc++'s `std::is_nothrow_move_constructible`.

clients/include/argument_model.hpp
leekillough: For some reason I thought it depended on template parameters, but I guess not. I moved it to a totally separate non-templated class.

TorreZuk:
Was there a multi-thread unit test in there I missed as there was a lot of files?   You could have only submitted a multi-thread unit-test plus the new thread safe streams before using it in all the functions.

leekillough:
> Was there a multi-thread unit test in there I missed as there was a lot of files? You could have only submitted a multi-thread unit-test plus the new thread safe streams before using it in all the functions.

I still need to write a unit test.

The old code was "thread-safe" in the sense of outputting the correct output and not crashing; it just didn't guarantee that two threads wouldn't interleave their output.

With `rocblas_ostream`, the output is guaranteed not to interleave. That's the main test I need to write.

The other benefit is that `rocblas_ostream` can be used the same way for string streams and files, making the code look simpler (it's called `rocblas_ostream` because it's similar to `std::ostream`, but it also has a default constructor for a string stream, and explicit constructors for filenames or filehandles).

Besides avoiding interleaving during multithreading, `rocblas_ostream` also has formatted output which is consistent across the library, and `rocblas_ostream` and makes sure to flush output at certain points (such as `rocblas_abort`).

99% of the changes are simply `s/std::cout/rocblas_cout/` and `s/std::cerr/rocblas_cerr`.

Because of the nature of the changes, it's hard to split it up into two parts now, but if you want, I could create a (non-buildable) branch with the mechanical changes above, and their `clang-format` side-effects, taken out.

TorreZuk:
Yes it would be good to have a multi-thread test as that was the main reason for your work to ensure it handles multiple threads nicely.   Ensuring deadlock doesn't occur and your queues are robust is difficult to judge by reading.

As for rocblas_arguments you haven't convinced me that the variadic template class is better than my original non-templated class.   Compiling the template for every parameter combination seems like added overhead where we don't need it for performance reasons, instancing my class would be a simpler design.   Your pattern I see potential for in greatly simplifying the runtime logging system where the parameters are templated so we are generating type specific code already.

leekillough:
> Yes it would be good to have a multi-thread test as that was the main reason for your work to ensure it handles multiple threads nicely. Ensuring deadlock doesn't occur and your queues are robust is difficult to judge by reading.

This has always been planned; there have just been a lot of roadblocks which have slowed things down. Also, this does not fit the pattern that most of the rest of rocBLAS uses with Google Tests, so we'll have to create a separate non-parameterized test case for it. It may also require using scratch space for files.

> As for rocblas_arguments you haven't convinced me that the variadic template class is better than my original non-templated class. Compiling the template for every parameter combination seems like added overhead where we don't need it for performance reasons, instancing my class would be a simpler design. Your pattern I see potential for in greatly simplifying the runtime logging system where the parameters are templated so we are generating type specific code already.

There may be more instantiations than in a single implementation which interprets at runtime, but ArgumentModel is in a header file for a reason, and we are not concerned with code size. I am fond of "header-only-libraries". I don't think that this implementation increases the compile time substantially over a runtime interpreted method; it certainly does not increase the runtime overhead over your implementation; it just increases the code size over a runtime interpreted single method, since it generates one template instantiation per model.

I thought the original motivation was to decrease code duplication and lines of code, and make adding tests easier, by having a simple way to express a model representing a list of arguments,  and providing methods which use the model. That is what this code does, at almost no additional runtime memory or time overhead. Even if it instantiates a template for each model separately, it's no worse than the original code which wrote everything out explicitly for each separate problem. It is the whole reason templates are used -- to provide a single class, function or variable which can be instantiated in a number of ways, without increasing source code size (only maybe generated code size).

I think the code is much easier to read than the your code with the switch statements, static variables, std::vector runtime initialization, and large number of macros. Now there's just a simple ArgumentModel<...>{} instantiation, and methods to apply to it, with no initialization or data payload such as std::vector. You just express a model with template arguments, and then you invoke methods on the model, which are able to iterate through the model's arguments with a simple parameter-pack expansion. Lambdas are able to extract Arguments from an instance,  given an enum specifying the argument to be extracted. So you can extract and operate on arguments according to the model's parameter list. Much, much simpler than static initialization of a model which requires std::vector initialization, and then calling methods which iterate through the vector and use complex switch statements.

The new FOR_EACH_ARGUMENT macro is to universalize iterating over the arguments with any other macro, so that the arguments are listed in only two places in the same file: the Arguments member declarations, and the FOR_EACH_ARGUMENT macro definition. Although the ArgumentModel doesn't need to be faster by knowing the arguments at compile-time, this is a desirable side-benefit of variadic templates.

The logging system is very templated, and I would not want to change that to a runtime model, which unnecessarily complicates things, including the readability of the code. Right now it is very free-form with variadic functions taking arbitrary names and arguments to log, and if a kernel needs to log a little differently, it can convert certain arguments specially, like alpha and beta, and it still works under the generic template logging system, without having to change the underlying implementation. That is why I said your original code deviated from the design of rocBLAS. Within a single library, source base, module, repository, etc., it is always preferred that generic templates be used for reuse, instead of macros or a runtime interpreted model when a compile-time model is just as simple, because the template definition and usage are within the same module, and they do not impact outside projects.

(Ironically, I have the exact opposite opinion on rocSOLVER<->rocBLAS interfacing; I think it should be a runtime SolverProblem/RocblasProblem descriptor model, not templates called across shared libraries, because separating rocBLAS and rocSOLVER from dependency on each other is the single most important goal, something that requiring compile-time knowledge of each others' APIs would not allow.)

leekillough:
(To be added to the [guidelines](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.github/CONTRIBUTING.md):)

21. `rocblas_cout` and `rocblas_cerr` should be used instead of `std::cout`, `std::cerr`, `stdout` or `stderr`, and `rocblas_ostream` should be used instead of `std::ostream`, `std::ofstream` or `std::ostringstream`.

    In `rocblas-bench` and `rocblas-test`, `std::cout`, `std::cerr`, `printf`, `fprintf`, `stdout`, `stderr`, and other symbols are "poisoned", to remind you to use `rocblas_cout`, `rocblas_cerr`, and `rocblas_ostream` instead.

    `rocblas_cout` and `rocblas_cerr` are instances of `rocblas_ostream` which output to standard output and standard error, but in a way that prevents interlacing of different threads' output.

    `rocblas_ostream` provides standardized thread-safe formatted output for rocBLAS datatypes. It can be constructed in 3 ways:
    - By default, in which case it behaves like a `std::ostringstream`
    - With a file descriptor number, in which case the file descriptor is `dup()`ed and the same file it points to is outputted to
    - With a string, in which case a new file is opened for writing, with file creation, truncation and appending enabled (`O_WRONLY | O_CREAT | O_TRUNC | O_APPEND`)

    `std::endl` or `std::flush` should be used at the end of an output sequence when an atomic flush of the output is needed (atomic meaning that multiple threads can be writing to the same file, but that their flushes will be atomic). Until then, the output will accumulate in the `rocblas_ostream` and will not be flushed until either `rocblas_ostream::flush()` is called, `std::endl` or `std::flush` is outputted, or the `rocblas_ostream` is destroyed.

    The `rocblas_ostream::yaml_on` and `rocblas_ostream::yaml_off` IO modifiers enable or disable YAML formatting, for when outputting abitrary types as YAML source code. For example, to output a `key: value` pair as YAML source code, you would use:

```c++
    os << key << ": " << rocblas_ostream::yaml_on << value << rocblas_ostream::yaml_off;
```
<p>The `key` is outputted normally as a bare string, but the `value` uses YAML metacharacters and lexical syntax to output the value, so that when it's read in as YAML, it has the type and value of `value`.


22. C++ templates, including variadic templates, are preferred to macros or runtime interpreting of values, although it is understood that sometimes macros are necessary.

    For example, when creating a class which models zero or more rocBLAS kernel arguments, it is preferable to use:
```c++
    template<rocblas_argument... Args>
    class ArgumentModel
    {
public:
        void func()
        {
            for (auto arg: { Args... })
            {
                // do something with argument arg
            }
        }
    };

    ArgumentModel<e_A, e_B>{}.func();

```
   instead of:
```c++
    class ArgumentModel
    {
         std::vector<rocblas_argument> args;
public:
         ArgumentModel(const std::vector<rocblas_argument>& args)
         : args(args)
         {
         }

        void func()
        {
            for (auto arg: args)
            {
                // do something with argument arg
            }
        }

    };

    static ArgumentModel model({e_A, e_B});
    model.func();
```
The former denotes the rocBLAS arguments as a list which is passed as a variadic template argument, and whose properties are known and can be optimized at compile-time, and which can be passed on as arguments to other templates, while the latter requires creating a dynamically-allocated runtime object which must be interpreted at runtime, such as by using `switch` statements on the arguments. The `switch` statement will need to list out and handle every possible argument, while the template solution simply passes the argument as another template argument, and hence can be resolved at compile-time.

leekillough:
Closing before release. Keeping draft PR open for review.

leekillough:
This is ready to be merged. It needs to be merged sooner rather than later.

PR #924, opened 2020-01-12
Remove hipMalloc and hipFree warnings from samples

leekillough:
This removes compilation warnings in the samples about `hipMalloc` and `hipFree` being deprecated.

Alternatively, we can `#undef` `hipMalloc` and `hipFree`, but we need to make sure it only applies to the samples, not to the main library or tests.

### DIFF
--- a/clients/samples/example_openmp.cpp
+++ b/clients/samples/example_openmp.cpp
@@ -59,8 +59,8 @@ int main()
     }
 
     // allocate memory on device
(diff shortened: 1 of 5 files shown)


YvanMokwinski:
> If one isn't supposed to use hipMalloc then the alternative should be shown to the reader. Using non-standard brackets to escape a define without comment doesn't seem proper.

I agree.

leekillough:
This is being folded into another change

PR #930, opened 2020-01-14
Blas2 tpmv functional implementation.

YvanMokwinski:
resolves JIRA 219155

Basic implementation of a functional tpmv (non-batched / batched / strided_batched)


library/src/blas2/rocblas_tpmv.cpp
```diff
@@ -0,0 +1,170 @@
...
+                            m,
+                            "incx",
+                            incx);
+            }
+        }
+
+        if(uplo != rocblas_fill_lower && uplo != rocblas_fill_upper)
+        {
+            return rocblas_status_not_implemented;
+        }
```
TorreZuk: There is now rocblas_status_invalid_value which I think is better as we don't expect to ever implement support other cases.

clients/gtest/tpmv_gtest.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+
+Tests:
+- name: tpmv_bad_arg
+  category: pre_checkin
+  function: tpmv_bad_arg
+  precision: *single_double_precisions
+  uplo: [L, U]
+  transA: [N, T, C]
+  diag: [N, U]
```
TorreZuk: Seems like these aren't used so could be deleted until we determine if we will improve the invalid value type checking pattern.

clients/include/testing_tpmv_strided_batched.hpp
```diff
@@ -0,0 +1,258 @@
...
+    // Initialize.
+    //
+    rocblas_init(hA, true);
+    rocblas_init(hx);
+
+    //
+    // Transfer.
+    //
+    CHECK_HIP_ERROR(dA.transfer_from(hA));
+    CHECK_HIP_ERROR(dx.transfer_from(hx));
```
TorreZuk: Leaving the memory uninitialized I think is fine.

library/include/rocblas-functions.h
```diff
@@ -3986,6 +3986,281 @@ ROCBLAS_EXPORT rocblas_status rocblas_ztrmv_strided_batched(rocblas_handle
                                                             rocblas_stride                stridex,
                                                             rocblas_int batch_count);
 
+/*! \brief BLAS Level 2 API
+
+    \details
+    tpmv performs one of the matrix-vector operations
+
+         x = A*x or x = A**T*x,
```
daineAMD: Should probably mention that A is a packed matrix here.

clients/include/testing_tpmv_strided_batched.hpp
YvanMokwinski: Which uninitialized memory?

library/src/blas2/rocblas_tpmv.cpp
YvanMokwinski: Good point

clients/include/testing_tpmv.hpp
```diff
@@ -0,0 +1,247 @@
...
+    auto inc = std::abs(that.inc());
+    os << " " << n << " ";
+    os << "{ " << that[0];
+    for(rocblas_int i = 1; i < n; ++i)
+    {
+        os << ", " << that[i * inc];
+    }
+    os << " }" << std::endl;
+    return os;
+}
```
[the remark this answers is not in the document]
YvanMokwinski: Cannot be in the host vector file ... because of half precision.

clients/include/testing_tpmv_strided_batched.hpp
TorreZuk: I'm saying I wouldn't bother with rocblas_init or upload.  This is inside bad_arg (look at larger code context), based on the parameters you pass the tests never use hA, hx, dA, dx values but if they do uninitialized memory is fine.

clients/common/rocblas_gentest.py
```diff
@@ -221,6 +221,14 @@ def setdefaults(test):
...
             test.setdefault('stride_c', int(test['stride_scale']) * 5)
 
+    elif test['function'] in ('tpmv_strided_batched'):
+        if all([x in test for x in ('M', 'incx', 'stride_scale')]):
+            ldx = int(test['M'] * abs(test['incx']) * test['stride_scale'])
+            test.setdefault('stride_x', ldx)
+        if all([x in test for x in ('M', 'stride_scale')]):
+            ldM = int( ( (test['M'] + 1) * test['M'] ) / 2 * test['stride_scale'])
+            test.setdefault('stride_a', ldM)
+
```
mahmoodw: I think Daine has added the setkey_product helper that you may find useful to clean up the code

YvanMokwinski: Thanks, will take a look and will adjust.

clients/gtest/tpmv_gtest.yaml
YvanMokwinski: Can be kept too. I'm keeping it.

library/include/rocblas-functions.h
YvanMokwinski: yes

clients/include/testing_tpmv.hpp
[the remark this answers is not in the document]
YvanMokwinski: removed.

library/src/blas2/rocblas_tpmv.cpp
YvanMokwinski: done.

clients/include/testing_tpmv_strided_batched.hpp
YvanMokwinski: done.

library/include/rocblas-functions.h
YvanMokwinski: done.

clients/common/rocblas_gentest.py
YvanMokwinski: done

PR #931, opened 2020-01-15
Add flag to ignore Tensile version

amdkila:
This PR adds a flag that allows the user to install an older version of Tensile without the install script returning an error code. Mostly, this is for CI purposes.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,11 +20,11 @@ rocBLAS build & installation helper script
       -l | --logic             Set Tensile logic target, e.g., asm_full, asm_lite, etc.
       -a | --architecture      Set Tensile GPU architecture target, e.g. all, gfx000, gfx803, gfx900, gfx906, gfx908
       -o | --cov               Set Tensile code_object_version (V2 or V3)
-      -t | --test-local-path   Use a local path for Tensile instead of remote GIT repo
-           --cpu-ref-lib       Specify library to use for CPU reference code in testing (blis or lapack)
+      -t | --test_local_path   Use a local path for Tensile instead of remote GIT repo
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -322,22 +324,22 @@ while true; do
...
         shift 2 ;;
-    -n|--no_tensile)
+    -n|--no-tensile)
         build_tensile=false
         shift ;;
     -s|--tensile-host)
         build_tensile_host=true
         shift ;;
-    --build_dir)
+    --build-dir)
```
saadrahim: why are you changing the spelling? If there was no request, we risk breaking scripts.

install.sh
```diff
@@ -20,12 +20,13 @@ rocBLAS build & installation helper script
...
-           --cpu_ref_lib       Specify library to use for CPU reference code in testing (blis or lapack)
+      -t | --test-local-path   Use a local path for Tensile instead of remote GIT repo
+           --cpu-ref-lib       Specify library to use for CPU reference code in testing (blis or lapack)
            --hip-clang         Build library for amdgpu backend using hip-clang
-           --build_dir         Specify name of output directory (default is ./build)
-      -n | --no_tensile        Build subset of library that does not require Tensile
+           --build-dir         Specify name of output directory (default is ./build)
+      -n | --no-tensile        Build subset of library that does not require Tensile
       -s | --tensile-host      Build with Tensile host
+      -u | --use-tag-only      Ignore Tensile version and just use the Tensile tag
```
saadrahim: This is such an uncommon flag that maybe it should be spelled out in full when using. It will make for more readable scripts rather than assign -u

PR #932, opened 2020-01-15
use wildcard to workaround installation issue for python3-distutils

scchan:


amdkila:
I think it's best we close this one and merge my PR #931 as it downloads less unrelated packages than this change. I specified python3*-distutils, which installs about 10 packages, while python*-distutils* installs 39 packages, including python2-distutils packages.

PR #934, opened 2020-01-15
Functional HPMV

daineAMD:
Resolves SWDEV-219156.

Functional implementation for chpmv and zhpmv (regular/batched/strided_batched versions).

[Hermitian packed matrix-vector multiply.](http://www.netlib.org/lapack/explore-html/d6/dee/group__complex__blas__level2_gad1dc14308bf0c1c8d9ba6ee068ac4d60.html)


library/src/blas2/rocblas_hpmv.hpp
```diff
@@ -0,0 +1,225 @@
...
+
+    hpmv_kernel_calc<DIM_X, DIM_Y>(upper, n, alpha, AP, x, incx, beta, y, incy);
+}
+
+/**
+  *  U is always: const T* (either host or device)
+  *  V is either: const T* OR const T* const*
+  *  W is either:       T* OR       T* const*
+  */
+template <typename U, typename V, typename W>
```
YvanMokwinski: template<typename S, typename A, typename X, typename Y>
S as scalar type, 
A as a-type
X as x-type
Y as y-type

library/src/blas2/rocblas_hpmv_batched.cpp
```diff
@@ -0,0 +1,179 @@
...
+            return rocblas_status_invalid_size;
+
+        if(!n || !batch_count)
+            return rocblas_status_success;
+
+        if(!AP || !x || !y || !alpha || !beta)
+            return rocblas_status_invalid_pointer;
+
+        return rocblas_hpmv_template(
+            handle, uplo, n, alpha, AP, 0, 0, x, 0, incx, 0, beta, y, 0, incy, 0, batch_count);
```
YvanMokwinski: naming (here zero) constants increases readability

library/src/blas2/rocblas_hpmv.hpp
daineAMD: I agree that U, V, W probably aren't the best descriptors...
I implemented a counter-proposal:
`TScal`, `TConstPtr`, `TPtr`
I think these are more descriptive that Scalar-type, a-type, etc. They also allow multiple variables to use the same type without adding confusion. Let me know what you think.

TorreZuk: I like that but are the Ts always the same type for this function?  Mixed precision function?

daineAMD: AFAIK the Ts should always be the same, I don't see otherwise. T should always either be rocblas_float_complex or rocblas_double_complex.

YvanMokwinski: I wonder what would we use after U, V, W :)
I wrote a document on some useful definition of these types, that would allow us many traits mechanisms.
So far, with some of our abstractions of the types, 

<U,V,W> + { auto + load_ptr_batch }

some routines could even perform operations involving a batched vector and a strided batched vector (benefit of skipping T and use the auto keyword when loading the batch ptr, 99% to 100% of the use of load_ptr_batch is let to the compiler (a grep would confirm)). If underlying T types are different I guess the compiler would scream ... 

There is a lot of U,V,W in rocBLAS. 
Fundamentally, we are doing nothing explicitly with those types, maybe a few std::are_same ..? We let the compiler do. So, it doesn't not really matter from a 'functional' point of view, neither readability except that I think U,V,W is weird.
 
Otherwise, the question is how many template parameters are used.
S,A,B,C (here 4) or TScal,TConstPtr,TConstPtr,TPtr (here 3, and matrix a and b are of the same pointer types).

- naming is a nice to have
- number of template parameters decides whether or not we allow a mix between batched and strided batched vector / matrix.

YvanMokwinski: Your routine works with V = const T* (strided_batched)  and W = T * const * (batched). (?)

daineAMD: That situation should never happen, right? I think it would work in that situation, though, as everything is handled in load_ptr_batch.

YvanMokwinski: Why not?... (only raising the question)

YvanMokwinski: For the mixed precision, I think the kernels form a barrier since (most of) they require everything being of the same type T.

daineAMD: Well at least from our current API we don't support mixing batched and strided_batched routines together (having some variables batched and other variables strided_batched). As far as I know we aren't planning on supporting this in the future.
But if this did ever come up, I don't think many code changes would have to happen here as `load_ptr_batch` should automatically do the necessary conversions from a batched array to a single pointer either way.

Mixed precision is it's own issue, it only seems to come up with certain functions (gemm, trsm, etc.) and it doesn't seem worthwhile to have _every_ function future-proof for mixed precision, unless this is something that might have to be implemented across the project, but I haven't heard anything about this.

YvanMokwinski: Agree, only discussing the constraints as a consequence of the current type abstraction, which does not translate completely the C-API constraint.

PR #936, opened 2020-01-15
Complex trtri trsm

mahmoodw:
resolves SWDEV-218025

Added complex support to trtri and trsm


clients/gtest/trsm_gtest.yaml
```diff
@@ -20,6 +20,14 @@ Definitions:
 
   - &alpha_range [ 1.0, -5.0 ]
 
+  - &complex_alpha_range
```
daineAMD: I don't think you need separate test cases for real/complex precision. You can just use `*single_double_precisions_complex_real` as in `trtri_gtest.yaml` and have a single alpha range (like this one), `alphai` should just be ignored in the real cases.

library/src/blas3/rocblas_trsm.hpp
```diff
@@ -19,11 +19,11 @@ namespace
     using std::min;
 
     template <typename T>
-    constexpr T negative_one = -1;
+    const T negative_one = -1;
```
daineAMD: `const T negative_one = T(-1);`, perhaps.

clients/include/testing_trsm.hpp
```diff
@@ -254,11 +259,12 @@ void testing_trsm(const Arguments& arg)
                     res_2 += std::abs(hXorB_2[j + i * ldb]);
                 }
             }
```
TorreZuk: If this doesn't exist as a helper I think it should be put there, can be reused in the batch cases.

clients/include/testing_trsm_strided_batched.hpp
```diff
@@ -153,7 +153,7 @@ void testing_trsm_strided_batched(const Arguments& arg)
             {
                 int idx = i + j * lda + b * stride_a;
                 hA[idx] = AAT[idx];
-                t += AAT[idx] > 0 ? AAT[idx] : -AAT[idx];
+                t += std::abs(AAT[idx]);
             }
             hA[i + i * lda + b * stride_a] = t;
         }
```
TorreZuk: Could be a helper function ?

clients/include/unit.hpp
```diff
@@ -398,4 +398,31 @@ inline void trsm_err_res_check(T max_error, rocblas_int M, T forward_tolerance,
...
 
+template <>
+inline void trsm_err_res_check(rocblas_float_complex max_error,
+                               rocblas_int           M,
+                               rocblas_float_complex forward_tolerance,
+                               rocblas_float_complex eps)
+{
+    trsm_err_res_check<float>(std::abs(max_error), M, std::abs(forward_tolerance), std::abs(eps));
+}
+
```
TorreZuk: Is this reusable beyond trsm, could rename and say what it does?

clients/gtest/trsm_gtest.yaml
mahmoodw: Thanks for the suggestion. I was following gemm style but this is cleaner.

clients/include/testing_trsm.hpp
leekillough: `rocblas_abs` should be used instead of `std::abs` in rocBLAS library code.

clients/include/testing_trsm_strided_batched.hpp
leekillough: `rocblas_abs` is the helper function.

clients/include/unit.hpp
leekillough: `rocblas_abs` should be used instead of `std::abs`.

clients/gtest/trsm_gtest.yaml
```diff
@@ -21,12 +21,12 @@ Definitions:
...
-    - { alpha:  2, alphai:  0 }
-    - { alpha:  5, alphai:  0 }
+    # - { alpha_beta:  2, alphai:  0 }
     - { alpha: -5, alphai: -5 }
     - { alpha:  1, alphai:  3 }
-    - { alpha: -1, alphai:  3 }
-    - { alpha:  0, alphai:  2 }
+    - { alpha:  5, alphai:  3 }
+    # - { alpha_beta:  0.0, alphai:  0.0 }
+    # - { alpha_beta:  0, alphai:  2 } #Will not work for non complex (need to fix zero alpha behaviour)
```
mahmoodw: Will remove

library/src/blas3/rocblas_trsm.hpp
```diff
@@ -1435,6 +1486,22 @@ rocblas_status rocblas_trsm_template(rocblas_handle    handle,
...
+    if(saved_pointer_mode == rocblas_pointer_mode_host)
+        alpha_h = *alpha;
+    else
+        RETURN_IF_HIP_ERROR(hipMemcpy(&alpha_h, alpha, sizeof(T), hipMemcpyDeviceToHost));
+
+    if(alpha_h == T(0.0))
+    {
+        set_block_unit<T>(handle, m, n, B, ldb, stride_B, batch_count, offset_B);
+        return rocblas_status_success;
+    }
```
mahmoodw: Added quick return to match Fortran behaviour  - not currently being tests but locally tested

PR #937, opened 2020-01-16
Enable hip-clang functionality with the rocBLAS master branch

amdkila:
All fixes required to make the rocBLAS master branch to build with hip-clang, while retaining the ability to build with hcc.


leekillough:
> 
> 
> -Also refactors CI, and adds hip-clang CI
> 
>     * Still need to fix linker issue- undefined reference to trunc

`-lm`

amdkila:
> > -Also refactors CI, and adds hip-clang CI
> > ```
> > * Still need to fix linker issue- undefined reference to trunc
> > ```
> 
> `-lm`

Tried that. Torre ended up figuring out the fix in his commit. Looks like needed to link a specific static header file.

PR #939, opened 2020-01-16
Syr x complex

TorreZuk:
* adds complex types
* refactors batched non-batched to common template
* CPU implementation for complex syr


clients/include/cblas_interface.hpp
```diff
@@ -1119,9 +1119,40 @@ inline void cblas_ger(rocblas_int m,
...
+    T* x = (incx < 0) ? xa - ptrdiff_t(incx) * (n - 1) : xa;
+
+    if(uplo == rocblas_fill_upper)
+    {
+        for(int j = 0; j < n; ++j)
+        {
+            T tmp = alpha * x[j * incx];
+            for(int i = 0; i <= j; ++i)
+            {
+                A[i + j * lda] = A[i + j * lda] + x[i * incx] * tmp;
```
daineAMD: The docs [here](http://www.netlib.org/lapack/explore-html/d2/d15/group__complex_s_yauxiliary_gab7fbaf7c4477110907f682b250b24abe.html#gab7fbaf7c4477110907f682b250b24abe) have the operation as `A := alpha*x*x**H + A`. Is `x**H` not the conjugate transpose of `x`? In their code there it doesn't seem as they're conjugating it either which makes me wonder if I'm interpreting the docs incorrectly.

library/src/blas2/rocblas_syr.hpp
```diff
@@ -5,35 +5,79 @@
...
+                                   U              alpha_device_host,
+                                   rocblas_stride stride_alpha,
+                                   V              xa,
+                                   ptrdiff_t      shiftx,
+                                   rocblas_int    incx,
+                                   rocblas_stride stridex,
+                                   W              Aa,
+                                   ptrdiff_t      shiftA,
+                                   rocblas_int    lda,
+                                   rocblas_stride strideA)
```
YvanMokwinski: You don't need the template parameter T. You are using auto for x but T for A, mind as well use auto for A ?

library/src/blas2/rocblas_syr.hpp
```diff
@@ -43,38 +87,51 @@ rocblas_status rocblas_syr_template(rocblas_handle handle,
...
 
-    if(incx < 0)
-        x -= ptrdiff_t(incx) * (n - 1);
+    // in case of negative inc shift pointer to end of data for negative indexing tid*inc
+    auto shiftx = incx < 0 ? offsetx - ptrdiff_t(incx) * (n - 1) : offsetx;
 
     if(rocblas_pointer_mode_device == handle->pointer_mode)
-        hipLaunchKernelGGL(rocblas_syr_kernel,
+    {
+        hipLaunchKernelGGL((rocblas_syr_kernel<T>),
```
YvanMokwinski: <T> not needed if T removed from the template parameters

library/src/blas2/rocblas_syr.hpp
```diff
@@ -5,35 +5,79 @@
...
+__global__ void rocblas_syr_kernel(rocblas_fill   uplo,
+                                   rocblas_int    n,
+                                   U              alpha_device_host,
+                                   rocblas_stride stride_alpha,
+                                   V              xa,
+                                   ptrdiff_t      shiftx,
+                                   rocblas_int    incx,
+                                   rocblas_stride stridex,
+                                   W              Aa,
+                                   ptrdiff_t      shiftA,
```
YvanMokwinski: T is not needed here, it can be removed with the use of the 'auto' keyword, partially used in this implementation, see x and A.

clients/include/cblas_interface.hpp
TorreZuk: I think you are interpreting the documents correctly but I hope just like ours they have bugs.   I am following the code.   Both cuda and magma docs say a transpose not **H ,  I better dig into the magma code to ensure this.   As I am writing the reference and function I better be sure.
The complex conjugates are in her, but of course that matrix is Hermitian too, this is a complex non-hermitian symmetric matrix.

library/src/blas2/rocblas_syr.hpp
TorreZuk: Yes I thought about that but I prefer to see the base type as a template.  This helps tracking the base type templates throughout the code starting from the API down to the kernel.   This is of course a personal preference, I also avoiding auto is good.

clients/include/cblas_interface.hpp
TorreZuk: Also matches blis/flame as reference but this looks like a straight copy of lapack on which my cpu code was based.   See newer commit.  I think we can call this resolved agreed?

daineAMD: Yup, that's fine with me. I'd maybe suggest stressing in the rocblas-functions.h documentation that it is still a transpose and **not** a conjugate-transpose in the complex case, but other than that this is good with me.

TorreZuk: Well it says A[i] := A[i] + alpha*x[i]*x[i]**T   ,  usually we say if an alternative **H is done for complex.   Should I say " for all data types including complex" or does that confuse?

daineAMD: I guess it's probably ok as is if you think. If anyone wants to be sure they can always look at the implementation.

TorreZuk: I'll leave it as it correctly says what we do.

library/src/blas2/rocblas_syr.hpp
YvanMokwinski: How does it help tracking the base type? 
This abstraction has been made most of the time, auto is used especially on this purpose when loading batch ptr.

TorreZuk: I think it is a better design pattern to keep the core template type explicit, you see the template functions stand out to the reader as you read through the function.   Using auto here to save refactoring cost I don't buy that argument in this case.     ( code << refactor << read ) by orders of magnitude so this personally I think removing a typename by using auto is less readable.   Of course this is up for debate, but like your proposal to make the typename indicate something (e.g. S for scalar, or what I did in one function XPtr ) it might make the code more readable.    An open source library should be very readable if we want people to extend it, so I am suggesting we lean toward readability over minor code size reduction.   Sure the compiler can figure things out but often a reader has to figure something out too.   We can continue this discussion on Monday over coffee.

library/src/blas2/rocblas_syr.cpp
```diff
@@ -94,38 +102,33 @@ namespace
...
-                            rocblas_int    lda)
-try
-{
-    return rocblas_syr_impl(handle, uplo, n, alpha, x, incx, A, lda);
-}
-catch(...)
-{
-    return exception_to_rocblas_status();
-}
+#define IMPL(routine_name_, T_)                                           \
```
leekillough: I think templates would be better than macros.

library/src/blas2/rocblas_syr.hpp
leekillough: I agree that `T` is not necessary. But rather than using `auto`, you can use `decltype` or traits to determine the type based on other, automatically deduced, template parameters. Such as:
```
using T = decltype(...); // Put an expression in (...) which evaluates to the type of T
T* A;
```
Or short of going complete `auto`, you can use `auto *` to indicate that it's a pointer.

library/src/blas2/rocblas_syr.cpp
TorreZuk: I will try to redo with a template.

library/src/blas2/rocblas_syr.hpp
leekillough: I'm late to this conversation, but I think that not requiring explicit template type arguments (having them deduced instead) is preferred. `decltype()` can be used to "compute" the type based on another expression, without requiring the user to explicitly specify the type.

I don't understand: "Yes I thought about that but I prefer to see the base type as a template". If you're talking about `rocblas_float_complex` -> `float` and `rocblas_double_complex` -> `double`, then I think traits are a better way to do this -- in one of the header files, define a `real_type` template:

```
template<typename>
struct rocblas_real_type
{
};

template<>
struct rocblas_real_type<rocblas_float_complex>
{
    using type = float;
};

template<>
struct rocblas_real_type<rocblas_double_complex>
{
    using type = double;
};

using T = typename rocblas_real_type<U>::type;
```

leekillough: Actually, it would be simpler to just say:
```
using T = decltype(std::real(X));
```

i.e., use `decltype()` with `std::real` or a similar expression.

library/src/blas2/rocblas_syr.cpp
YvanMokwinski: ?

library/src/blas2/rocblas_syr.hpp
TorreZuk: Will resolve with a later PR as preference based.  Resolving this conversation.

library/src/blas2/rocblas_syr.cpp
TorreZuk: Can't see how to easily replace the extern C symbols with an instatiated template.  This pattern exists elsewhere so a new PR can be made if Lee or someone else shows a good template pattern for these C APIs to be generated.  Resolving this conversation.   Use the rocblas project if making a ticket for this.

library/src/blas2/rocblas_syr.hpp
YvanMokwinski: > I'm late to this conversation, but I think that not requiring explicit template type arguments (having them deduced instead) is preferred. `decltype()` can be used to "compute" the type based on another expression, without requiring the user to explicitly specify the type.
> 
> I don't understand: "Yes I thought about that but I prefer to see the base type as a template". If you're talking about `rocblas_float_complex` -> `float` and `rocblas_double_complex` -> `double`, then I think traits are a better way to do this -- in one of the header files, define a `real_type` template:
> 
> ```
> template<typename>
> struct rocblas_real_type
> {
> };
> 
> template<>
> struct rocblas_real_type<rocblas_float_complex>
> {
>     using type = float;
> };
> 
> template<>
> struct rocblas_real_type<rocblas_double_complex>
> {
>     using type = double;
> };
> 
> using T = typename rocblas_real_type<U>::type;
> ```

I came with a document on these topics.

library/src/blas2/rocblas_syr.cpp
leekillough: > Can't see how to easily replace the extern C symbols with an instatiated template. This pattern exists elsewhere so a new PR can be made if Lee or someone else shows a good template pattern for these C APIs to be generated. Resolving this conversation. Use the rocblas project if making a ticket for this.

You can't create `extern "C"` template instantiations, since their names are always C++-binding (i.e., mangled), but you can avoid some of the repetitious code such as the `try/catch` block with a template.

I imagine this will come up in the solver proposal. I'll comment more later.

PR #942, opened 2020-01-17
Functional SPR

daineAMD:
Resolves SWDEV-194775


library/src/blas2/rocblas_spr.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+template <typename T>
+__device__ void
+    spr_kernel_calc(bool upper, rocblas_int n, T alpha, const T* x, rocblas_int incx, T* AP)
+{
+    rocblas_int tx = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
+    rocblas_int ty = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
+
+    int index = upper ? ((ty * (ty + 1)) / 2) + tx : ((ty * (2 * n - ty + 1)) / 2) + (tx - ty);
+    if(index < 0 || index > (n * (n + 1)) / 2)
```
YvanMokwinski: index = (n(n+1))/2 is okay?

library/src/blas2/rocblas_spr.cpp
```diff
@@ -0,0 +1,181 @@
...
+        if(uplo != rocblas_fill_lower && uplo != rocblas_fill_upper)
+            return rocblas_status_not_implemented;
+        if(n < 0 || !incx)
+            return rocblas_status_invalid_size;
+        if(!n)
+            return rocblas_status_success;
+        if(!x || !AP)
+            return rocblas_status_invalid_pointer;
+
+        constexpr rocblas_int    offset_x = 0, offset_A = 0, batch_count = 1;
```
YvanMokwinski: static constexpr?

library/src/blas2/rocblas_spr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    auto*       AP    = load_ptr_batch(APa, hipBlockIdx_z, shift_A, stride_A);
+    const auto* x     = load_ptr_batch(xa, hipBlockIdx_z, shift_x, stride_x);
+    auto        alpha = load_scalar(alphaa);
+
+    spr_kernel_calc(upper, n, alpha, x, incx, AP);
+}
+
+/**
+ * TScal     is always: const T* (either host or device)
+ * TConstPtr is either: const T* OR const T* const*
```
YvanMokwinski: but what is T here ;) ?

library/src/blas2/rocblas_spr.hpp
daineAMD: Looks like that check is pointless as the tx and ty checks ensure that index is indeed in that range (and exclusively less than n(n+1)/2)

clients/include/flops.hpp
```diff
@@ -262,6 +262,25 @@ constexpr double ger_gflop_count(rocblas_int m, rocblas_int n)
...
 }
 
+/* \brief floating point counts of SPR */
+template <typename T>
+constexpr double spr_gflop_count(rocblas_int n)
+{
+    return (n * (n + 1) + n) / 1e9;
+}
+
+template <>
```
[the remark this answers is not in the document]
daineAMD: This is from http://www.netlib.org/blas/sblas2time.f, I will go over it tomorrow to make sure

clients/include/near.hpp
```diff
@@ -281,4 +281,73 @@ inline void near_check_general(rocblas_int                         M,
...
+inline void near_check_general(rocblas_int             M,
+                               rocblas_int             N,
+                               rocblas_int             batch_count,
+                               rocblas_int             lda,
+                               rocblas_double_complex* hCPU[],
+                               rocblas_double_complex* hGPU[],
+                               double                  abs_error)
+{
+    NEAR_CHECK_B(M, N, batch_count, lda, hCPU, hGPU, abs_error, NEAR_ASSERT_COMPLEX);
+}
```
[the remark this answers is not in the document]
daineAMD: It looks like none of the unit/near checks have `const`s. I can go ahead and make a different PR for this unless there's a reason that they don't have this.

clients/include/testing_spr.hpp
```diff
@@ -0,0 +1,191 @@
...
+            {
+                unit_check_general<T>(1, size_A, 1, hA_gold, hA_1);
+                unit_check_general<T>(1, size_A, 1, hA_gold, hA_2);
+            }
+            else
+            {
+                const double tol = N * sum_error_tolerance<T>;
+                near_check_general<T>(1, size_A, 1, hA_gold, hA_1, tol);
+                near_check_general<T>(1, size_A, 1, hA_gold, hA_2, tol);
+            }
```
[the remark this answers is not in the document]
leekillough: > 
> 
> I would put
> line 140: assert(std::is_same<T, rocblas_float_complex>{} || std::is_same<T, rocblas_double_complex>{});
> 
> It would help if anything ...

I would use `static_assert` instead of `assert`.

clients/include/testing_spr_strided_batched.hpp
```diff
@@ -79,7 +79,7 @@ void testing_spr_strided_batched(const Arguments& arg)
     }
 
     size_t abs_incx = incx >= 0 ? incx : -incx;
-    size_t size_A   = size_t(N * (N + 1) / 2);
+    size_t size_A   = size_t((N * (N + 1)) / 2);
```
leekillough:
```
size_t size_A   = size_t((N * (N + 1)) / 2);
```
Should be:
```
size_t size_A = size_t(N) * (N+1) / 2;
```
i.e., `N` needs to be converted to `size_t` before the multiplication to avoid overflow in case `N` is `rocblas_int`, and the outer parentheses are redundant.

clients/include/testing_spr_batched.hpp
```diff
@@ -67,7 +67,7 @@ void testing_spr_batched(const Arguments& arg)
         return;
     }
 
-    size_t size_A = N * (N + 1) / 2;
+    size_t size_A = (N * (N + 1)) / 2;
```
leekillough: Should be `size_t size_A = size_t(N) * (N + 1) /2;`

clients/include/testing_spr.hpp
```diff
@@ -63,7 +63,7 @@ void testing_spr(const Arguments& arg)
     }
 
     size_t abs_incx = incx >= 0 ? incx : -incx;
-    size_t size_A   = N * (N + 1) / 2;
+    size_t size_A   = (N * (N + 1)) / 2;
```
leekillough: `size_t size_A = size_t(N) * (N + 1) / 2;`

clients/include/flops.hpp
```diff
@@ -262,6 +262,25 @@ constexpr double ger_gflop_count(rocblas_int m, rocblas_int n)
     return (2.0 * m * n + min) / 1e9;
 }
 
+/* \brief floating point counts of SPR */
+template <typename T>
+constexpr double spr_gflop_count(rocblas_int n)
+{
+    return (n * (n + 1) + n) / 1e9;
```
leekillough: Should be
```
return (double(n) * (n + 1) + n) / 1e9;
```

i.e., convert to `double` before doing calculations. Only one conversion at the innermost level is needed; the rest are implicit.

clients/include/near.hpp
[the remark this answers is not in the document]
leekillough: Yes, the unit/near checks should have pointers to `const`-qualified data, at both the outer and inner levels, e.g.,
```
const T *const a[];
```

clients/include/testing_spr.hpp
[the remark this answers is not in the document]
daineAMD: If I'm not mistaken I can't use `static_assert` here unless this was an `if constexpr` statement, which we aren't using yet.
I've changed the order of the `if` statement to default to using `unit_check`s. I think this should solve any problems as if any new types are introduced we default to strict checking, rather than the more lenient `near_check`.

daineAMD:
CI looks good besides a sles packaging issue and a hipclang build issue involving omp.h, not related to this PR so merging.

PR #944, opened 2020-01-20
Added tuned arcturus transformer configs

amdkila:


amdkila:
@zaliu full rocblas-test suite has passed

PR #947, opened 2020-01-21
Ger x

TorreZuk:
adds complex variants geru gerc refactors ger
wildcard instead of explicit yaml file dependencies


library/src/blas2/rocblas_ger.cpp
```diff
@@ -9,14 +9,37 @@
 
 namespace
 {
-    template <typename>
+    template <bool, typename>
     constexpr char rocblas_ger_name[] = "unknown";
     template <>
-    constexpr char rocblas_ger_name<float>[] = "rocblas_sger";
+    constexpr char rocblas_ger_name<false, float>[] = "rocblas_sger";
```
YvanMokwinski: reversed order ? sometimes it is <bool,typename> and sometimes <typename, bool>.
would worth it to make the code uniform.

library/src/blas2/rocblas_ger.hpp
```diff
@@ -35,11 +35,42 @@ __global__ void ger_kernel(rocblas_int    m,
         const T* __restrict__ x = load_ptr_batch(xa, hipBlockIdx_z, shiftx, stridex);
         const T* __restrict__ y = load_ptr_batch(ya, hipBlockIdx_z, shifty, stridey);
 
-        A[tx + lda * ty] += alpha * x[tx * incx] * y[ty * incy];
+        A[tx + lda * ty] += alpha * x[tx * incx] * T(CONJ ? conj(y[ty * incy]) : y[ty * incy]);
```
YvanMokwinski: You don't need T do you?

T(CONJ ? conj(y[ty * incy]) : y[ty * incy])

library/src/blas2/rocblas_ger.cpp
leekillough: You need to use `rocblas_operation` instead of boolean. When comparing against transpose, you compare against `rocblas_operation_none`, i.e., use:
```
if(trans != rocblas_operation_none)
```
instead of:
```
if(trans == rocblas_operation_transpose)
```
In real functions, `rocblas_operation_conjugate_transpose` should be considered identical to `rocblas_operation_transpose`.

library/src/blas2/rocblas_ger.hpp
TorreZuk: I don't think so, let me test again without.

library/src/blas2/rocblas_ger.cpp
TorreZuk: I followed the two existing patterns, in kernels the code modifiers came first, while in testing they came second (see dot).    I think it may have been for default value use was what led to the latter, but I don't have that need as it can be explicitly set.   I am fine with either or both, someone will complain either way as it doesn't match elsewhere in the framework.  Thus I chose to match each framework's implicit convention.

clients/gtest/gerc_gtest.cpp
```diff
@@ -0,0 +1,144 @@
+/* ************************************************************************
```
daineAMD: Since ger/gerc/geru are basically the same function, would it be better to have a single `ger_gtest.cpp` rather than splitting it up into three?

library/include/rocblas-functions.h
```diff
@@ -5609,12 +5610,57 @@ ROCBLAS_EXPORT rocblas_status rocblas_dger(rocblas_handle handle,
...
+                                            const rocblas_double_complex* y,
+                                            rocblas_int                   incy,
+                                            rocblas_double_complex*       A,
+                                            rocblas_int                   lda);
+
 /*! \brief BLAS Level 2 API
 
     \details
-    xGER_BATCHED performs a batch of the matrix-vector operations
+    xGER.xGERU,xGERC_BATCHED performs a batch of the matrix-vector operations
```
daineAMD: `xGER`**,**`xGERU,...` (comma instead of period)

clients/gtest/CMakeLists.txt
```diff
@@ -195,7 +197,7 @@ endif( )
 set( ROCBLAS_TEST_DATA "${PROJECT_BINARY_DIR}/staging/rocblas_gtest.data")
 add_custom_command( OUTPUT "${ROCBLAS_TEST_DATA}"
                     COMMAND ../common/rocblas_gentest.py -I ../include rocblas_gtest.yaml -o "${ROCBLAS_TEST_DATA}"
-                    DEPENDS ../common/rocblas_gentest.py rocblas_gtest.yaml ../include/rocblas_common.yaml known_bugs.yaml blas1_gtest.yaml gemm_gtest.yaml gemm_batched_gtest.yaml gemm_strided_batched_gtest.yaml trmv_gtest.yaml gbmv_gtest.yaml gemv_gtest.yaml hbmv_gtest.yaml hemv_gtest.yaml hpmv_gtest.yaml tbmv_gtest.yaml tpmv_gtest.yaml symv_gtest.yaml syr_gtest.yaml ger_gtest.yaml trsm_gtest.yaml trtri_gtest.yaml geam_gtest.yaml set_get_vector_gtest.yaml set_get_matrix_gtest.yaml trmm_gtest.yaml trsv_gtest.yaml logging_mode_gtest.yaml set_get_pointer_mode_gtest.yaml
+                    DEPENDS ../common/rocblas_gentest.py ../include/rocblas_common.yaml *.yaml
```
daineAMD: Finally, the end of merge conflicts in this file!

clients/include/flops.hpp
```diff
@@ -255,9 +255,10 @@ constexpr double symv_gflop_count(rocblas_int n)
 }
 
 /* \brief floating point counts of GER */
-template <typename T>
+template <typename T, bool CONJ>
 constexpr double ger_gflop_count(rocblas_int m, rocblas_int n)
 {
+    // conjugate not counted
```
daineAMD: Need to have a different calculation for complex types.

clients/gtest/gerc_gtest.cpp
TorreZuk: I did that and it was so ugly as the 3 names need to be switched between it was much simpler making 3 test instantiations.   I think if we refactor the test instantiation class this may be doable in the future, but can likely use one class for all functions.

clients/gtest/CMakeLists.txt
TorreZuk: I couldn't take it anymore.

clients/include/flops.hpp
TorreZuk: Thanks forgot, will add it.

clients/gtest/CMakeLists.txt
YvanMokwinski: :)

library/src/blas2/rocblas_ger.cpp
TorreZuk: Lee this bool only controls conjugation not any transpose operation, see dot for the pattern bool CONJ, or Level 3 bool BATCHED variants.

library/src/blas2/rocblas_ger.hpp
TorreZuk: Works without so resolving

clients/gtest/geruc_gtest.yaml
```diff
@@ -0,0 +1,156 @@
...
+- name: geruc_small
+  category: quick
+  function:
+  - geru
+  - gerc
+  precision: *single_double_precisions_complex
+  matrix_size: *small_matrix_size_range
+  incx_incy: *incx_incy_range
+  alpha: [ -0.5, 2.0, 0.0 ]
+
```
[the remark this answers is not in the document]
TorreZuk: I'll add some imaginary components, also likely have to reduce the matrix sizes as we are getting pretty long

PR #951, opened 2020-01-22
Functional Spr2

daineAMD:
implements batched/strided_batched/regular float (sspr2) and double (dspr2) versions of spr2. Resolves SWDEV-219106.


clients/include/testing_spr2.hpp
```diff
@@ -0,0 +1,200 @@
...
+    // allocate memory on device
+    device_vector<T> dA_1(size_A);
+    device_vector<T> dx(size_x);
+    device_vector<T> dy(size_y);
+    CHECK_HIP_ERROR(dA_1.memcheck());
+    CHECK_HIP_ERROR(dx.memcheck());
+    CHECK_HIP_ERROR(dy.memcheck());
+
+    EXPECT_ROCBLAS_STATUS(rocblas_spr2<T>(handle, uplo, N, &alpha, nullptr, incx, dy, incy, dA_1),
+                          rocblas_status_invalid_pointer);
```
TorreZuk: Can pass uplo == rocblas_fill_full and check for invalid_value return?   I'm doing that in all new functions

daineAMD: Yup, I'll add that.

PR #953, opened 2020-01-24
add complex rocblas_ctrmm and rocblas_ztrmm

amcamd:
- add complex functions rocblas_ctrmm and rocblas_ztrmm.
- add CONJ template parameter to internal function rocblas_copy_template. This provides the option for  conjugation with copy.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -353,11 +353,11 @@ struct perf_blas<T,
                 {"tbmv", testing_tbmv<T>},
                 {"tbmv_batched", testing_tbmv_batched<T>},
                 {"tbmv_strided_batched", testing_tbmv_strided_batched<T>},
-                {"trmm", testing_trmm<T>},
 #if BUILD_WITH_TENSILE
                 {"gemm", testing_gemm<T>},
                 {"gemm_batched", testing_gemm_batched<T>},
                 {"gemm_strided_batched", testing_gemm_strided_batched<T>},
+                {"trmm", testing_trmm<T>},
 #endif
               };
         run_function(map, arg);
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -353,6 +353,7 @@ struct perf_blas<T,
                 {"tbmv", testing_tbmv<T>},
                 {"tbmv_batched", testing_tbmv_batched<T>},
                 {"tbmv_strided_batched", testing_tbmv_strided_batched<T>},
+                {"trmm", testing_trmm<T>},
 #if BUILD_WITH_TENSILE
                 {"gemm", testing_gemm<T>},
                 {"gemm_batched", testing_gemm_batched<T>},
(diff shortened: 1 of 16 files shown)


clients/gtest/trmm_gtest.yaml
```diff
@@ -18,7 +18,12 @@ Definitions:
     - { M:  1024, N:  1024, lda:  1024, ldb:  1024 }
     - { M:  2000, N:  2000, lda:  2000, ldb:  2000 }
 
-  - &alpha_range [ 1.0, -5.0 ]
+  - &alpha_range [ 1.0, -3.0 ]
+
+  - &complex_alpha_range
```
mahmoodw: Daine had previously suggested that we can keep only the complex version and the imaginary portion will be ignored for the real cases. This is a bit cleaner but if you are trying to control the quantity of tests then maybe it should remain split up.

clients/gtest/trmm_gtest.yaml
```diff
@@ -92,15 +97,25 @@ Definitions:
     - { M:  44928,  N: 384,  lda: 44928, ldb:   44928 }
     - { M:  53376,  N: 384,  lda: 53376, ldb:   53376 }
 
-
 Tests:
+- name: trmm_small
+  category: quick
+  function: trmm
```
mahmoodw: if we combine the alpha ranges then we don't need separate categories here

clients/include/unit.hpp
```diff
@@ -409,4 +409,13 @@ inline void trsm_err_res_check(T max_error, rocblas_int M, T forward_tolerance,
 #endif
 }
 
+template <typename T>
+double get_epsilon()
```
mahmoodw: I noticed some helpers were being created to help identity the underlying type in complex cases. I was thinking of switching over to using these helpers instead of creating a new helper for the epsilon. You can see the addition here:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/8d8d415db5c04c62e835dca68948b3d0aa2d6522/library/src/include/utility.h#L367

clients/benchmarks/client.cpp
```diff
@@ -353,6 +353,7 @@ struct perf_blas<T,
                 {"tbmv", testing_tbmv<T>},
                 {"tbmv_batched", testing_tbmv_batched<T>},
                 {"tbmv_strided_batched", testing_tbmv_strided_batched<T>},
+                {"trmm", testing_trmm<T>},
```
mahmoodw: Should be moved down in the build with tensile section?

mahmoodw:
I've made some comments - besides from that tests look like they are passing and the rest LGTM.

clients/benchmarks/client.cpp
amcamd: Thanks for catch, moved to build with tensile section

clients/include/unit.hpp
amcamd: Will switch to use specializations once Daine's PR is merged

clients/gtest/trmm_gtest.yaml
amcamd: I will keep it split to reduce the number of tests

clients/gtest/trmm_gtest.yaml
amcamd: I think keeping them separate will reduce the number of tests. We have 2 values in alpha_range, and 3 values in complex_alpha renge.

PR #955, opened 2020-01-24
Funcitonal HPR

daineAMD:
See [documentation](http://www.netlib.org/lapack/explore-html/d6/dee/group__complex__blas__level2_gac98100eee08124fb7a5d55effbb85a65.html#gac98100eee08124fb7a5d55effbb85a65).
Pretty much spr with tiny logic changes in the kernel.

### DIFF
--- a/clients/gtest/hpr_gtest.cpp
+++ b/clients/gtest/hpr_gtest.cpp
@@ -64,8 +64,9 @@ namespace
             }
             else
             {
-                name << '_' << (char)std::toupper(arg.uplo) << '_' << arg.N << '_' << arg.alpha
-                     << '_' << arg.incx;
+                // For alpha: T doesn't really matter here, just whether it's real or complex, in hpr's case it's always complex.
+                name << '_' << (char)std::toupper(arg.uplo) << '_' << arg.N << '_'
+                     << arg.get_alpha<rocblas_double_complex>() << '_' << arg.incx;
 
                 if(HPR_TYPE == HPR_STRIDED_BATCHED)
                     name << '_' << arg.stride_x;
--- a/clients/gtest/hpr_gtest.cpp
+++ b/clients/gtest/hpr_gtest.cpp
@@ -1,0 +1,138 @@
+/* ************************************************************************
+ * Copyright 2018-2020 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "rocblas_data.hpp"
+#include "rocblas_datatype2string.hpp"
+#include "rocblas_test.hpp"
+#include "testing_hpr.hpp"
+#include "testing_hpr_batched.hpp"
+#include "testing_hpr_strided_batched.hpp"
+#include "type_dispatch.hpp"
+#include <cctype>
+#include <cstring>
+#include <type_traits>
+
+namespace
+{
+    // possible test cases
+    enum hpr_test_type
+    {
+        HPR,
+        HPR_BATCHED,
+        HPR_STRIDED_BATCHED,
+    };
+
+    //hpr test template
+    template <template <typename...> class FILTER, hpr_test_type HPR_TYPE>
+    struct hpr_template : RocBLAS_Test<hpr_template<FILTER, HPR_TYPE>, FILTER>
+    {
+        // Filter for which types apply to this suite
+        static bool type_filter(const Arguments& arg)
+        {
+            return rocblas_simple_dispatch<hpr_template::template type_filter_functor>(arg);
+        }
+
+        // Filter for which functions apply to this suite
+        static bool function_filter(const Arguments& arg)
+        {
+            switch(HPR_TYPE)
+            {
+            case HPR:
+                return !strcmp(arg.function, "hpr") || !strcmp(arg.function, "hpr_bad_arg");
+            case HPR_BATCHED:
+                return !strcmp(arg.function, "hpr_batched")
+                       || !strcmp(arg.function, "hpr_batched_bad_arg");
+            case HPR_STRIDED_BATCHED:
+                return !strcmp(arg.function, "hpr_strided_batched")
+                       || !strcmp(arg.function, "hpr_strided_batched_bad_arg");
+            }
+            return false;
+        }
+
+        // Google Test name suffix based on parameters
+        static std::string name_suffix(const Arguments& arg)
+        {
+            RocBLAS_TestName<hpr_template> name;
+
+            name << rocblas_datatype2string(arg.a_type);
+
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
+            }
+            else
+            {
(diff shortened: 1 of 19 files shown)


clients/include/testing_hpr.hpp
```diff
@@ -0,0 +1,190 @@
...
+                          rocblas_status_invalid_pointer);
+
+    EXPECT_ROCBLAS_STATUS((rocblas_hpr<T, U>)(nullptr, uplo, N, &alpha, dx, incx, dA_1),
+                          rocblas_status_invalid_handle);
+}
+
+template <typename T>
+void testing_hpr(const Arguments& arg)
+{
+    using U = decltype(std::real(T(0)));
```
YvanMokwinski: It would be nice to make this functionality available globally ... with traits ...

template < typename T >
struct precision_traits
{
  using comp_t = T;
};

template <>
struct precision_traits<rocblas_float_complex>
{
  using comp_t = float;
};

template <>
struct precision_traits<rocblas_double_complex>
{
  using comp_t = double;
};

template < typename T >
using precision_part_t = typename precision_traits< T >::comp_t;

Then you fetch your type with precision_part_t< T >.

clients/include/testing_hpr.hpp
```diff
@@ -0,0 +1,190 @@
...
+        gpu_time_used = get_time_us(); // in microseconds
+
+        for(int iter = 0; iter < number_hot_calls; iter++)
+        {
+            rocblas_hpr<T, U>(handle, uplo, N, &h_alpha, dx, incx, dA_1);
+        }
+
+        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops    = hpr_gflop_count<T>(N) / gpu_time_used * 1e6;
+        rocblas_bandwidth = (2.0 * N * (N + 1)) / 2 * sizeof(T) / gpu_time_used / 1e3;
```
YvanMokwinski: shouldn't we have a xxx_bandwith function?

library/src/blas2/rocblas_hpr.hpp
```diff
@@ -0,0 +1,125 @@
...
+    rocblas_int ty = hipBlockIdx_y * hipBlockDim_y + hipThreadIdx_y;
+
+    int index = upper ? ((ty * (ty + 1)) / 2) + tx : ((ty * (2 * n - ty + 1)) / 2) + (tx - ty);
+
+    if(upper ? ty < n && tx < ty : tx < n && ty < tx)
+        AP[index] += alpha * x[tx * incx] * conj(x[ty * incx]);
+    else if(tx == ty && tx < n)
+    {
+        T x_real  = std::real(x[tx * incx]);
+        T x_imag  = std::imag(x[tx * incx]);
```
YvanMokwinski: T is complex and we fetch the real part as a complex?

clients/include/testing_hpr.hpp
daineAMD: I think Torre is working on this.

library/src/blas2/rocblas_hpr.hpp
daineAMD: Good catch,  will fix.

clients/gtest/hpr_gtest.cpp
```diff
@@ -0,0 +1,139 @@
...
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
+            }
+            else
+            {
+                // For alpha: T doesn't really matter here, just whether it's real or complex, in hpr's case it's always complex.
+                name << '_' << (char)std::toupper(arg.uplo) << '_' << arg.N << '_'
+                     << arg.get_alpha<rocblas_double_complex>() << '_' << arg.incx;
+
```
TorreZuk: Looks good if we standardize on this.

clients/include/testing_hpr.hpp
TorreZuk: Sorry you need to wait for PR #956 then add to bytes.hpp

clients/include/testing_hpr.hpp
daineAMD: Added this in library/.../utility.h now. Willing to move somewhere else if people have concerns. Considered putting into rocblas-complex-types.h or in clients/.../utility.h

clients/include/testing_hpr_strided_batched.hpp
```diff
@@ -90,7 +90,7 @@ void testing_hpr_strided_batched_bad_arg()
 template <typename T>
 void testing_hpr_strided_batched(const Arguments& arg)
 {
-    using U = decltype(std::real(T(0)));
+    using U = typename rocblas_real_type<T>::type;
```
YvanMokwinski: There is some way to write things (see my comments) where we should not have to specify typename here.

template < typename T >
using rocblas_real_t = typename rocblas_real_type< T >::type;

Then, you can write

rocblas_real_t< T > x_real = 78.0;

or 

using U = rocblas_real_t< T >.
U x_real = 78.0;

library/src/include/utility.h
```diff
@@ -343,6 +343,23 @@ __device__ __host__ inline rocblas_half rocblas_abs(rocblas_half x)
     return t.x;
 }
 
+template <typename>
+struct rocblas_real_type
+{
+};
```
YvanMokwinski: should we have default for float and double?

clients/include/testing_hpr_strided_batched.hpp
YvanMokwinski: I am not a big fan to specify typename each time you want to use the type.

library/src/include/utility.h
daineAMD: I could change it to default to the passed-in type, might be useful in the future. Or could just leave it in case we want to change it in the future since we aren't using it now.

leekillough: Why does this have to be so complicated?

Just define specializations:
```
template <typename>
struct rocblas_real_type
{
};

template <>
struct rocblas_real_type<rocblas_float_complex>
{
   using type = float;
};

template <>
struct rocblas_real_type<rocblas_double_complex>
{
   using type = double;
};

daineAMD: > If we think we'll need to use them at some point for T=double/float (which we might) I'll implement them to use their own type.
> I think we should leave the default one empty as it is now.

The only difference to what you described is that I've defined specialisations for float and double. For `hpr`, this isn't needed, but if we refactor say, `scal`, we could use these since we implement both real and complex types.
If we don't want these now, that's fine, we just need to have an agreement so I don't have to keep changing it back and forth for this PR :)

YvanMokwinski: What is the definition of 'so complicated' ?

daineAMD:
I think I've addressed all comments for now. Looks like CI is passing so wouldn't mind an approval or re-review :)

YvanMokwinski:
> I think I've addressed all comments for now. Looks like CI is passing so wouldn't mind an approval or re-review :)

I will have a look and make a decision on whether or not allowing to merge.

PR #956, opened 2020-01-24
L2 sXmv

TorreZuk:
* adds symv, sbmv, spmv, only symv handles complex
* adds bytes.hpp to match flops.hpp

### DIFF
--- a/clients/gtest/symv_gtest.yaml
+++ b/clients/gtest/symv_gtest.yaml
@@ -6,11 +6,10 @@ Definitions:
   - &medium_matrix_size_range
     - { N:    -1, lda:   -1 }
     - { N:    10, lda:    2 }
-    - { N:   500, lda:  500 }
+    - { N:    33, lda:   33 }
+    - { N:   300, lda:  600 }
 
   - &large_matrix_size_range
-    - { N:  1000, lda:  1000 }
-    - { N:  2000, lda:  2000 }
     - { N:  4011, lda:  4011 }
     - { N:  8000, lda:  8000 }
 
@@ -18,15 +17,26 @@ Definitions:
     - { incx:   1, incy:   1 }
     - { incx:   0, incy:  -1 }
     - { incx:   2, incy:   1 }
-    - { incx:  10, incy: 100 }
+    - { incx:  -1, incy:  -1 }
 
   - &alpha_beta_range
     - { alpha:  1.0, beta:  0.0 }
     - { alpha: -1.0, beta: -1.0 }
     - { alpha:  2.0, beta:  1.0 }
-    - { alpha:  0.0, beta:  1.0 }
+    - { alpha:  0.0, beta:  1.0 } # quick success
+
+  - &complex_alpha_beta_range
+    - { alpha:  2, alphai:  0, beta:  2, betai:  0 }
+    - { alpha:  0, alphai:  2, beta:  0, betai:  1 }
+    - { alpha:  1, alphai:  2, beta:  1, betai:  2 }
+    - { alpha:  1.0, alphai:  1, beta:  0.5, betai:  1 }
 
 Tests:
+- name: symv_bad
+  category: pre_checkin
+  function: symv_bad_arg
+  precision: *single_precision
+
 - name: symv_medium
   category: pre_checkin
   function: symv
@@ -36,6 +46,15 @@ Tests:
   incx_incy: *incx_incy_range
   alpha_beta: *alpha_beta_range
 
+- name: symv_medium_c
+  category: pre_checkin
+  function: symv
+  precision: *single_double_precisions_complex
+  uplo: [ U, L ]
+  matrix_size: *medium_matrix_size_range
+  incx_incy: *incx_incy_range
+  alpha_beta: *complex_alpha_beta_range
+
 - name: symv_large
   category: nightly
   function: symv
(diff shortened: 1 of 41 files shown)


YvanMokwinski:
You can't split it into few PRs? :) (41 files ...)

TorreZuk:
Weekend reading.  Once you've read one you've read them all  :-)

clients/include/testing_sbmv.hpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    // Initial Data on CPU
+    rocblas_seedrand();
+    rocblas_init_symmetric<T>(hA, N, lda);
+
+    rocblas_init<T>(hx, 1, N, abs_incx);
+    rocblas_init<T>(hy, 1, N, abs_incy);
+
+    // for (int i = 0; i < N; i++)
```
daineAMD: Debugging code can be removed.

clients/include/testing_sbmv.hpp
```diff
@@ -0,0 +1,278 @@
...
+    device_vector<T> dy(size_Y);
+    if(!dA || !dx || !dy)
+    {
+        CHECK_HIP_ERROR(hipErrorOutOfMemory);
+        return;
+    }
+
+    // Initial Data on CPU
+    rocblas_seedrand();
+    rocblas_init_symmetric<T>(hA, N, lda);
```
daineAMD: This is being treated as a banded matrix, so there's no use using a special initialization, the symmetry won't be recognized.

If this were treated as a regular matrix, I still don't think we should be initializing it as a symmetric matrix in test code. For symv, for example, we could be accessing elements on the wrong side of the diagonal and still get the correct result.
I think a better way to do it is to just initialize the entire matrix with random data as we do usually, then if we happen to be accessing data on the wrong side of the diagonal, this should be caught by the test.

clients/include/testing_spmv.hpp
```diff
@@ -0,0 +1,246 @@
...
+    host_vector<T> beta(1);
+    alpha[0] = arg.alpha;
+    beta[0]  = arg.beta;
+
+    rocblas_fill uplo = char2rocblas_fill(arg.uplo);
+
+    size_t abs_incx = incx >= 0 ? incx : -incx;
+    size_t abs_incy = incy >= 0 ? incy : -incy;
+
+    size_t size_A = size_t(N) * N;
```
daineAMD: The size of A can be `(n * (n + 1) / 1)`.

clients/include/testing_spmv.hpp
```diff
@@ -0,0 +1,246 @@
...
+    device_vector<T> dy(size_Y);
+    if(!dA || !dx || !dy)
+    {
+        CHECK_HIP_ERROR(hipErrorOutOfMemory);
+        return;
+    }
+
+    // Initial Data on CPU
+    rocblas_seedrand();
+    rocblas_init_symmetric<T>(hA, N, N);
```
daineAMD: Same comment as in sbmv.
Since A is more-or-less a vector here (of size (n * (n+1) / 2)) in the packed version of these algorithms, we can initialize it like a regular vector, we don't have to worry about symmetry, the implementation of spmv handle it correctly either way.

clients/include/testing_spmv_batched.hpp
```diff
@@ -0,0 +1,314 @@
...
+
+    CHECK_HIP_ERROR(dA.memcheck());
+    CHECK_HIP_ERROR(dx.memcheck());
+    CHECK_HIP_ERROR(dy.memcheck());
+
+    // Initial Data on CPU
+    rocblas_seedrand();
+    for(int i = 0; i < batch_count; i++)
+    {
+        //rocblas_init_symmetric<T>(hA[i], N, lda);
```
daineAMD: Hmmm

clients/include/testing_symv.hpp
```diff
@@ -76,91 +142,122 @@ void testing_symv(const Arguments& arg)
...
+        cblas_symv<T>(uplo, N, alpha[0], hA, lda, hx, incx, beta[0], hg, incy);
+
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = symv_gflop_count<T>(N) / cpu_time_used * 1e6;
     }
 
-    // copy output from device to CPU
-    hipMemcpy(hy, dy, sizeof(T) * N * incy, hipMemcpyDeviceToHost);
+    // clear non-fill half
+    rocblas_clear_symmetric(uplo, hA.data(), N, lda);
```
daineAMD: Ah, I see! Initializing it as symmetric makes sense to me now haha. Still don't think we need to use the symmetric initialization in sbmv or spmv, though.

Also, should this be done before the call to `cblas_symv` so we're actually passing in the same information to the cblas version and our version?

library/include/rocblas-functions.h
```diff
@@ -5525,27 +5525,804 @@ ROCBLAS_EXPORT rocblas_status rocblas_dtrsv_strided_batched(rocblas_handle    ha
...
+              specifies whether the upper 'rocblas_fill_upper' or lower 'rocblas_fill_lower'
+              if rocblas_fill_upper, the lower part of A is not referenced
+              if rocblas_fill_lower, the upper part of A is not referenced
+    @param[in]
+    n         [rocblas_int]
+    @param[in]
+    alpha
+              specifies the scalar alpha
+    @param[in]
+    A         pointer storing matrix A on the GPU
```
daineAMD: Since we're finishing up L2 functions, maybe next week we should discuss documentation. I think it might be better to have a specific section in the documentation on packed/banded matrices that we can reference here rather than duplicating it (I think you mentioned this same idea previously).

library/src/blas2/rocblas_sbmv.cpp
```diff
@@ -0,0 +1,173 @@
...
+    try                                                                                          \
+    {                                                                                            \
+        return rocblas_sbmv_impl<T_>(handle, uplo, n, k, alpha, A, lda, x, incx, beta, y, incy); \
+    }                                                                                            \
+    catch(...)                                                                                   \
+    {                                                                                            \
+        return exception_to_rocblas_status();                                                    \
+    }
+
+IMPL(rocblas_ssbmv, float);
```
daineAMD: I see that cuBLAS only supports float/double for sbmv and spmv, however LAPACK does in fact support complex. Not sure if we should also be supporting them or not.

library/src/blas2/rocblas_sbmv.hpp
```diff
@@ -0,0 +1,172 @@
...
+                           beta,
+                           stride_beta,
+                           y,
+                           shifty,
+                           incy,
+                           stridey);
+    }
+    else
+    {
+        if(!*alpha && *beta == 1)
```
daineAMD: We are ignoring the possibility of stride_alpha/stride_beta != 0 here. That's fine for now, I just think maybe we should remove them as parameters for now? Alternatively this can be moved into `rocblas_sbmv_kernel`.

library/src/blas2/rocblas_sbmv.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                    ptrdiff_t      shiftx,
+                                    rocblas_int    incx,
+                                    rocblas_stride stridex,
+                                    V              beta_device_host,
+                                    rocblas_stride stride_beta,
+                                    W              ya,
+                                    ptrdiff_t      shifty,
+                                    rocblas_int    incy,
+                                    rocblas_stride stridey)
+{
```
daineAMD: hbmv (gemv) _might_ have a quicker implementation if you want some inspiration. We could probably delay this for later if we want, though, since we might be able to come up with something better for banded (or packed) matrices.

library/src/blas2/rocblas_spmv.hpp
```diff
@@ -0,0 +1,154 @@
...
+                                    rocblas_int    incx,
+                                    rocblas_stride stridex,
+                                    V              beta_device_host,
+                                    rocblas_stride stride_beta,
+                                    W              ya,
+                                    ptrdiff_t      shifty,
+                                    rocblas_int    incy,
+                                    rocblas_stride stridey)
+{
+    ptrdiff_t tx = hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x;
```
daineAMD: Same here as in rocblas_sbmv.hpp

clients/gtest/symv_gtest.yaml
```diff
@@ -36,6 +46,15 @@ Tests:
   incx_incy: *incx_incy_range
   alpha_beta: *alpha_beta_range
 
+- name: symv_medium_c
```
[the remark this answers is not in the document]
TorreZuk: You are probably right, I did have more permutations of complex values (which would be duplicate tests for the real) but I deleted them as the tests were too long.   Will just reuse the unique values for both.

clients/include/bytes.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * ===========================================================================
+ */
+
+/*
+ * ===========================================================================
+ *    level 2 BLAS
+ * ===========================================================================
+ */
+
+inline size_t tri_count(rocblas_int n)
```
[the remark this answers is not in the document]
TorreZuk: tri count is just a "triangle count" look in your PR I think you repeat this code many times.   Can you suggest a better name such that you would use it your functions too.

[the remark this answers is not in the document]
daineAMD: Ah! Makes sense, I should have looked closer. I think tri_count is fairly self-explanatory if you pay an ounce of attention more than I did.

clients/include/testing_sbmv.hpp
TorreZuk: Yes will go with a random fill, before I found the blis functions I was calling gemm to test I'll remove all those.   Clearing half but maybe clear half to nan would be better.

clients/include/bytes.hpp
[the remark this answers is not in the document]
TorreZuk: One might argue it is not worth a function but it seems to come up a lot with symmetric matrices

clients/include/testing_spmv.hpp
TorreZuk: Don't you mean tri_count :-)    Will reduce.

library/include/rocblas-functions.h
TorreZuk: Yes I didn't copy paste your great diagrams but should if we don't have a general section

library/src/blas2/rocblas_sbmv.cpp
TorreZuk: Matched blas and cublas which is SD only.  most complex symmetrics are hermitian complex I presumed so hbmv, hpmv.   Added CZ to symv to match cublas.  Can add them in a different PR if there is interest, maybe Juan would like them.

library/src/blas2/rocblas_sbmv.hpp
TorreZuk: Yes to move them in and slow down the kernel if they won't be used.  Did you remove stride_alpha/beta from others?

daineAMD: I don't think I added stride_alpha or stride_beta in any of the recent ones. I believe it was removed from trsm and everything at the time as well.

TorreZuk: I'll should just delete them to be consistent but will check again with Juan too that it was a blanket decision.

library/src/blas2/rocblas_sbmv.hpp
TorreZuk: Yes these are baseline versions so will need some caching for X.   But first version is functional only not to be addressed in this PR.

library/src/blas2/rocblas_sbmv.hpp
```diff
@@ -139,7 +182,8 @@ rocblas_status rocblas_sbmv_template(rocblas_handle handle,
     }
     else
     {
-        if(!*alpha && *beta == 1)
+        // quick return only for non-batched
```
daineAMD: Seems good.
Alternatively, if we're always passing in stride_alpha = stride_beta = 0 from the C API, you could do a check like:
`if(stride_alpha == 0 && stride_beta == 0 && !*alpha && *beta == 1)`
This way even if batch_count != 1 we can still quick return if we're always using the quick return alpha/beta.

TorreZuk: For batched non-strided you can't rely on strides == 0 test

PR #959, opened 2020-01-27
Add tuned arcturus big transformer configs

amdkila:


amdkila:
I'm not sure what to make of that- it seems to be doing that for all configs. I've also noticed that nightly/trsm.blas3/f32 tests are failing as well, but are not failing on the latest develop (I ran the tests on both branches). I'm looking into the issue now.

amdkila:
yes that's right. I take that it should be safe to merge then?

PR #960, opened 2020-01-27
Functional hpr2

daineAMD:
See [documentation](http://www.netlib.org/lapack/explore-html/d6/dee/group__complex__blas__level2_ga537b8bcd05cbe626ebce0fbc8d66a2d6.html#ga537b8bcd05cbe626ebce0fbc8d66a2d6).

Fairly similar to previous PR for hpr.


library/src/blas2/rocblas_hpr2.hpp
```diff
@@ -0,0 +1,150 @@
+/* ************************************************************************
+ * Copyright 2016-2020 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+#pragma once
+#include "handle.h"
+#include "rocblas.h"
+
+template <typename T, typename U>
```
YvanMokwinski: Is U needed? Is 'T alpha' a problem here?

clients/include/testing_hpr2_strided_batched.hpp
```diff
@@ -0,0 +1,333 @@
...
+                                            stride_y,
+                                            dA_1,
+                                            stride_A,
+                                            batch_count);
+        }
+
+        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        rocblas_gflops    = batch_count * hpr2_gflop_count<T>(N) / gpu_time_used * 1e6;
+        rocblas_bandwidth = batch_count * (2.0 * N * (N + 1)) / 2 * sizeof(T) / gpu_time_used / 1e3;
+
```
TorreZuk: again if you wait put in bytes.hpp

library/src/blas2/rocblas_hpr2.hpp
```diff
@@ -0,0 +1,150 @@
...
+                                    ptrdiff_t      shift_y,
+                                    rocblas_int    incy,
+                                    rocblas_stride stride_y,
+                                    TPtr           APa,
+                                    ptrdiff_t      shift_A,
+                                    rocblas_stride stride_A)
+{
+    rocblas_int num_threads = hipBlockDim_x * hipBlockDim_y * hipBlockDim_z;
+    if(DIM_X * DIM_Y != num_threads)
+        return; // need to launch exactly the number of threads as template parameters indicate.
```
TorreZuk: Is this a check on the compiler?   Or guard against reuse?

library/src/blas2/rocblas_hpr2.hpp
daineAMD: Nope, alpha should be able to be a T, this must have been left from hpr. Will fix.

library/src/blas2/rocblas_hpr2_strided_batched.cpp
```diff
@@ -0,0 +1,209 @@
...
+                            stridey,
+                            "stride_a",
+                            strideA,
+                            "batch_count",
+                            batch_count);
+        }
+
+        if(uplo != rocblas_fill_lower && uplo != rocblas_fill_upper)
+            return rocblas_status_not_implemented;
+        if(n < 0 || !incx || !incy || batch_count < 0)
```
TorreZuk: same mismatch

library/src/blas2/rocblas_hpr2.hpp
YvanMokwinski: I don't like it. No error message, no status, nothing.

YvanMokwinski: Actually, following a discussion with Scott, we should make the kernel crashing.

daineAMD: The check seems redundant to me, could reach out and see why it was initially added into gemv in the first place.

PR #961, opened 2020-01-28
Make new Tensile client the default

wbgilmartin:
Will enable the tensile new client by default

added -r / --no-tensile-host option to install.sh to build with old client.

Fixed a bug which prevented the packaging of the tensile library for getting installed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -26,7 +26,6 @@ rocBLAS build & installation helper script
            --build_dir         Specify name of output directory (default is ./build)
       -n | --no_tensile        Build subset of library that does not require Tensile
       -r | --no-tensile-host   Do not build with Tensile host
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -25,6 +25,7 @@ rocBLAS build & installation helper script
            --hip-clang         Build library for amdgpu backend using hip-clang
            --build_dir         Specify name of output directory (default is ./build)
       -n | --no_tensile        Build subset of library that does not require Tensile
+      -r | --no-tensile-host   Do not build with Tensile host
       -s | --tensile-host      Build with Tensile host
```
TorreZuk: Free up -s option if you are making it default.   Does any CI use -s right now?

PR #963, opened 2020-01-28
Arguments tuple refactor sans mechanical changes

leekillough:
This is all of the non-mechanical changes in PR #922 -- the substance of the changes. It contains:
- The new `rocblas_ostream` for non-interleaved thread-safe IO
- Consolidation of all formatted IO routines into one place (`rocblas_ostream`), with support for YAML or plaintext format via a modifier
- Torre's idea for an `ArgumentModel`, and its initial application to `dot`. `ArgumentModel` accepts a variadic list of `rocblas_argument` `enum`s defining the arguments important to a particular model. Compile-time queries on the `ArgumentModel`'s arguments list can be done to make certain decisions, and compile-time expansion and printing of the `ArgumentModel`'s argument list is done.
- Refactored `Arguments` class, to eliminate duplication by using macros like `FOR_EACH_ARGUMENT`
- Simplification of `std::tuple` helper functions, and moving them out of `logging.h` into a new header `tuple_helper.hpp`.
- Refactoring of `testing_logging.hpp` to use the new `rocblas_ostream` and get correct results
- Preparation to switch some code to use C++17 later (right now C++17 cannot be used if ROCm is still required to support Ubuntu 16.04 LTS).
- A workaround for the deprecated `hipMalloc` and `hipFree` warnings in `samples` (undefining the macros which created the warnings, everywhere except `rocblas-test` and `rocblas-bench`)
- Poisoning `std::cout`, `std::cerr`, `stdout`, `stderr`, `printf`, and a few other functions, but only inside `rocblas-test` and `rocblas-bench`, so that programmers are reminded to use `rocblas_cout` and `rocblas_cerr`, or to open their own `rocblas_ostream` objects, rather than relying on system IO, which does not have the consistent formatting and thread-safety that `rocblas_ostream` has.
- (TODO: Write a multithreaded IO stress-test, to see if multiple threads can write their IO without interleaving it with other threads.)

**This is not supposed to build correctly, nor to be merged. It is just to highlight the important changes in #922 .**

PR #922 has all of these functional changes here, plus:
- Renaming of symbols, like `std::cout` -> `rocblas_cout` and `std::cerr` -> `rocblas_cerr`
- Merges of other PRs (some not merged into develop yet)
- Minor typo fixes
- Addition of missing copyright headers at the top of files
- Changes to `CMakeLists.txt` for build issues
- `clang-format` changes caused by all of the other changes


clients/include/testing_dot_strided_batched.hpp
```diff
@@ -226,7 +226,7 @@ void testing_dot_strided_batched(const Arguments& arg)
         if(arg.norm_check)
         {
             std::cout << "cpu=" << cpu_result << ", gpu_host_ptr=" << rocblas_result_1
-                      << ", gpu_device_ptr=" << rocblas_result_2 << "\n";
+                      << ", gpu_device_ptr=" << rocblas_result_2 << std::endl;
```
amcamd: should this be rocblas_cout   rather than std::cout   ?

leekillough: Yes, that is part of the "mechanical changes", s/std::cout/rocblas_cout/. This PR left out those mechanical changes. std::cout is a poisoned identifier in rocblas-test and rocblas-bench with this PR, so if you use it, those won't compile (that poisoning does not affect outside users of rocBLAS such as samples; the poisoning is only in the rocblas-test / rocblas-bench source code, and not the library's).

clients/include/utility.hpp
```diff
@@ -19,7 +19,18 @@
  * \brief provide common utilities
  */
 
-// Passed into gtest's SUCCEED macro when skipping a test.
+// We use std::cout and std::cerr instead of stdout, stderr, std::cout and std::cerr
+// This must come after the header #includes above, to avoid poisoning system headers.
```
amcamd: should this be:   We use rocblas_cout and rocblas_cerr instead of stdout, stderr, std::cout and std::cerr

clients/include/utility.hpp
```diff
@@ -131,26 +97,21 @@ inline void rocblas_print_matrix(
     for(size_t i = 0; i < m; i++)
         for(size_t j = 0; j < n; j++)
         {
-            printf("matrix  col %d, row %d, CPU result=%f, GPU result=%f\n",
-                   i,
-                   j,
-                   CPU_result[j + i * lda],
-                   GPU_result[j + i * lda]);
+            std::cout << "matrix  col " << i << ", row " << j
+                      << ", CPU result=" << CPU_result[j + i * lda]
```
amcamd: should this be rocblas_cout in place of std::cout ?

clients/include/utility.hpp
leekillough: That comment is, again, part of the mechanical changes. I basically undid the mechanical changes in the other PR to come up with this PR, so it replaced rocblas_cout and rocblas_cerr with std::cout and std::cerr, even in the comments. Refer to PR#922 for the real code if you have questions about this code and the differences look mechanical like they were edited by a sed script.

clients/include/utility.hpp
leekillough: Ignore all rocblas_cout/cerr and std::cout/cerr changes as they are mechanical.

PR #964, opened 2020-01-28
Removing large stride_scale for large hpr tests.

daineAMD:
I think the failing tests are from memory allocation failures. Reducing the size of stride_scale for large hpr tests, will wait to see if this passes CI.


daineAMD:
This passed all nightly hpr tests on Ubuntu, had various other failures though (sles, failure to connect to ixt-rack-40, hip-clang failures). I'll rerun tonight and likely merge tomorrow morning if everything looks good.

PR #966, opened 2020-01-29
Added llvm7 dependency for sles

amdkila:
- llvm7 is a mandatory library dependency for the new Tensile client. This PR adds it to the install script for sles and fedora

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -148,9 +148,9 @@ install_packages( )
                                       "zlib-devel" "wget" )
   local library_dependencies_fedora=( "make" "cmake" "rpm-build"
                                       "python34" "PyYAML" "python3*-PyYAML" "python3*-distutils-extra"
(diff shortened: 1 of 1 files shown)


amcamd:
The CI job is still failing with:

[ 70%] Building CXX object library/src/CMakeFiles/rocblas.dir/buildinfo.cpp.o
[ 71%] Linking CXX shared library librocblas.so
ld: cannot find -lLLVMObjectYAML
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [library/src/CMakeFiles/rocblas.dir/build.make:2882: library/src/librocblas.so.0.1] Error 1
make[1]: *** [CMakeFiles/Makefile2:210: library/src/CMakeFiles/rocblas.dir/all] Error 2

I guess that either it is not installing libLLVMObjectYAML.so, or it is not looking in the correct directory.

PR #967, opened 2020-01-29
Functional Her

daineAMD:
[Docs.](http://www.netlib.org/lapack/explore-html/d6/dee/group__complex__blas__level2_gaa59d93fbbd8d0b1be4a51634cb437cc1.html#gaa59d93fbbd8d0b1be4a51634cb437cc1)


clients/gtest/her_gtest.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
+            }
+            else
+            {
+                // For alpha: T doesn't really matter here, just whether it's real or complex, in her's case it's always complex.
+                name << '_' << (char)std::toupper(arg.uplo) << '_' << arg.N << '_' << arg.lda << '_'
+                     << arg.get_alpha<rocblas_double_complex>() << '_' << arg.incx;
```
TorreZuk: This is her so it's always real not complex right?   THis looks like it is her2 code where it is always complex.

clients/gtest/her_gtest.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+  - &alpha_range
+    - { alpha:  2.0, alphai:  1.5 }
+    - { alpha: -1.0, alphai:  0.5 }
+    - { alpha:  2.0, alphai: -1.5 }
+    - { alpha:  0.0, alphai: -0.5 }
+    - { alpha:  0.0, alphai:  0.0 }
+
+  - &alpha_range_small
+    - { alpha: 4.0, alphai: 2.0 }
```
TorreZuk: Should be real only.

clients/include/testing_her.hpp
```diff
@@ -0,0 +1,195 @@
...
+    rocblas_local_handle handle;
+
+    // argument check before allocating invalid memory
+    if(N < 0 || !incx || lda < 1 || lda < N)
+    {
+        EXPECT_ROCBLAS_STATUS(
+            (rocblas_her<T, U>)(handle, uplo, N, nullptr, nullptr, incx, nullptr, lda),
+            rocblas_status_invalid_size);
+
+        return;
```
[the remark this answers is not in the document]
daineAMD: As long as the allocation and everything works, I like leaving the N==0 case to go through everything so it checks the result as well.

PR #968, opened 2020-01-30
Update tensile tag to include atomicAdd fix

amdkila:
This applies changes to the develop branch that were added to the hip-clang-bkc-998 that allows rocBLAS to build all Tensile architecture sizes.


amdkila:
This PR doesn't solve the problem, so I'm closing it.

PR #975, opened 2020-02-03
Functional Syr2

daineAMD:
Implements syr2. Also adds complex alpha to testing for spr and syr.

Note:
cblas doesn't seem to implement csyr2 and zsyr2. Cublas does, however. In cblas_interface I implemented a local version of the complex versions of syr2 to check against.


clients/include/cblas_interface.hpp
```diff
@@ -1625,6 +1625,118 @@ void cblas_syr(
...
+            T tmpx = alpha * x[j * incx];
+            T tmpy = alpha * y[j * incy];
+            for(int i = j; i < n; ++i)
+            {
+                A[i + j * lda] = A[i + j * lda] + x[i * incx] * tmpy + y[i * incy] * tmpx;
+            }
+        }
+    }
+}
+
```
TorreZuk: It looks like blis has the symbols, you just have to hack in the extern "C" but what you did looks good too.
0000000000001ab0 T bli_csyr2
00000000000019f0 T bli_dsyr2
0000000000001930 T bli_ssyr2
0000000000001b80 T bli_zsyr2

clients/include/testing_syr2_batched.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+    EXPECT_ROCBLAS_STATUS(
+        rocblas_syr2_batched<T>(
+            handle, rocblas_fill_full, N, &alpha, dx, incx, dy, incy, dA_1, lda, batch_count),
+        rocblas_status_invalid_value);
+
+    EXPECT_ROCBLAS_STATUS(
+        rocblas_syr2_batched<T>(
+            handle, uplo, N, &alpha, nullptr, incx, dy, incy, dA_1, lda, batch_count),
+        rocblas_status_invalid_pointer);
```
TorreZuk: Does alpha get a nullptr test ?

clients/include/testing_syr2_batched.hpp
```diff
@@ -0,0 +1,255 @@
...
+        {
+            cblas_syr2<T>(uplo, N, h_alpha, hx[i], incx, hy[i], incy, hA_gold[i], lda);
+        }
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = batch_count * syr2_gflop_count<T>(N) / cpu_time_used * 1e6;
+
+        if(arg.unit_check)
+        {
+            unit_check_general<T>(1, size_A, batch_count, 1, hA_gold, hA_1);
+            unit_check_general<T>(1, size_A, batch_count, 1, hA_gold, hA_2);
```
TorreZuk: Don't these fail without near check for complex?

daineAMD: Hm, I changed this to unit_checks for all syr2 tests and they still all seem to pass. I'll fiddle around and see what happens with syr/spr/spr2/etc.

daineAMD: They all seem to be passing with unit_checks. I'm not sure which functions in particular were originally failing with unit_checks. I can change the near_checks in all these to unit_checks in another PR later and see how they do on CI.

TorreZuk: Sounds fine, it seems suspicious if they accumulate complex values I would have thought they would need near.

clients/include/testing_syr2.hpp
```diff
@@ -0,0 +1,216 @@
...
+        // CPU BLAS
+        cpu_time_used = get_time_us();
+        cblas_syr2<T>(uplo, N, h_alpha, hx, incx, hy, incy, hA_gold, lda);
+        cpu_time_used = get_time_us() - cpu_time_used;
+        cblas_gflops  = syr2_gflop_count<T>(N) / cpu_time_used * 1e6;
+
+        if(arg.unit_check)
+        {
+            unit_check_general<T>(N, N, lda, hA_gold, hA_1);
+            unit_check_general<T>(N, N, lda, hA_gold, hA_2);
```
leekillough: Use `unit_check_general<T, T>()` to resolve compilation errors. An extra template argument was added, to allow `float` and `rocblas_bfloat16` to be compared, allowing the `rocblas_bfloat16` to either equal the rounded or truncated value of the `float`. But this is only used in `gemm_ex`, and most functions should simply use two equal type arguments.

clients/include/testing_syr2_batched.hpp
leekillough: > near_check

Like `unit_check_general`, `near_check_general` now has two template arguments, which should be the same for the majority of functions.

clients/include/testing_syr2.hpp
TorreZuk: Lee I presume you didn't set a default template value U=T so something isn't overlooked?

leekillough: Default template arguments won't work. The reason for the explicit template arguments is to force the `host_vector` / `device_vector` arguments into pointers without having to use explicit casts or function calls.

By explicitly specifying the template arguments, which are used as `T*` and `U*` in the function parameter declarations, any `device_vector` or `host_vector` passed as an argument gets implicitly converted to a pointer to its data, while if even if we used default template arguments, such as
```
template<typename T, typename U = T>
void unit_check_general(T* x, U*x, ...)
{
}
```
Then a call of:

```
unit_check_general<T>(vec1<T>, vec2<U>);
```
would try to deduce `U*` from the  `vec2<U>` argument, and would fail. By specifying:
```
unit_check_general<T, U>(vec1<T>, vec2<U>);
```
we force the `T` and `U` to be explicitly specified and not deduced, and then the `T*` and `U*` parameters will force their `device_vector` or `host_vector` arguments to be converted to the pointers to their data.

leekillough: Note that a _failed deduction_ is not the same as a _non-deduced context_. The above uses `U*` in a _deduced context_, and that deduction fails if we pass an argument of type `device_vector` or `host_vector`.

If we wanted to make the second template argument optional, while keeping the functionality, we could do it this way:

```
template<typename T, typename U = T>
void unit_check_general(T* x, std::type_identity_t<U>* y)
{
}
```
`std::type_identity_t` is new to C++20, but it's trivial class we could implement ourselves:
```
template<typename T>
struct type_identity
{ 
   using type = T;
};
template<typename T>
using type_identity_t = typename type_identity<T>::type;
```
The reason this works is that `std::type_identity_t<U>` is a _non-deduced context_ for `U`, and since there are no _deduced contexts_ for `U` in the function parameter list, the default template argument definition of `U = T` can apply.

The way the code is written now, with `unit_check_general(T* x, U* y)`, both `T` and `U` are in deduced contexts, and hence if we were to specify a default `U = T`, it would be ignored.

I'm just rambling on in language-legaleze to show that I've thought a lot about it.

It's simpler now to just require both arguments to be explicitly specified. If this were to break users of a library who were used to passing a single type argument to `unit_check_general`, then we should go out of our way to implement the non-deduced context with `type_identity_t`, so that if `U` is not explicitly specified, it defaults to `T`.

TorreZuk: Figured there was a reason, good to know.   Thanks.

PR #983, opened 2020-02-05
merge develop into master for ROCm 3.2 Feature complete

amcamd:


mahmoodw:
FYI the team has discovered some issues with gfx908 kernels not being picked up on the new client. This is probably an issue since we are now defaulting to the new client? Some tests are being done to properly diagnose the issue. May want to hold off...

PR #984, opened 2020-02-05
Deduplicate template parameters

leekillough:
Some functions, such as complex, have an associated real type.

Rather than pass the real type as a separate template argument, derive the real type from the main type argument (real or complex).

Change `rocblas_real_t<T>` to `real_t<T>` for more concise name.

Remove unnecessary template wrappers in `testing_asum` and `testing_nrm2` which called templates with `<COMPLEX, REAL>` type pairs, relying only on the main type now.

In macro calls with used to have `<T1, T2>` or `<T, U>` template argument lists parenthesized inside of them, remove the redundant parentheses, now that only `<T>` is used.

Make `cblas_gemm` work with only one template argument.

Makes code easier to understand, and 300+ lines smaller.

In the future, use `real_t<T>` to get the real type of a real/complex/integer type, without having to pass extra template arguments.

Note: SCAL can multiply a complex vector by either a real or complex scalar, so it still has two independent template parameters.


leekillough:
I deduplicated the `near_check_general` and `unit_check_general` template parameters too, but it required adding `std::common_type_t< >` around the template parameters in the function declarations, to prevent them from being in _deduced contexts_. `std::identity_type_t` is the C++20 way, but for C++14 we can use `std::common_type_t` with a single argument to accomplish the same thing.

Now all the unnecessary duplicates are gone.

PR #998, opened 2020-02-13
restrict warnings to library

TorreZuk:
allow limited warnings to apply to library build not clients or anything outside


leekillough:
I had a different solution, using macros and not affecting CMake.

leekillough:
See this:

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/01185a82d73e9ddb35de54d49ef7f36c2fa50369/clients/include/rocblas_test.hpp#L23

PR #999, opened 2020-02-13
L3 syrk herk

TorreZuk:
Functional no work memory no tensile


library/src/blas3/rocblas_syrk_strided_batched.cpp
```diff
@@ -0,0 +1,232 @@
...
+                              transA,
+                              n,
+                              k,
+                              log_trace_scalar_value(alpha),
+                              A,
+                              lda,
+                              stride_a,
+                              log_trace_scalar_value(beta),
+                              C,
+                              ldc,
```
daineAMD: stride_c.

library/src/blas3/rocblas_herk_strided_batched.cpp
```diff
@@ -0,0 +1,226 @@
...
+                              n,
+                              k,
+                              log_trace_scalar_value(alpha),
+                              A,
+                              lda,
+                              stride_a,
+                              log_trace_scalar_value(beta),
+                              C,
+                              ldc,
+                              batch_count);
```
daineAMD: stride_c should be here too.

library/src/blas3/rocblas_herk.hpp
```diff
@@ -0,0 +1,260 @@
...
+        return rocblas_status_invalid_value;
+    if(transA != rocblas_operation_none && transA != rocblas_operation_conjugate_transpose)
+        return rocblas_status_invalid_value;
+
+    if(n < 0 || k < 0 || batch_count < 0 || ldc < n || (transA == rocblas_operation_none && lda < n)
+       || (transA != rocblas_operation_none && lda < k))
+        return rocblas_status_invalid_size;
+    if(!n || !batch_count)
+        return rocblas_status_success;
+    if(!AP || !CP || !alpha || !beta)
```
daineAMD: I recall for gemm coming across a jira ticket which brought up the issue of "if k == 0, do we care if A is a nullptr?" I'll forward you the email chain we had.
I think that for gemm, if k == 0, A is allowed to be a nullptr. I'm not sure if we should adopt that standard library-wide or if this can be ignored for now. We can probably talk about it at the meeting on Monday.

library/src/blas3/rocblas_syrk.hpp
```diff
@@ -0,0 +1,369 @@
...
+        return rocblas_status_invalid_value;
+    if(transA != rocblas_operation_none && transA != rocblas_operation_transpose)
+        return rocblas_status_invalid_value;
+
+    if(n < 0 || k < 0 || batch_count < 0 || ldc < n || (transA == rocblas_operation_none && lda < n)
+       || (transA != rocblas_operation_none && lda < k))
+        return rocblas_status_invalid_size;
+    if(!n || !batch_count)
+        return rocblas_status_success;
+    if(!AP || !CP || !alpha || !beta)
```
daineAMD: Same as comment in herk.

clients/gtest/herk_gtest.yaml
```diff
@@ -0,0 +1,117 @@
...
+include: known_bugs.yaml
+
+Definitions:
+  - &quick_matrix_size_range
+    - { N:  -1,  lda:  -1,  K:  1,  ldc:  1 } # bad n
+    - { N:   0,  lda:   2,  K: -1,  ldc: -1 } # bad k
+    - { N:   0,  lda:   3,  K:  3,  ldc:  3 } # n==0
+    - { N:   3,  lda:   3,  K:  0,  ldc:  3 } # k==0
+    - { N:   3,  lda:   1,  K:  3,  ldc:  3 } # bad lda
+    - { N:   3,  lda:   3,  K:  3,  ldc:  1 } # bad ldc
```
daineAMD: Might want to add tests where N > lda > K, or N < lda < K,  since these cases are ok (depending on transpose).

library/src/blas3/rocblas_herk.hpp
TorreZuk: That makes sense I can add if (k > 0 && (!AP || !alpha) || ...

clients/include/testing_herk.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+    EXPECT_ROCBLAS_STATUS(
+        (rocblas_herk<T, U>)(handle, uplo, transA, N, K, &alpha, dA, lda, &beta, nullptr, ldc),
+        rocblas_status_invalid_pointer);
+
+    // quick return with invalid pointers
+    EXPECT_ROCBLAS_STATUS(
+        (rocblas_herk<T,
+                      U>)(handle, uplo, transA, 0, K, nullptr, nullptr, lda, nullptr, nullptr, ldc),
+        rocblas_status_success);
```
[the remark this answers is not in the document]
TorreZuk: Subtle difference, but the nullptr can definitely be considered bad arg.    "special" arguments is the purpose here.    Could likely expand to cover all quick return cases, otherwise we repeat the same tests over and over if included in yaml sets.

clients/include/testing_herk.hpp
```diff
@@ -0,0 +1,275 @@
...
+        device_vector<T> dC(safe_size);
+        if(!dA || !dC)
+        {
+            CHECK_HIP_ERROR(hipErrorOutOfMemory);
+            return;
+        }
+
+        EXPECT_ROCBLAS_STATUS(
+            (rocblas_herk<T, U>)(handle, uplo, transA, N, K, &alpha, dA, lda, &beta, dC, ldc),
+            invalidSize ? rocblas_status_invalid_size : rocblas_status_success);
```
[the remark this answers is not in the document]
TorreZuk: Sure will add in some nulls to ensure the invalid check comes before pointer check.

library/src/blas3/rocblas_herk.hpp
leekillough: This will be examined when I do the sweep of the argument error-checking across all routines.

library/src/blas3/rocblas_syrk.hpp
leekillough: Use the Fortran philosophy: If an argument is not required for the evaluation of a function, then do not reference it or care whether it is valid.

library/src/blas3/rocblas_herk_batched.cpp
```diff
@@ -0,0 +1,200 @@
...
+namespace
+{
+    template <typename>
+    constexpr char rocblas_herk_name[] = "unknown";
+    template <>
+    constexpr char rocblas_herk_name<rocblas_float_complex>[] = "rocblas_cherk_batched";
+    template <>
+    constexpr char rocblas_herk_name<rocblas_double_complex>[] = "rocblas_zherk_batched";
+
+    template <typename T, typename U>
```
leekillough: Why two template parameters?  You should use one argument and `real_t<T>` for the real part.

We cannot support arbitrary mixed precision without a completely different implementation and tests, and so adding flexibility for the sake of flexibility is just confusing. Stick to one main type, and derive related types using traits like `real_t<T>`

clients/benchmarks/client.cpp
```diff
@@ -289,6 +297,10 @@ struct perf_blas<T, U, std::enable_if_t<std::is_same<T, float>{} || std::is_same
                 {"symv", testing_symv<T>},
                 {"symv_batched", testing_symv_batched<T>},
                 {"symv_strided_batched", testing_symv_strided_batched<T>},
+                // L3
+                {"syrk", testing_syrk<T>},
+                {"syrk_batched", testing_syrk_batched<T>},
+                {"syrk_strided_batched", testing_syrk_strided_batched<T>},
 #if BUILD_WITH_TENSILE
```
amcamd: do you need herk, herk_batched, herk_strided_batched here?

TorreZuk: herk primary template parameter only supports complex so it is only in the second case.  Without the context this does look suspicious.

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
 
+    herk performs one of the matrix-matrix operations for a Hermitian rank-k update
+
+    C := alpha*op( A )*op( A )^H + beta*C
+
+    where  alpha and beta are scalars, op(A) is an n by k matrix, and
+    C is a n x n Hermitian matrix stored as either upper or lower.
+
+        op( A ) = A   or   op( A ) = A^H
+
```
amcamd: can you move the information defining op(A) from the transA comment to here 
op( A ) = A, and A is  n by k  if transA  == rocblas_operation_none      
op( A ) = A^H and A is k by n  if transA  == rocblas_operation_conjugate_transpose

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
+    herk_batched performs a batch of the matrix-matrix operations for a Hermitian rank-k update
+
+    C_i := alpha*op( A_i )*op( A_i )^H + beta*C_i
+
+    where  alpha and beta are scalars, op(A) is an n by k matrix, and
+    C_i is a n x n Hermitian matrix stored as either upper or lower.
+
+        op( A_i ) = A_i   or   op( A_i ) = A_i^H
+
```
amcamd: same here as comment for rocblas_Xherk

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
+    herk_strided_batched performs a batch of the matrix-matrix operations for a Hermitian rank-k update
+
+    C_i := alpha*op( A_i )*op( A_i )^H + beta*C_i
+
+    where  alpha and beta are scalars, op(A) is an n by k matrix, and
+    C_i is a n x n Hermitian matrix stored as either upper or lower.
+
+        op( A_i ) = A_i   or   op( A_i ) = A_i^H
+
+
```
amcamd: same here as comment for rocblas_Xherk

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
+    syrk_batched performs a batch of the matrix-matrix operations for a symmetric rank-k update
+
+    C_i := alpha*op( A_i )*op( A_i )^T + beta*C_i
+
+    where  alpha and beta are scalars, op(A_i) is an n by k matrix, and
+    C_i is a symmetric n x n matrix stored as either upper or lower.
+
+        op( A_i ) = A_i   or   op( A_i ) = A_i^T
+
```
amcamd: same here as comment for rocblas_Xherk

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
+    syrk_strided_batched performs a batch of the matrix-matrix operations for a symmetric rank-k update
+
+    C_i := alpha*op( A_i )*op( A_i )^T + beta*C_i
+
+    where  alpha and beta are scalars, op(A_i) is an n by k matrix, and
+    C_i is a symmetric n x n matrix stored as either upper or lower.
+
+        op( A_i ) = A_i   or   op( A_i ) = A_i^T
+
```
amcamd: same here as comment for rocblas_Xherk

library/include/rocblas-functions.h
```diff
@@ -8838,6 +8838,664 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyr2_strided_batched(rocblas_handle
...
+    syrk performs one of the matrix-matrix operations for a symmetric rank-k update
+
+    C := alpha*op( A )*op( A )^T + beta*C
+
+    where  alpha and beta are scalars, op(A) is an n by k matrix, and
+    C is a symmetric n x n matrix stored as either upper or lower.
+
+        op( A ) = A   or   op( A ) = A^T
+
```
amcamd: same here as comment for rocblas_Xherk

clients/include/testing_herk.hpp
```diff
@@ -0,0 +1,275 @@
...
+        {
+            rocblas_herk<T, U>(handle, uplo, transA, N, K, h_alpha, dA, lda, h_beta, dC, ldc);
+        }
+        gpu_time_used  = get_time_us() - gpu_time_used;
+        rocblas_gflops = herk_gflop_count<T>(N, K) * number_hot_calls / gpu_time_used * 1e6;
+
+        std::cout << "uplo,transA,N,K,alpha,lda,beta,ldc,rocblas-Gflops,us";
+
+        if(arg.unit_check || arg.norm_check)
+            std::cout << ",CPU-Gflops,us,norm-error";
```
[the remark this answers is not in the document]
TorreZuk: You are correct the unit_check is forced off for bench so only norm_check is currently possible in timing if scope.  This knowledge is outside the scope so should have been enforced by a class framework, I'll fix all the if cases.

library/include/rocblas-functions.h
TorreZuk: Sure I'll put a description of dimensions in the intro.   Is it worth repeating either on transpose param detail or directly on the n and k param descriptions?    I just read trsm intro and it describes k usage without even saying what k is (k is defined in A param detail).

clients/include/testing_syrk_strided_batched.hpp
```diff
@@ -0,0 +1,429 @@
...
+                                            strideA,
+                                            h_beta,
+                                            dC,
+                                            ldc,
+                                            strideC,
+                                            batch_count);
+        }
+        gpu_time_used  = get_time_us() - gpu_time_used;
+        rocblas_gflops = syrk_gflop_count<T>(N, K) * number_hot_calls / gpu_time_used * 1e6;
+
```
amcamd: I think you need a   * batch_count    in the gflops calculation

clients/include/testing_herk.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+    const auto size_A = size_t(lda) * (transA == rocblas_operation_none ? K : N);
+    const auto size_C = size_t(ldc) * N;
+
+    // allocate memory on device
+    device_vector<T> dA(size_A);
+    device_vector<T> dC(size_C);
+    device_vector<U> d_alpha(1);
+    device_vector<U> d_beta(1);
+    if(!dA || !dC || !d_alpha || !d_beta)
```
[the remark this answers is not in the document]
TorreZuk: I would suggest we could just throw an exception from the device vector constructor and delete all the CHECK calls for memory, as knowing the line isn't too important as we'll debug it anyway by looking at the sizes.

leekillough:
Not appropriate. Flagged for review.

PR #1004, opened 2020-02-19
Revert "Avoid using Float16 type if using llvm clang++ (#1000)"

amdkila:
This reverts commit 24c189f941fad4692cf6a5ef8b299c2d90f232fd.

Looks like this commit breaks rocBLAS, so this should be removed.

### DIFF
--- a/library/include/rocblas-types.h
+++ b/library/include/rocblas-types.h
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright 2016-2020 Advanced Micro Devices, Inc.
+ * Copyright 2016-2019 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
 /*! \file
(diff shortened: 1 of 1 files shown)


amdkila:
[ 50%] Building CXX object library/src/CMakeFiles/hipblas.dir/hcc_detail/hipblas.cpp.o
In file included from /rocm_build/src/hipBLAS/library/src/hcc_detail/hipblas.cpp:6:
In file included from /rocm_build/install/include/rocblas.h:13:
In file included from /rocm_build/install/include/rocblas-auxiliary.h:8:
/rocm_build/install/include/rocblas-types.h:52:9: error: _Float16 is not supported on this target
typedef _Float16 rocblas_half;
        ^
This is the error that I'm getting. I'm using a clang++ 9.0 compiler, so I would expect _Float16 to exist. This change in rocBLAS allowed hipBLAS to build, but broke rocBLAS. Any idea how to fix the issue?

leekillough:
Something in hipBLAS is causing it to pick up the wrong compiler or something, different than rocBLAS.

I seem to recall there's a lot of CMake logic in rocBLAS for choosing the compiler (and libraries?) whether it's a hip-clang build or not, and setting different paths for things, etc. We should make sure similar logic is used in hipBLAS. There's no reason _Float16 should fail in hipBLAS but succeed in rocBLAS. They should ideally be using the same build environment (compilers + libraries).

PR #1006, opened 2020-02-19
SWDEV-208995

daineAMD:
[SWDEV-208995](http://ontrack-internal.amd.com/browse/SWDEV-208995).

Moves a test from nightly to pre-checkin, I'm guessing to check for lda/ldb edge case. @leekillough please let me know if this is what was intended from the ticket or not because it wasn't 100% clear to me.


leekillough:
I'm not sure of the original intent. Ask @bensander .

PR #1007, opened 2020-02-20
Support and test K==0 in GEMM

leekillough:
Resolves SWDEV-223351

Summary of proposed changes:

When `K == 0`, rocBLAS will set `K == 1` and `alpha == 0`, because Tensile does not support 0-sized dimensions right now.

Add `K == 0` tests, since they weren't being tested.

Change the computation of default `LDA` and `LDB` to use `1` if `K == 0`, because `0` is not a valid leading dimension.


leekillough:
Please review

PR #1009, opened 2020-02-20
add command line option of setting cold calls

wbgilmartin:
add parameter to the rocblas-bench to set the number of cold calls used for bench marking. 

this feature is enabled for the gemm tests.

### DIFF
--- a/clients/include/rocblas_common.yaml
+++ b/clients/include/rocblas_common.yaml
@@ -258,6 +258,7 @@ Arguments:
   - unit_check: rocblas_int
   - timing: rocblas_int
   - iters: rocblas_int
+  - colde_iters: rocblas_int
   - algo: c_uint
   - solution_index: c_int
   - flags: c_uint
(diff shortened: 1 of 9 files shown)


clients/include/rocblas_common.yaml
```diff
@@ -258,6 +258,7 @@ Arguments:
   - unit_check: rocblas_int
   - timing: rocblas_int
   - iters: rocblas_int
+  - colde_iters: rocblas_int
```
leekillough: misspelling

clients/include/testing_gemm_strided_batched.hpp
```diff
@@ -238,7 +238,7 @@ void testing_gemm_strided_batched(const Arguments& arg)
 
     if(arg.timing)
     {
-        int number_cold_calls = 2;
+        int number_cold_calls = arg.iters;
```
leekillough: should be `cold_iters`

leekillough: Shouldn't that be `arg.cold_iters`, or is it a special case?

wbgilmartin: fixed in last commit

PR #1010, opened 2020-02-21
functional syr2k and her2k

TorreZuk:
Summary of proposed changes:
- functional syr2k and her2k
- removed /efi partition formatting feature
- Daine I think your commit was temporarily added in a rebase that is why you are in the commit

### DIFF
--- a/clients/gtest/her2k_gtest.cpp
+++ b/clients/gtest/her2k_gtest.cpp
@@ -1,0 +1,163 @@
+/* ************************************************************************
+ * Copyright 2020 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "rocblas_data.hpp"
+#include "rocblas_datatype2string.hpp"
+#include "rocblas_test.hpp"
+#include "testing_her2k.hpp"
+#include "testing_her2k_batched.hpp"
+#include "testing_her2k_strided_batched.hpp"
+#include "type_dispatch.hpp"
+#include <cstring>
+#include <type_traits>
+
+namespace
+{
+    // possible gemv test cases
+    enum her2k_test_type
+    {
+        HER2K,
+        HER2K_BATCHED,
+        HER2K_STRIDED_BATCHED,
+    };
+
+    //ger test template
+    template <template <typename...> class FILTER, her2k_test_type HER2K_TYPE>
+    struct her2k_template : RocBLAS_Test<her2k_template<FILTER, HER2K_TYPE>, FILTER>
+    {
+        // Filter for which types apply to this suite
+        static bool type_filter(const Arguments& arg)
+        {
+            return rocblas_simple_dispatch<her2k_template::template type_filter_functor>(arg);
+        }
+
+        // Filter for which functions apply to this suite
+        static bool function_filter(const Arguments& arg)
+        {
+            switch(HER2K_TYPE)
+            {
+            case HER2K:
+                return !strcmp(arg.function, "her2k") || !strcmp(arg.function, "her2k_bad_arg");
+            case HER2K_BATCHED:
+                return !strcmp(arg.function, "her2k_batched")
+                       || !strcmp(arg.function, "her2k_batched_bad_arg");
+            case HER2K_STRIDED_BATCHED:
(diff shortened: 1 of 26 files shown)


clients/gtest/her2k_gtest.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    // possible gemv test cases
+    enum her2k_test_type
+    {
+        HER2K,
+        HER2K_BATCHED,
+        HER2K_STRIDED_BATCHED,
+    };
+
+    //ger test template
```
daineAMD: Update this comment and the previous gemv one :)

clients/gtest/syr2k_gtest.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "testing_syr2k.hpp"
+#include "testing_syr2k_batched.hpp"
+#include "testing_syr2k_strided_batched.hpp"
+#include "type_dispatch.hpp"
+#include <cstring>
+#include <type_traits>
+
+namespace
+{
+    // possible gemv test cases
```
daineAMD: Same here

clients/gtest/her2k_gtest.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+            name << rocblas_datatype2string(arg.a_type);
+
+            if(strstr(arg.function, "_bad_arg") != nullptr)
+            {
+                name << "_bad_arg";
+            }
+            else
+            {
+                // alpha & beta only real
```
daineAMD: Only beta is real in her2k right? Lines 87-89 can be removed, but it's insignificant.

library/src/blas3/rocblas_syr2k.hpp
```diff
@@ -0,0 +1,450 @@
...
+                                       TPtr              CP,
+                                       rocblas_int       offsetC,
+                                       rocblas_int       ldc,
+                                       rocblas_stride    strideC,
+                                       rocblas_int       batch_count)
+{
+    if(uplo != rocblas_fill_lower && uplo != rocblas_fill_upper)
+        return rocblas_status_invalid_value;
+
+    if(trans != rocblas_operation_none && trans != rocblas_operation_transpose)
```
daineAMD: See comment in rocblas-functions.h

library/include/rocblas-functions.h
```diff
@@ -9498,6 +9834,432 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyrk_strided_batched(rocblas_handle
...
+    handle    [rocblas_handle]
+              handle to the rocblas library context queue.
+
+    @param[in]
+    uplo    [rocblas_fill]
+            rocblas_fill_upper:  C is an upper triangular matrix
+            rocblas_fill_lower:  C is a  lower triangular matrix
+
+    @param[in]
+    trans  [rocblas_operation]
```
daineAMD: Oddly enough, in the ssyr2k documentation on netlib, they allow a conjugate-transpose operation, presumably this does the same as the simple transpose operation. For the complex cases (csyr2k, eg.) they do not support conjugate-transpose. Seems a little odd, not sure if we should support that as well.

library/src/blas3/rocblas_syr2k.hpp
```diff
@@ -0,0 +1,450 @@
...
+{
+    __shared__ T atile[TILE_NK][TILE_NK];
+    __shared__ T btile[TILE_NK][TILE_NK];
+
+    int col_pos = blockIdx.y * TILE_NK;
+    int row_pos = blockIdx.x * TILE_NK;
+
+    int tilefrom = upper ? row_pos : col_pos;
+    int tileto   = upper ? col_pos : row_pos;
+    if(!alpha || tilefrom > tileto)
```
daineAMD: Shouldn't need !alpha check here.

library/include/rocblas-functions.h
TorreZuk: Yes that is in the documentation but the code only does simple transpose and non-transpose.  As the other implementations don't mention it (e.g magmaTrans & magmaNoTrans) I think it is better to keep it clear this doesn't do conjugate transpose as Hermitian function her2k handles this.

PR #1011, opened 2020-02-22
Not running tests when more memory is needed than available.

daineAMD:
Potentially resolves [SWDEV-223558](http://ontrack-internal.amd.com/browse/SWDEV-223558) and [SWDEV-220984](http://ontrack-internal.amd.com/browse/SWDEV-220984).

Summary
------------
This allows tests to check how much memory will be allocated on the device, and compares this to how much memory is free. If the amount of memory to be allocated is greater, it immediately succeeds, and allows logging that a test was skipped.
This seems a little hacky, so any advice is appreciated.

Implementation
--------------------
In utility.cpp, I use `hipMemGetInfo` to query how much memory is available. In each test, I calculate how much memory is needed, and if it's  greater, I use googletest's SUCCESS with a unique message. In rocblas_gtest_main.cpp the listener detects this message and records the skipped test.

Example Output
---------
`[----------] Global test environment tear-down`
`[==========] 13697 tests from 2 test cases ran. (18047 ms total)`
`[  PASSED  ] 13697 tests.`
`[ SKIPPED  ] 592 tests.`

Alternatives
---------------
Alternatively, [GTEST_SKIP()](https://github.com/google/googletest/pull/1544) is a more elegant solution that I believe is available in a newer release of googletest (1.10.0) (we're currently using 1.8.0, not sure if it would be an issue to update).

Notes
-------
I only added the memory check for the functions noted in the two tickets. I also updated `ger_batched` and `syr_batched` to the new `device_batch_vector`s as well.
I haven't ran this on a V340L machine to see if it works, but it worked for my manual testing. Will do this and update.

### DIFF
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -58,16 +58,6 @@ double get_time_us_sync(hipStream_t stream)
 };
 
 /* ============================================================================================ */
-bool is_limited_memory(size_t needed_mem)
-{
-    size_t     free_mem  = 0;
-    size_t     total_mem = 0;
-    hipError_t err       = hipMemGetInfo(&free_mem, &total_mem);
-
-    return (err != hipSuccess || free_mem < needed_mem);
-}
-
-/* ============================================================================================ */
 /*  device query and print out their ID and name; return number of compute-capable devices. */
 rocblas_int query_device_property()
 {
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -58,6 +58,16 @@ double get_time_us_sync(hipStream_t stream)
 };
 
 /* ============================================================================================ */
+bool is_limited_memory(size_t needed_mem)
+{
+    size_t     free_mem  = 0;
+    size_t     total_mem = 0;
+    hipError_t err       = hipMemGetInfo(&free_mem, &total_mem);
+
+    return (err != hipSuccess || free_mem < needed_mem);
+}
+
+/* ============================================================================================ */
 /*  device query and print out their ID and name; return number of compute-capable devices. */
 rocblas_int query_device_property()
 {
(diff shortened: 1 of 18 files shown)


saadrahim:
This should be the standard for PRs for everyone, impressive.

saadrahim:
Please revert https://github.com/ROCmSoftwarePlatform/rocBLAS/commit/42f8f01497d7f742935a78bdb4ab65b5b4238780

to test on V340L

clients/common/utility.cpp
```diff
@@ -57,6 +57,16 @@ double get_time_us_sync(hipStream_t stream)
     return (tv.tv_sec * 1000 * 1000) + tv.tv_usec;
 };
 
+/* ============================================================================================ */
+bool is_limited_memory(size_t needed_mem)
+{
+    size_t     free_mem  = 0;
+    size_t     total_mem = 0;
+    hipError_t err       = hipMemGetInfo(&free_mem, &total_mem);
+
```
TorreZuk: What is the cost of this in terms of time?     Wondering if it is feasible to just move this into the device* allocators and return a failure that is handled by the CHECK_HIP_ERROR on device memory rather than all these special cases?

TorreZuk: You could also replace the CHECK_HIP_ERROR calls with a CHECK_HIP_DEV_ALLOC if hipMalloc failure is fine to trigger without checking memory free.

daineAMD:
I addressed Torre's comments on this and moved the checking code into the device*.hpp files instead of individual tests since the check takes only a few microseconds, and now it's generic when adding future tests.
I added another macro other than CHECK_HIP_ERROR to use to do this new check. In the future I think we maybe could instead just update CHECK_HIP_ERROR and create a new macro for CPU-side testing and whatnot.

PR #1012, opened 2020-02-24
update tensile init to only load valid modules

wbgilmartin:
change in tensile host initialization to only load modules for the current architecture.


amdkila:
Thanks for fixing this Bill. Please go ahead and merge this PR.

PR #1018, opened 2020-02-26
Gemm ex profile

mahmoodw:
Summary
--------------------
These changes add a gemm_ex_profile API which accepts events as additional arguments. These events are passed through to the tensile infrastructure that is already in place and used during kernel launch. (Old client and new client)

Implementation
--------------------
I added default event parameters to the following functions. (default to nullptr as per previous behaviour)
The following carry all parameters to the final tensile launch as always
- gemm_ex template
- gemm_ex_batched_template
- gemm_ex_typecasting
These are the final link between rocBLAS and Tensile
- call_tensile_ex (used for the old client)
- runContractionProblem (for the new client)

Originally the event parameters were defaulted to nullptrs by the macro and sent through in a similar manner.

I have added this API to a separate header file for a hidden approach. The idea is to separate this implementation from our public APIs and documentation to avoid the necessity of explicit testing and maintenance.

Alternatives
--------------------
The alternative discussed was to store events in the handle. The reason I lean back towards the current approach is because full control of the events should be given to the developer using the API. Someone using this feature for timing may want to use a unique event between iterations to identity timing concerns over concurrently launched kernels. With a single event, syncs would be required, in turn losing some flexibility.

Adding a vector of events and additional features to keep this flexibility seems to me like an over complication to the handle and may be counter intuitive for the user. (More documentation and user effort)

I feel this method keeps things simple and flexible. 

Comments
--------------------
I understand adding this sort of API may not be this simple and there may be other factors to consider. Please share your thoughts.


mahmoodw:
Original PR for reference:
https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/989

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -683,7 +691,14 @@ rocblas_status gemm_ex_batched_template(rocblas_handle    handle,
         handle, trans_a,  trans_b, m,    n,   k,        alpha, a,   lda,      stride_a,   b,
         ldb,    stride_b, beta,    c_in, ldi, stride_i, d,     ldd, stride_d, batch_count};
 
-    return handle->host->runContractionProblem(problem);
+    if(startEvent && stopEvent)
+    {
+        hipStream_t stream;
+        rocblas_get_stream(handle, &stream);
+        return handle->host->runContractionProblem(problem, &stream, startEvent, stopEvent);
```
leekillough: `runContractionProblem`  has access to the handle, so we do not need to pass `stream` as a separate argument, unless you want to support arbitrary streams independent of `handle`.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -66,13 +66,13 @@ static rocblas_status device_strided_batched_matrix_copy(const To*      src,
     To *dataD, const To *dataC, const Ti *dataA, const Ti *dataB, Tc alpha, Tc beta,              \
         size_t strideD1J, size_t strideD2K, size_t strideC1J, size_t strideC2K, size_t strideA1L, \
         size_t strideA2K, size_t strideB1J, size_t strideB2K, size_t sizeI, size_t sizeJ,         \
-        size_t sizeK, size_t sizeL, hipStream_t stream, size_t numInputEvents, void *dummy1,      \
-        void *dummy2
+        size_t sizeK, size_t sizeL, hipStream_t stream, size_t numInputEvents,                    \
+        hipEvent_t *startEvent, hipEvent_t *stopEvent
```
leekillough: This is only used in the old Tensile client. Do you intend to add this to both the old and new client?

library/src/blas_ex/rocblas_gemm_ex.hpp
leekillough: In `runContractionProblem`, simply use:
```
hipStream_t stream;
rocblas_get_stream(problem.handle, &stream);
```
and then you have `stream.

library/src/blas_ex/rocblas_gemm_ex.hpp
```diff
@@ -874,9 +899,10 @@ rocblas_status rocblas_gemm_ex_template(rocblas_handle    handle,
 
     rocblas_status rb_status = rocblas_status_not_implemented;
 
-#define EX_TYPECASTING_PARM                                                                   \
-    handle, trans_a, trans_b, m, n, k, alpha, a, offsetAin, lda, stride_a, b, offsetBin, ldb, \
-        stride_b, beta, c, offsetCin, ldc, stride_c, d, offsetDin, ldd, stride_d, batch_count
+#define EX_TYPECASTING_PARM                                                                    \
+    handle, trans_a, trans_b, m, n, k, alpha, a, offsetAin, lda, stride_a, b, offsetBin, ldb,  \
+        stride_b, beta, c, offsetCin, ldc, stride_c, d, offsetDin, ldd, stride_d, batch_count, \
+        startEvent, stopEvent
```
leekillough: Only used in old client.

library/src/tensile_host.cpp
```diff
@@ -411,7 +411,10 @@ TensileHost* createTensileHost()
  ******************************************************************************/
 template <typename Ti, typename To, typename Tc>
 rocblas_status
-    TensileHost::runContractionProblem(const RocblasContractionProblem<Ti, To, Tc>& problem)
+    TensileHost::runContractionProblem(const RocblasContractionProblem<Ti, To, Tc>& problem,
+                                       hipStream_t*                                 stream,
```
leekillough: See comment above about getting `stream` from `problem.handle`.

library/src/blas_ex/rocblas_gemm_ex.hpp
mahmoodw: good catch, I'll make the changes

library/src/blas_ex/rocblas_gemm_ex.hpp
mahmoodw: I already had the changes made for the old client... indifferent how we go about it.

leekillough: > I already had the changes made for the old client... indifferent how we go about it.

As long as it passes CI when built with old client (`install.sh -r`), I'm fine with it.

PR #1019, opened 2020-02-26
Fix skipping tests

daineAMD:
Fix for commit 5a3b4689233375618e453b2b92238a1081a1c0f4.

The previous commit would print that tests were skipped for any failed tests due to an incorrect use of strcmp. This should be fixed here.

### DIFF
--- a/clients/gtest/rocblas_gtest_main.cpp
+++ b/clients/gtest/rocblas_gtest_main.cpp
@@ -81,7 +81,7 @@ public:
 
     void OnTestPartResult(const TestPartResult& result) override
     {
-        if(strcmp(result.message(), LIMITED_MEMORY_STRING))
+        if(strcmp(result.message(), LIMITED_MEMORY_STRING_GTEST) == 0)
         {
             skipped_tests++;
             if(showInlineSkips)
--- a/clients/include/utility.hpp
+++ b/clients/include/utility.hpp
@@ -19,8 +19,15 @@
  * \brief provide common utilities
  */
 
+// Passed into gtest's SUCCEED macro when skipping a test.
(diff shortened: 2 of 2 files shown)


clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -81,7 +81,7 @@ class ConfigurableEventListener : public TestEventListener
 
     void OnTestPartResult(const TestPartResult& result) override
     {
-        if(strcmp(result.message(), LIMITED_MEMORY_STRING))
+        if(strcmp(result.message(), LIMITED_MEMORY_STRING_GTEST) == 0)
```
leekillough: My fault for not reporting this earlier. I saw it and thought it was strange, but failed to comment on it, due to personal issues and an urgent request to approve the PR.

daineAMD:
> Okay with the fix, but does the TODO have a path forward? Does the new gtest have a way to set the message.

I can't see any obvious fix for it if we continue to use a string  message to let the listener know the test was skipped.

Solutions for the TODO:
1. I haven't tried using gtest 1.10.0, but since it has a new macro (GTEST_SKIP()) I'm assuming that we wouldn't have to use the message portion and would be able to detect the skip using their framework instead of this hack.
2. Come up with a different way to pass the fact that we're skipping a test. I've only looked a little into googletest, but with 1.8.0 it doesn't seem like there's an easier/supported way to get the desired functionality here.

TorreZuk:
Sounds good.   If you have a jira ticket for this you could paste there, or make a rocblas project ticket.    I created a build system once that reported all TODOs after each build, would be easy to add but might be annoying.

PR #1020, opened 2020-02-26
revert to build with old tensile client

wbgilmartin:
Resets the install.sh to build with the new tensile client to false so by default we build with the old client.

The pre_checkin tests passed locally on a vega20 machine.


leekillough:
Why is this being reverted? Is there an unfixable bug holding up master?

wbgilmartin:
[AMD Official Use Only - Internal Distribution Only]


Yes. the performance.
________________________________

PR #1023, opened 2020-02-27
Updating strided_batched tests to use CHECK_DEVICE_ALLOCATION().

daineAMD:
- Updating strided_batched tests to use `CHECK_DEVICE_ALLOCATION()` over `CHECK_HIP_ERROR()` when allocating memory on device.
- Updated some L2 and L3 strided_batched tests to pass in nullptrs when quick returning or returning invalid_value
- Updated some tests to use arg.iters for rocblas-bench.

Haven't locally tested the nullptr tests so will definitely wait for CI to pass.

### DIFF
--- a/clients/include/testing_syr_strided_batched.hpp
+++ b/clients/include/testing_syr_strided_batched.hpp
@@ -36,11 +36,8 @@ void testing_syr_strided_batched_bad_arg()
     // allocate memory on device
     device_vector<T> dA_1(size_A);
     device_vector<T> dx(size_x);
-    if(!dA_1 || !dx)
-    {
-        CHECK_HIP_ERROR(hipErrorOutOfMemory);
-        return;
-    }
+    CHECK_DEVICE_ALLOCATION(dA_1.memcheck());
+    CHECK_DEVICE_ALLOCATION(dx.memcheck());
 
     EXPECT_ROCBLAS_STATUS(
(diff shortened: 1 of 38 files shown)


clients/include/testing_syr_strided_batched.hpp
```diff
@@ -75,21 +72,20 @@ void testing_syr_strided_batched(const Arguments& arg)
...
-        EXPECT_ROCBLAS_STATUS(
-            rocblas_syr_strided_batched<T>(
-                handle, uplo, N, &h_alpha, dx, incx, stridex, dA_1, lda, strideA, batch_count),
-            N < 0 || lda < N || lda < 1 || !incx || batch_count < 0 ? rocblas_status_invalid_size
-                                                                    : rocblas_status_success);
+        EXPECT_ROCBLAS_STATUS(rocblas_syr_strided_batched<T>(handle,
+                                                             uplo,
+                                                             N,
+                                                             &h_alpha,
+                                                             nullptr,
```
leekillough: Interesting, so you pass `nullptr` instead of unnecessarily allocating `safe_size` when passing other quick-return success or invalid arguments, since the pointers don't need to be dereferenced if there's a quick-return success or an invalid argument, and the pointers aren't checked until after the quick-return success or other invalid argument checks.

I did not originate the `safe_size` idea; it has been there for a while. But this is much simpler.

PR #1024, opened 2020-02-27
Add rocblas_init

leekillough:
Add `rocblas_init` function to initialize rocBLAS structures (such as new Tensile client), to avoid first-time overhead.

(Note: Some of the changes are untabification or removal of trailing spaces.)

### DIFF
--- a/clients/gtest/rocblas_gtest_main.cpp
+++ b/clients/gtest/rocblas_gtest_main.cpp
@@ -226,6 +226,9 @@ void catch_signals_and_exceptions_as_failures(const std::function<void()>& test)
  *****************/
 int main(int argc, char** argv)
 {
+    // Initialize rocBLAS (not explicitly needed; just included for testing)
+    rocblas_init();
+
     // Set signal handler
     rocblas_test_sigaction();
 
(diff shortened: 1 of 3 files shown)


leekillough:
> I thought the point of allowing cold calls count to zero was to capture all the initialization costs? Should this initialization be timed?

Braga says some clients may want to do the initialization explicitly, to avoid startup costs, such as if they're doing their own benchmarks.

leekillough:
> I thought the point of allowing cold calls count to zero was to capture all the initialization costs? Should this initialization be timed?

Braga says some clients may want to do the initialization explicitly, to avoid startup costs, such as if they're doing their own benchmarks.

leekillough:
> I thought the point of allowing cold calls count to zero was to capture all the initialization costs? Should this initialization be timed?

Braga says some clients may want to do the initialization explicitly, to avoid initialization costs, such as if they're doing their own benchmarks. This provides a way for them to explicitly do the initialization themselves.

leekillough:
Sorry, Github was telling me "You can't comment at this time," hence the repeated attempts.

PR #1030, opened 2020-02-28
Functional tpsv

daineAMD:
Basic implementation of tpsv using a simple substitution method. This method requires no additional memory and has good performance for smaller sizes, however as the size increase the performance degrades. This can be addressed in the future.

Performance:
I did some manual testing comparing this implementation of tpsv with the current implementation of trsv. Results are shown below for f32_r type:

| Size (m == n == lda) | % change |
----------------------------- | ---------------
| 32  | -71% |
| 128  | -75% |
| 512 | -7% |
| 1024 | +55% |
| 4096 | +400% |

Where % change is calculated as (tpsv time - trsv time) / (trsv time) * 100. This is averaged over Upper/Lower and Transpose/Non-Transpose cases.
As seen, performance is comparable up to a size of around 512. When we get to 4096, we see a 4x slowdown compared to trsv. This would get worse as the size grows.
Optimising larger sizes can be tackled in the future by 1. Using multiple blocks to do the algorithm or 2. Using a different algorithm (such as trsv's).

Thanks to @jsandham for helping with this.


clients/include/testing_tpsv.hpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        CHECK_ROCBLAS_ERROR(rocblas_set_pointer_mode(handle, rocblas_pointer_mode_host));
+        EXPECT_ROCBLAS_STATUS(
+            rocblas_tpsv<T>(handle, uplo, transA, diag, N, nullptr, nullptr, incx),
+            rocblas_status_invalid_size);
+        return;
+    }
+
+    size_t size_A   = size_t(N) * N;
+    size_t size_AP  = size_t(N) * (N + 1) / 2.0;
```
jsandham: Should this be a double 2.0?

jsandham: Or should it be int 2

clients/include/testing_tpsv.hpp
```diff
@@ -0,0 +1,266 @@
...
+        hA[i + i * N] = t;
+    }
+    //  calculate Cholesky factorization of SPD matrix hA
+    cblas_potrf<T>(char_uplo, N, hA, N);
+
+    //  make hA unit diagonal if diag == rocblas_diagonal_unit
+    if(char_diag == 'U' || char_diag == 'u')
+    {
+        if('L' == char_uplo || 'l' == char_uplo)
+            for(int i = 0; i < N; i++)
```
jsandham: I think you can use openmp for loops here if you want. Not sure if that is done in rcoBLAS or not.

library/src/blas2/rocblas_tpsv.hpp
```diff
@@ -0,0 +1,304 @@
...
+                // solve element that can be solved
+                if(tx == j && !diag)
+                {
+                    rocblas_int colA   = j + i;
+                    rocblas_int rowA   = j + i;
+                    rocblas_int indexA = packed_matrix_index(trans, trans, n, rowA, colA);
+                    xshared[tx]        = xshared[tx] / (CONJ ? conj(A[indexA]) : A[indexA]);
+                }
+
+                __syncthreads();
```
jsandham: Are these syncthreads safe? I think that all threads in a warp must be able to reach the syncthread, however if some threads here hit your break statement and some dont (within in the same warp) I dont know if this is safe or not. Maybe double check with torre if you dont know. See also: https://stackoverflow.com/questions/6666382/can-i-use-syncthreads-after-having-dropped-threads

library/src/blas2/rocblas_tpsv.hpp
```diff
@@ -0,0 +1,304 @@
...
+            // apply solved diagonal block to the rest of the array
+            // 1. Iterate down rows
+            for(rocblas_int j = BLK_SIZE + i; j < n; j += BLK_SIZE)
+            {
+                if(tx + j >= n)
+                    break;
+
+                // 2. Sum result (across columns) to be subtracted from original value
+                T val = 0;
+                for(rocblas_int p = 0; p < BLK_SIZE; p++)
```
jsandham: SInce BLK_SIZE is known at compile time I think this could be completely unrolled

jsandham: Maybe not important right now though depending on whether we stick with this algorithm in the future when we are optimizing

jsandham: Actually even better I think you can just insert a #pragma unroll above the for loop here (and anywhere else where the number of loop iterations is known at compile time) and it will unroll them for you instead of you doing it manually.

TorreZuk: The optimizer seems to do the unrolling of reasonable loops automatically from what Yvan has found.   One should experiment.

clients/include/testing_tpsv.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+    size_t size_A = N * size_t(N);
+    size_t size_x = N * size_t(incx);
+
+    device_vector<T> dA(size_A);
+    device_vector<T> dx(size_x);
+    CHECK_DEVICE_ALLOCATION(dA.memcheck());
+    CHECK_DEVICE_ALLOCATION(dx.memcheck());
+
+    //
```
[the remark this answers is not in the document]
daineAMD: This is done in the other function, `testing_tpsv` where we ensure we're getting a `rocblas_status_invalid_size` return for bad n or incx.

clients/include/testing_tpsv.hpp
daineAMD: Yep, fixed.

library/src/blas2/rocblas_tpsv.hpp
daineAMD: I remember seeing some sort of documentation on this somewhere, I'll try to find it again as a reference. If only the documentation was easily accessible...
It seems that what Torre said stands, I haven't tried anything more complicated than what's done here, not sure how `if`/`else` statements would be handled.

daineAMD: I couldn't find any documentation :( . For now I changed it to conform to the rules set by cuda. Can take another look if you want to make sure.

library/include/rocblas-functions.h
```diff
@@ -6507,47 +6507,288 @@ ROCBLAS_EXPORT rocblas_status rocblas_ztrsv_strided_batched(rocblas_handle
...
+    TPSV solves
 
-        y := alpha*A*x + beta*y,
+         A*x = b or A**T*x = b, or A**H*x = b,
 
-    where alpha and beta are scalars, x and y are n element vectors and
-    A is an n by n Hermitian(Symmetric) matrix.
+    where x and b are vectors and A is a triangular matrix stored in the packed format.
+
+    The vector x is overwritten on b.
```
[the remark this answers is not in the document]
daineAMD: Updated it, let me know what you think.

clients/include/testing_tpsv.hpp
daineAMD: Didn't see any other openmp loops in rocBLAS so left it for now, it's just testing code anyway.

library/src/blas2/rocblas_tpsv.hpp
daineAMD: No changes yet, I might take a look for this PR, but it can always be looked at later during the optimization stage.

clients/include/testing_tpsv.hpp
leekillough: We could parallelize CPU code with OpenMP to improve performance a little bit, which might become necessary at some point if the testing time gets too large.

library/include/rocblas-functions.h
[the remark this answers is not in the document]
TorreZuk: It is better

daineAMD:
Looks like 5fb7572 passed CI on the machines that were working. Merging.

clients/include/testing_tpsv.hpp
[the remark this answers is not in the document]
jsandham: Ok sounds good

library/src/blas2/rocblas_tpsv.hpp
jsandham: OK sounds good. As a side test it would be interesting to try the case where BLK_SIZE is the warp size (I know its 32 for cuda but maybe its 64 for hip?). In this case I dont think you need the syncthreads as long as you make the shared memory volatile. Be interesting to see if its faster.

PR #1031, opened 2020-02-29
Make profiling code type-safe and simpler

leekillough:
`hipEvent_t` can be pre-declared in `rocblas-types.h` so that it can be used in the declaration of `rocblas_gemm_ex_profile`'s event parameters. This avoids needing to use `void*` and casting them to `hipEvent_t*`.

`#include "rocblas-profile.h"` should be added to `rocblas.h` so that it's always included, rather than requiring the user to `#include <rocblas-profile.h>`.

The code
```
    if(startEvent && stopEvent)
        return handle->host->runContractionProblem(problem, startEvent, stopEvent);
    else
        return handle->host->runContractionProblem(problem);
```
... is redundant, since `runContractionProblem` already checks for `nullptr` `startEvent` and `stopEvent`, and is declared to always have `startEvent` and `stopEvent` parameters, so there is nothing gained by calling it two different ways. If the overhead of checking `startEvent` and `stopEvent` were significant, then two separate overloads of `runContractionProblem`, one with and one without `startEvent` and `stopEvent`, could be used, but that is not the case here.


library/include/rocblas.h
```diff
@@ -13,6 +13,7 @@
 #include "rocblas-auxiliary.h"
 #include "rocblas-export.h"
 #include "rocblas-functions.h"
+#include "rocblas-profile.h"
```
mahmoodw: The reason I wanted to avoid doing this is because I wanted to keep this API "hidden". Since it is not in the main rocblas-functions, I guess it's good enough?

leekillough: It should probably be public, and not "hidden", but I agree with separating the functions into their own headers.

leekillough: What do others think about keeping it "hidden", or making it a part of the headers included by `rocblas.h`? I favor making it a part of the functions declared when `rocblas.h` is `#include`'d, but I agree with putting it in a separate header than `rocblas-functions.h`.

leekillough:
The latest changes remove `rocblas-profile.h` and just add 4 undocumented functions to `rocblas_auxiliary.cpp` to manipulate events stored in the handle.

It's better to store the events in the handle, because then the events do not need to be passed around as extra parameters to functions, and they can be applied to any functions, not just `rocblas_gemm_ex`, without having to create extra functions with `_profile` suffixes.

Whereas before you would use:
```
rocblas_gemm_ex_profile(..., startEvent, stopEvent);
```
Now you would use:
```
rocblas_set_start_event(handle, startEvent);
rocblas_set_stop_event(handle, stopEvent);

// call roblas_gemm_ex normally after setting startEvent and stopEvent in the handle
```
Also note that `hipEvent_t` is already a pointer, so we do not need to use `hipEvent_t*` when passing events around, since that causes another level of indirection to be required. We only need `hipEvent_t*` when _returning_ an event through a pointer to the result, such as in `rocblas_get_start_event()` and `rocblas_get_stop_event()`.

mahmoodw:
> The latest changes remove `rocblas-profile.h` and just add 4 undocumented functions to `rocblas_auxiliary.cpp` to manipulate events stored in the handle.
> 
> It's better to store the events in the handle, because then the events do not need to be passed around as extra parameters to functions, and they can be applied to any functions, not just `rocblas_gemm_ex`, without having to create extra functions with `_profile` suffixes.
> 
> Whereas before you would use:
> 
> ```
> rocblas_gemm_ex_profile(..., startEvent, stopEvent);
> ```
> 
> Now you would use:
> 
> ```
> rocblas_set_start_event(handle, startEvent);
> rocblas_set_stop_event(handle, stopEvent);
> 
> // call roblas_gemm_ex normally after setting startEvent and stopEvent in the handle
> ```
> 
> Also note that `hipEvent_t` is already a pointer, so we do not need to use `hipEvent_t*` when passing events around, since that causes another level of indirection to be required. We only need `hipEvent_t*` when _returning_ an event through a pointer to the result, such as in `rocblas_get_start_event()` and `rocblas_get_stop_event()`.

Good changes and I agree with most of the changes. 

Choosing not to store the events in the handle is something I mentioned in the original PR. I am pasting my comments for discussion:

"The alternative discussed was to store events in the handle. The reason I lean back towards the current approach is because full control of the events should be given to the developer using the API. Someone using this feature for timing may want to use a unique event between iterations to identity timing concerns over concurrently launched kernels. With a single event, syncs would be required, in turn losing some flexibility.

Adding a vector of events and additional features to keep this flexibility seems to me like an over complication to the handle and may be counter intuitive for the user. (More documentation and user effort)

I feel this method keeps things simple and flexible."

Please share you thoughts about this. Additionally, I will need to test this implementation with GemmStress to verify it continues to work as expected after making some changes.

leekillough:
> 
> 
> > The latest changes remove `rocblas-profile.h` and just add 4 undocumented functions to `rocblas_auxiliary.cpp` to manipulate events stored in the handle.
> > It's better to store the events in the handle, because then the events do not need to be passed around as extra parameters to functions, and they can be applied to any functions, not just `rocblas_gemm_ex`, without having to create extra functions with `_profile` suffixes.
> > Whereas before you would use:
> > ```
> > rocblas_gemm_ex_profile(..., startEvent, stopEvent);
> > ```
> > 
> > 
> > Now you would use:
> > ```
> > rocblas_set_start_event(handle, startEvent);
> > rocblas_set_stop_event(handle, stopEvent);
> > 
> > // call roblas_gemm_ex normally after setting startEvent and stopEvent in the handle
> > ```
> > 
> > 
> > Also note that `hipEvent_t` is already a pointer, so we do not need to use `hipEvent_t*` when passing events around, since that causes another level of indirection to be required. We only need `hipEvent_t*` when _returning_ an event through a pointer to the result, such as in `rocblas_get_start_event()` and `rocblas_get_stop_event()`.
> 
> Good changes and I agree with most of the changes.
> 
> Choosing not to store the events in the handle is something I mentioned in the original PR. I am pasting my comments for discussion:
> 
> "The alternative discussed was to store events in the handle. The reason I lean back towards the current approach is because full control of the events should be given to the developer using the API. Someone using this feature for timing may want to use a unique event between iterations to identity timing concerns over concurrently launched kernels. With a single event, syncs would be required, in turn losing some flexibility.
> 
> Adding a vector of events and additional features to keep this flexibility seems to me like an over complication to the handle and may be counter intuitive for the user. (More documentation and user effort)
> 
> I feel this method keeps things simple and flexible."
> 
> Please share you thoughts about this. Additionally, I will need to test this implementation with GemmStress to verify it continues to work as expected after making some changes.

We could add a helper macro to set / clear the events at every call, like:
```
#define ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, call) \
    do                                                                        \
    {                                                                         \
        rocblas_handle h = (handle);                                          \
        rocblas_set_start_event(h, (startEvent));                             \
        rocblas_set_stop_event(h, (stopEvent));                               \
        call;                                                                 \
        rocblas_set_start_event(h, nullptr);                                  \
        rocblas_set_stop_event(h, nullptr);                                   \
    } while(0)
```

And then invoke it like:
```
rocblas_status status;
ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, status = rocblas_gemm_ex(handle, ...));
```

To keep the code smaller and to keep stuff "private", we can even combine the `rocblas_set_start_event()` and `rocblas_set_stop_event()` functions into one function which takes both events, like `rocblas_set_start_stop_events(handle, startEvent, stopEvent)` and we do not even need `rocblas_get_start_event()` and `rocblas_get_stop_event()` if we know that we will only be setting them through the macro above, and clearing them after a call. The macro above could be simplified to:
```
#define ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, call) \
    do                                                                        \
    {                                                                         \
        rocblas_handle h = (handle);                                          \
        rocblas_set_start_stop_events(h, (startEvent), (stopEvent));          \
        call;                                                                 \
        rocblas_set_start_stop_events(h, nullptr, nullptr);                   \
    } while(0)

mahmoodw:
> > > The latest changes remove `rocblas-profile.h` and just add 4 undocumented functions to `rocblas_auxiliary.cpp` to manipulate events stored in the handle.
> > > It's better to store the events in the handle, because then the events do not need to be passed around as extra parameters to functions, and they can be applied to any functions, not just `rocblas_gemm_ex`, without having to create extra functions with `_profile` suffixes.
> > > Whereas before you would use:
> > > ```
> > > rocblas_gemm_ex_profile(..., startEvent, stopEvent);
> > > ```
> > > 
> > > 
> > > Now you would use:
> > > ```
> > > rocblas_set_start_event(handle, startEvent);
> > > rocblas_set_stop_event(handle, stopEvent);
> > > 
> > > // call roblas_gemm_ex normally after setting startEvent and stopEvent in the handle
> > > ```
> > > 
> > > 
> > > Also note that `hipEvent_t` is already a pointer, so we do not need to use `hipEvent_t*` when passing events around, since that causes another level of indirection to be required. We only need `hipEvent_t*` when _returning_ an event through a pointer to the result, such as in `rocblas_get_start_event()` and `rocblas_get_stop_event()`.
> > 
> > 
> > Good changes and I agree with most of the changes.
> > Choosing not to store the events in the handle is something I mentioned in the original PR. I am pasting my comments for discussion:
> > "The alternative discussed was to store events in the handle. The reason I lean back towards the current approach is because full control of the events should be given to the developer using the API. Someone using this feature for timing may want to use a unique event between iterations to identity timing concerns over concurrently launched kernels. With a single event, syncs would be required, in turn losing some flexibility.
> > Adding a vector of events and additional features to keep this flexibility seems to me like an over complication to the handle and may be counter intuitive for the user. (More documentation and user effort)
> > I feel this method keeps things simple and flexible."
> > Please share you thoughts about this. Additionally, I will need to test this implementation with GemmStress to verify it continues to work as expected after making some changes.
> 
> We could add a helper macro to set / clear the events at every call, like:
> 
> ```
> #define ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, call) \
>     do                                                                        \
>     {                                                                         \
>         rocblas_handle h = (handle);                                          \
>         rocblas_set_start_event(h, (startEvent));                             \
>         rocblas_set_stop_event(h, (stopEvent));                               \
>         call;                                                                 \
>         rocblas_set_start_event(h, nullptr);                                  \
>         rocblas_set_stop_event(h, nullptr);                                   \
>     } while(0)
> ```
> 
> And then invoke it like:
> 
> ```
> rocblas_status status;
> ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, status = rocblas_gemm_ex(handle, ...));
> ```
> 
> To keep the code smaller and to keep stuff "private", we can even combine the `rocblas_set_start_event()` and `rocblas_set_stop_event()` functions into one function which takes both events, like `rocblas_set_start_stop_events(handle, startEvent, stopEvent)` and we do not even need `rocblas_get_start_event()` and `rocblas_get_stop_event()` if we know that we will only be setting them through the macro above, and clearing them after a call. The macro above could be simplified to:
> 
> ```
> #define ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, call) \
>     do                                                                        \
>     {                                                                         \
>         rocblas_handle h = (handle);                                          \
>         rocblas_set_start_stop_events(h, (startEvent), (stopEvent));          \
>         call;                                                                 \
>         rocblas_set_start_stop_events(h, nullptr, nullptr);                   \
>     } while(0)
> ```

This sounds like a good idea and handles my concerns.

leekillough:
> > ```
> > #define ROCBLAS_INVOKE_START_STOP_EVENTS(handle, startEvent, stopEvent, call) \
> >     do                                                                        \
> >     {                                                                         \
> >         rocblas_handle h = (handle);                                          \
> >         rocblas_set_start_stop_events(h, (startEvent), (stopEvent));          \
> >         call;                                                                 \
> >         rocblas_set_start_stop_events(h, nullptr, nullptr);                   \
> >     } while(0)
> > ```
> 
> This sounds like a good idea and handles my concerns.

I'm welcome to any better-sounding names for the macro.

mahmoodw:
Updating and testing GemmStress with these changes

leekillough:
Running full tests offline

mahmoodw:
Tested w/ GemmStress using both old and new client. LGTM

PR #1033, opened 2020-03-01
Removing rocm-dev from dependency installation

pruthvistony:
ROCm is considered already installed on the machine, so removing it from dependency list.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,6 +29,7 @@ rocBLAS build & installation helper script
       -r | --no-tensile-host   Do not build with Tensile host
       -u | --use-tag-only      Ignore Tensile version and just use the Tensile tag
            --skipldconf        Skip ld.so.conf entry
+           --skip_rocm_install Skip rocm component install
 EOF
 #          --prefix            Specify an alternate CMAKE_INSTALL_PREFIX for cmake
 #          --cuda              Build library for cuda backend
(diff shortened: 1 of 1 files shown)


saadrahim:
@amdkila is going to take a look at this. rocBLAS team wants to continue to keep rocm-dev as a dependency.

leekillough:
As we discussed, we may want to add an `install.sh` option to explicitly specify the dependency, which could include explicit version information, or be left generic to pick up the latest.

leekillough:
It might be as simple as a command-line option like `-v | --rocm-dev  Set rocm-dev package dependency (default rocm-dev)`

pruthvistony:
Flag “–rocm_installed” to install.sh, which will skip ROCm components from installation.
Flag is false by default.

pruthvistony:
@leekillough 
I have added the an option to skip the rocm-dev installation (--skip_rocm_install) default this var is false.

leekillough:
> 
> 
> @amdkila is going to take a look at this. rocBLAS team wants to continue to keep rocm-dev as a dependency.

Any update? We need to resolve this soon.

amdkila:
> > @amdkila is going to take a look at this. rocBLAS team wants to continue to keep rocm-dev as a dependency.
> 
> Any update? We need to resolve this soon.

See PR #1040

TorreZuk:
So need to get agreement that #1040 into develop can supplant this.  This is into master.

leekillough:
Will cherry-pick #1040 into master

PR #1036, opened 2020-03-02
Updating memchecks in tests.

daineAMD:
Main purpose is to update tests to use `CHECK_DEVICE_ALLOCATION` rather than `CHECK_HIP_ERROR` where appropriate.

Also updated some non-batched L2 tests to use nullptrs instead of the `safe_size` pattern when expecting a quick-return or invalid-value return.


daineAMD:
This didn't fix all of Alex's errors (another fix is on the way), but these are still good fixes to have. CI passed at 8223f25 on working nodes so I'll merge soon.

leekillough:
Memory leaks can occur if `SIGSEGV` or other signals are received and the program is allowed to continue; there's no easy solution for this memory leak (I tried a variety of solutions), but we should be able to adapt to the leaked memory, and continue to run tests which still have enough memory. The tests just need to be able to adapt to decreasing memory after each test.

PR #1038, opened 2020-03-03
Change CMake target hip::hip_hcc to hip::device

aaronenyeshi:
hip::device is supported by both hip_hcc and hip_vdi, so switching to this new cmake target should fix issues on HIP-Clang CI when building tests.
https://github.com/ROCm-Developer-Tools/HIP/blob/master/CMakeLists.txt#L351

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -128,8 +128,6 @@ else()
     endif()
 endif()
 
-get_target_property( HIPHCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-
 if( CUDA_FOUND )
   target_include_directories( rocblas-bench
     PRIVATE
@@ -141,7 +139,7 @@ if( CUDA_FOUND )
 else( )
   # auto set in hip_common.h
   #target_compile_definitions( rocblas-bench PRIVATE __HIP_PLATFORM_HCC__ )
-  target_link_libraries( rocblas-bench PRIVATE ${HIPHCC_LOCATION} )
+  target_link_libraries( rocblas-bench PRIVATE hip::device )
 endif( )
 
 if( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -198,8 +198,6 @@ else()
     endif()
 endif()
 
-get_target_property( HIPHCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-
 if( CUDA_FOUND )
   target_include_directories( rocblas-test
     PRIVATE
@@ -211,7 +209,7 @@ if( CUDA_FOUND )
 else( )
   # auto set in hip_common.h
   #target_compile_definitions( rocblas-test PRIVATE __HIP_PLATFORM_HCC__ )
-  target_link_libraries( rocblas-test PRIVATE ${HIPHCC_LOCATION})
+  target_link_libraries( rocblas-test PRIVATE hip::device )
 endif( )
 
 set( ROCBLAS_TEST_DATA "${PROJECT_BINARY_DIR}/staging/rocblas_gtest.data")
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -2,8 +2,6 @@
 # Copyright 2016-2020 Advanced Micro Devices, Inc.
 # ########################################################################
 
-get_target_property( HIPHCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-
 set( rocblas_samples_common ../common/utility.cpp )
 
 set( THREADS_PREFER_PTHREAD_FLAG ON )
@@ -54,7 +52,7 @@ foreach( exe ${sample_list} )
 
     # auto set in hip_common.h
     #target_compile_definitions( ${exe} PRIVATE __HIP_PLATFORM_HCC__ )
-    target_link_libraries( ${exe} PRIVATE ${HIPHCC_LOCATION} )
+    target_link_libraries( ${exe} PRIVATE hip::device )
   endif( )
 
   if( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )


aaronenyeshi:
HIP-Clang CI is failing on this issue, and needs to be unblocked.

saadrahim:
If CI passes, we should accept this PR.

amdkila:
> There are two other instances of hip_hcc being linked in the library/src/CMakeLists file. I'll remove those as well

Never mind, hip-clang purposely avoids those two instances due to if statement condition. Looks good to me, I'll approve it- but yeah, should let it run on CI first.

aaronenyeshi:
It looks like failures are related to docker config?
```
+ docker inspect -f . rocblas/rocm-docker-3.1/ubuntu16gfx900rocm31/0:latest

.

Failed to run image 'rocblas/rocm-docker-3.1/ubuntu16gfx900rocm31/0:latest'. Error: docker: Error response from daemon: open /data/lib/docker/overlay2/c8ddedbe6355c156d994357b636f9fac2edc4644a49794bf218605a76d542ed6/link: no such file or directory.

See 'docker run --help'.
```

amdkila:
> It looks like failures are related to docker config?
> 
> ```
> + docker inspect -f . rocblas/rocm-docker-3.1/ubuntu16gfx900rocm31/0:latest
> 
> .
> 
> Failed to run image 'rocblas/rocm-docker-3.1/ubuntu16gfx900rocm31/0:latest'. Error: docker: Error response from daemon: open /data/lib/docker/overlay2/c8ddedbe6355c156d994357b636f9fac2edc4644a49794bf218605a76d542ed6/link: no such file or directory.
> 
> See 'docker run --help'.
> ```

yep you're right. I set that machine offline, but we have some load restrictions with Vega 10s, so haven't been able to start a new job yet. I think if this gets through CI on any build, we can merge this.

amdkila:
There's a CMake bug that needs to be fixed...looks like HIPHCC_LOCATION needs to be removed? Anyway, I'll take a look a this ASAP

leekillough:
Lots of build errors:
```
*******************************************************************************

*----------------------------------- ERROR -----------------------------------*

* The variable 'CMAKE_CXX_FLAGS' should only be set by the cmake toolchain,

* either by calling 'cmake -DCMAKE_CXX_FLAGS="-isystem /opt/rocm/include -isystem /opt/rocm/hcc/lib/clang/10.0.0/include -isystem /opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/include -isystem /opt/rocm/include  -hc -fopenmp=libgomp -pthread -fopenmp=libgomp -pthread"' or

* set in a toolchain file and added with 

* 'cmake -DCMAKE_TOOLCHAIN_FILE=<toolchain-file>'.

*-----------------------------------------------------------------------------*

*******************************************************************************


CMake Warning at /opt/rocm/share/rocm/cmake/ROCMChecks.cmake:27 (message):

  The toolchain variable 'CMAKE_CXX_FLAGS' is modified in the CMakeLists.txt.

Call Stack (most recent call first):

  clients/gtest/CMakeLists.txt:9999 (rocm_check_toolchain_var)

  clients/gtest/CMakeLists.txt:243 (set)



rocblas (>=2.17.0.1993-PR-1038-6-53434ad)

CMake Error: The following variables are used in this project, but they are set to NOTFOUND.

Please set them or make sure they are set and tested correctly in the CMake files:

HIPHCC_LOCATION

    linked by target "example-sscal" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/samples

    linked by target "example-scal-template" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/samples

    linked by target "example-sgemm-strided-batched" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/samples

    linked by target "example-sgemm" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/samples

    linked by target "rocblas-bench" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/benchmarks

    linked by target "rocblas-test" in directory /var/jenkins_home/workspace/ckin-rocm-docker_rocBLAS_PR-1038@2/GBSHDYEw8/rocblas/clients/gtest
```

aaronenyeshi:
Is it possible to get this HIPHCC_LOCATION in a different way?

aaronenyeshi:
I've updated the change, I believe you can target_link_libraries directly with the hip::device INTERFACE rather than using get_target_property and IMPORTED_LOCATION_RELEASE.

clients/benchmarks/CMakeLists.txt
```diff
@@ -141,7 +139,7 @@ if( CUDA_FOUND )
 else( )
   # auto set in hip_common.h
   #target_compile_definitions( rocblas-bench PRIVATE __HIP_PLATFORM_HCC__ )
-  target_link_libraries( rocblas-bench PRIVATE ${HIPHCC_LOCATION} )
+  target_link_libraries( rocblas-bench PRIVATE hip::device )
```
pfultz2: Does the clients have kernels in them? You should only link with `hip::device` when you have gpu kernels. 

Also, using `hip::host` shouldn't be needed either, as that should be provided by `roc::rocblas`(and if its not, then that should be considered a bug).

leekillough: The clients depend on rocblas, and rocblas may eventually become a static library.

aaronenyeshi: @amdkila do you know if rocblas-bench, rocblas-test or any samples have GPU kernels?

amdkila: rocblas-bench is a tool for GPU kernel benchmarking, so definitely yes. rocblas-test does CPU and GPU comparisons for accuracy, so I would think both hip::host and hip::device should be used. I do not know if hip::host is provided in roc::rocblas...but if it is, then @pfultz2 is right that there's no need to include hip::host. I do not know anything about rocBLAS samples, so I don't want to comment on whether GPU kernels are needed

pfultz2: > The clients depend on rocblas, and rocblas may eventually become a static library.
 
Yes, even more so why rocblas should make sure its usage requirements are correct. The `hip::host` should be linked as `PUBLIC` or `INTERFACE` to rocblas, so that when you link against rocblas in the client, cmake will automatically link against `hip::host` for you as well.

Looking here:

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/CMakeLists.txt#L315

It appears rocblas does not link with `hip::host` correctly. It should add this line to the same cmake:

```cmake
target_link_libraries( rocblas PUBLIC hip::host )
```

Then the clients wont need to explicitly link against hip unless they are compiling device kernels. I assume this is not the case for clients since they are supposed to be built with vanilla gcc.

saadrahim:
It looks like @pfultz2 has some suggestions. Not sure if it is best address them in the future or do it now. I don't want to hold up @aaronenyeshi but we shouldn't miss the chance to improve this behaviour.

aaronenyeshi:
sure, let's submit this first, and we can address the usage of hip::host in rocblas in a future PR?

PR #1044, opened 2020-03-04
Enable support on CentOS-6

parallelo:
The end goal for these CentOS6-related mods is to support a manylinux2010 build environment on ROCm.

On CentOS-6, the llvm7.0-* packages do not exist.  In order to install a modern version of LLVM on CentOS-6, it must be built from source.  Hence, this PR's purpose is to remove the llvm7.0-* packages from the dependency list for versions of CentOS that are less than 7.  

Additionally, CentOS-6 doesn't include a '/etc/os-release' file, so we grab the details from the '/etc/centos-release' file instead (if it exists).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -173,7 +173,7 @@ install_packages( )
   fi
 
   if [[ "${ID}" == "centos" ]]; then
-    if [[ "${VERSION_ID}" >= "7" ]]; then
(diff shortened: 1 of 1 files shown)


TorreZuk:
Akila I thought you were going to reverse the logic of the fall through.  Can discuss tomorrow

PR #1049, opened 2020-03-05
Add missing __device__ label

amdkila:


leekillough:
> 
> 
> My fault for missing this, thanks Akila.

How did it even compile, though? :) Was there a `__host__` fallback function?

amdkila:
The failure is infrastructure-related (laggy file system). Another job should fail as rack 88 lost connection. If any 1 of these passes the compile stage, I'll merge this in.

aaronenyeshi:
> > My fault for missing this, thanks Akila.
> 
> How did it even compile, though? :) Was there a `__host__` fallback function?

HIP-Clang is more strict, whereas HCC will allow `__device__` to call `__host__` functions, since it sets all device callees as device too.

leekillough:
> 
> 
> > > My fault for missing this, thanks Akila.
> > 
> > 
> > How did it even compile, though? :) Was there a `__host__` fallback function?
> 
> HIP-Clang is more strict, whereas HCC will allow `__device__` to call `__host__` functions, since it sets all device callees as device too.

Ah, got it.

PR #1050, opened 2020-03-05
Functional tbsv

daineAMD:
Implementing functional tbsv (SWDEV-194776).
Uses the same logic as used for tpsv just with different indexing and slightly different bounds. Will have similar performance - good for smaller matrices (n <= 1024), poor for larger matrices.

I ran the tests on my local machine (Vega 20) with the following times:
|  | Quick | Pre-Checkin | Nightly | Total |
| ---- | -------- | ---------------- | --------- | ------  |
| tbsv                             | 2.3s | 3.6s | 25.6s | 31.5s |
| tbsv_batched              | 6.4s | 9.0s | 76.6s | 92.0s |
| tbsv_strided_batched | 10.7s | 8.8s | 76.6s | 96.1s |
| Sum                            | 19.4s | 21.4s | 178.8s | 219.6s |

These are slightly below the targeted BLAS2 times in addtest.rst.


library/src/blas2/rocblas_tbsv.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+            __syncthreads();
+        }
+    }
+
+    // Uses backward substitution to solve Ax = b. Used for a non-transposed upper-triangular matrix
+    // or a transposed lower-triangular matrix.
+    template <bool CONJ, rocblas_int BLK_SIZE, typename T>
+    __device__ void tbsv_backward_substitution_calc(
+        bool diag, bool trans, int n, int k, const T* A, rocblas_int lda, T* x, rocblas_int incx)
```
TorreZuk: same as earlier function trans as a template parameter may help with loop unrolling by reducing the amount of code

library/src/blas2/rocblas_tbsv_strided_batched.cpp
```diff
@@ -0,0 +1,254 @@
...
+                            "stride_x",
+                            stride_x,
+                            "batch_count",
+                            batch_count);
+        }
+
+        if(uplo != rocblas_fill_lower && uplo != rocblas_fill_upper)
+            return rocblas_status_invalid_value;
+        if(batch_count < 0 || ((n < 0 || k < 0 || lda < k + 1 || !incx) && batch_count > 0))
+            return rocblas_status_invalid_size;
```
TorreZuk: Not sure why batch_count=0 can overrule invalid args but n=0 does not.   If this was specified in our design doc, I may have forgotten.   If we choose or have chosen to go this way the !batch_count return status_sucess could just be moved ahead of invalid_size checks for clarity.

daineAMD: Don't see it anywhere in the design doc or anything.
I remember having the discussion awhile back, the thinking for this was that if batch_count == 0, it shouldn't matter what any of the other parameters are as there's no batches. This is different from n == 0 since there _is_ a valid batch, just that one of the parameters makes it so no calculations are necessary.
This currently matches trsv, I'll leave it for now. Do you know if there's a ticket for the "argument error-checking" task we've discussed in the past? I can make one or add this discussion to it.

clients/include/testing_tbsv.hpp
```diff
@@ -0,0 +1,226 @@
...
+    rocblas_fill      uplo   = char2rocblas_fill(char_uplo);
+    rocblas_operation transA = char2rocblas_operation(char_transA);
+    rocblas_diagonal  diag   = char2rocblas_diagonal(char_diag);
+
+    rocblas_status       status;
+    rocblas_local_handle handle;
+
+    // check here to prevent undefined memory allocation error
+    if(N < 0 || K < 0 || lda < K + 1 || !incx)
+    {
```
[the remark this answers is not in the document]
daineAMD: Yeah good catch, added N == 0 quick return tests.

library/src/blas2/rocblas_tbsv_strided_batched.cpp
TorreZuk: Lee may have an open ticket but there is enough work to make a new one I think.  I have been coding up batch_count == 0 as an equivalent to other quick returns.  This happens after invalid argument checks, so there is no difference to n==0, simpler in my opinion there is not another rule to consider.   magma dsyr2k_batched implementation does it this way,  should compare against cublas.

PR #1055, opened 2020-03-10
Fix known_bug_platforms

leekillough:
`known_bug_platforms` doesn't work 100% right now, and the way it behaves may be confusing because of how it handles `category`.

New method causes `category` to be set to the original category or to `known_bug` when forming tests (previously it set `category` to `known_bug_platforms_<original_category>` and left it that way no matter whether it matched the current platform).

Add GFX908 known bugs to the `known_bugs.yaml` file, so that they are only considered known bugs on GFX908.

GFX908 testing is going on right now; `known_bugs.yaml` might slightly change depending on results.


clients/include/rocblas_arguments.hpp
```diff
@@ -78,9 +78,9 @@ struct Arguments
     int32_t  solution_index;
     uint32_t flags;
 
-    char function[64];
-    char name[64];
-    char category[64];
+    char         function[64];
+    char         name[64];
+    mutable char category[64];
```
TorreZuk: Could comment here why this is mutable.

clients/common/utility.cpp
```diff
@@ -156,6 +158,8 @@ bool match_test_category(const char* category,
                 break;
             }
         }
+
+        strcpy(arg.category, test_category);
```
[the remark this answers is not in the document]
leekillough: Good catch, will change it to:
```
memmove(arg.category, test_category, strlen(test_category) + 1);
```
`test_category` is either `"known_bug"` or the original `arg.category` with the `known_bug_platforms_` prefix removed, so it will never exceed the size of `arg.category` (the `rocblas_gentest.py` script will error if a string larger than `category` is specified in YAML). Most of the time the suffix is shorter than the `known_bug_platforms_` prefix so they won't overlap, but in case the suffix is longer than the `known_bug_platforms_` prefix, `memmove` is necessary.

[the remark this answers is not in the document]
leekillough: There is no longer need to do the `memmove`.

Now, the `category` is either left unchanged, or it is set to `known_bug`.

clients/include/rocblas_arguments.hpp
leekillough: It is no longer need to be `mutable`. Instead, in the one place where `Arguments` is `const` but we need to modify `category`, we use `const_cast`.

clients/common/rocblas_gentest.py
```diff
@@ -233,7 +235,7 @@ def setdefaults(test):
                               'ger_strided_batched', 'geru_strided_batched',
                               'gerc_strided_batched', 'trsv_strided_batched'):
         if test['function'] in ('ger_strided_batched', 'geru_strided_batched',
-                                'gerc_strided_batched','trsv_strided_batched'
+                                'gerc_strided_batched', 'trsv_strided_batched'
```
leekillough: This and the changes above are to satisfy `flake8`.

leekillough:
The new method passes the `category` unchanged, without the `known_bug_platforms_` prefix to `rocblas-test`. If `known_bug_platforms` is set, then `rocblas-test` changes the `category` to `known_bug` if the platform matches the current GPU platform (using `const_cast` instead of `mutable` to get around the otherwise-`const` `Arguments`). There is no need to use `memmove` to move overlapping strings, and the code is simpler.

So far all 11 GFX908 failures have been detected and listed in `known_bugs.yaml` with `known_bug_platforms: gfx908`. GFX906 has no failures with either the old or new Tensile client. GFX900 is being tested right now with both the old and new Tensile clients, and any failures will be added to `known_bugs.yaml`, with `known_bug_platforms` set to `gfx900`.

PR #1056, opened 2020-03-10
Template api

TorreZuk:
* exports solver req. templates except L1 functions with internalized checks and allocations


library/src/template-proto.awk
```diff
@@ -0,0 +1,14 @@
+# multi-line
```
leekillough: `s/awk/perl/`

TorreZuk: s/awk/gawk but it looks like will have to s/gawk/python

leekillough: Use whatever tool is best for the job. TIMTOWDI!!!

PR #1058, opened 2020-03-11
Remove pre-check for memory allocation.

daineAMD:
- Removes the use of `hipMemGetInfo()` introduced in #1011 to query available device memory before allocation. This has been proven to be an unreliable way of getting available memory on the device. We shouldn't need this query since we're skipping tests on failed memory allocation as well.

### DIFF
--- a/clients/include/device_batch_vector.hpp
+++ b/clients/include/device_batch_vector.hpp
@@ -208,9 +208,7 @@ public:
     //!
     hipError_t memcheck() const
     {
-        if(!m_enough_memory)
-            return hipErrorMemoryAllocation;
-        else if((bool)*this)
+        if((bool)*this)
             return hipSuccess;
         else
             return hipErrorOutOfMemory;
@@ -222,7 +220,6 @@ private:
     rocblas_int m_batch_count{};
     T**         m_data{};
     T**         m_device_data{};
-    bool        m_enough_memory{};
 
     //!
     //! @brief Try to allocate the ressources.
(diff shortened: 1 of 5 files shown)


clients/include/device_batch_vector.hpp
```diff
@@ -208,9 +208,7 @@ class device_batch_vector : private d_vector<T, PAD, U>
     //!
     hipError_t memcheck() const
     {
-        if(!m_enough_memory)
-            return hipErrorMemoryAllocation;
-        else if((bool)*this)
+        if((bool)*this)
```
leekillough: `(bool)` cast is not necessary, since when used in `if(...)` or other boolean contexts, the argument is converted to `bool`, even if only an `explicit` conversion to `bool` exists. `if(*this)` is sufficient.

clients/include/device_strided_batch_vector.hpp
```diff
@@ -212,9 +210,7 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
     //!
     hipError_t memcheck() const
     {
-        if(!m_enough_memory)
-            return hipErrorMemoryAllocation;
-        else if((bool)*this)
+        if((bool)*this)
```
leekillough: See above.

clients/include/device_vector.hpp
```diff
@@ -138,9 +134,7 @@ class device_vector : private d_vector<T, PAD, U>
 
     hipError_t memcheck() const
     {
-        if(!m_enough_memory)
-            return hipErrorMemoryAllocation;
-        else if((bool)*this)
+        if((bool)*this)
```
leekillough: Again

PR #1060, opened 2020-03-12
Move quick-return and invalid-size into own function

daineAMD:
See [LWPMLSE-37](http://ontrack-internal.amd.com/browse/LWPMLSE-37).
A proposal for moving quick-return and invalid_size checks into their own function to reduce the explosion of repeated test cases.
This takes any useful quick return and invalid_size combinations and moves them to their own `testing_xxx_arg_check` function and out of the regular quick tests.
For example, this will remove testing all `incx`/`incy` and `alpha`/`beta` combinations when `M == -1`.

This draft only implements this for gbmv to help discussion. If everyone likes this I can move ahead with other functions.

Previously quick gbmv tests (not batched nor strided_batched) ran 10368 tests in around 15.8s. With this change quick gbmv tests run 5618 tests in around 10.8s. A 5s (or ~33%) improvement but will add up across the > 100 functions being tested.

### DIFF
--- a/clients/gtest/gbmv_gtest.cpp
+++ b/clients/gtest/gbmv_gtest.cpp
@@ -39,8 +39,7 @@ namespace
             switch(GBMV_TYPE)
             {
             case GBMV:
-                return !strcmp(arg.function, "gbmv") || !strcmp(arg.function, "gbmv_bad_arg")
-                       || !strcmp(arg.function, "gbmv_arg_check");
+                return !strcmp(arg.function, "gbmv") || !strcmp(arg.function, "gbmv_bad_arg");
             case GBMV_BATCHED:
                 return !strcmp(arg.function, "gbmv_batched")
                        || !strcmp(arg.function, "gbmv_batched_bad_arg");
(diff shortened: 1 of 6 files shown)


clients/gtest/gbmv_gtest.cpp
```diff
@@ -39,7 +39,8 @@ namespace
             switch(GBMV_TYPE)
             {
             case GBMV:
-                return !strcmp(arg.function, "gbmv") || !strcmp(arg.function, "gbmv_bad_arg");
+                return !strcmp(arg.function, "gbmv") || !strcmp(arg.function, "gbmv_bad_arg")
+                       || !strcmp(arg.function, "gbmv_arg_check");
```
leekillough: Does `arg_check` need to be its own category, since it is just a combination of certain parameters and/or NaNs with a certain expected output?

Unlike `bag_arg`, I'm not sure we need to separate `arg_check` into its separate category.

daineAMD:
An obvious alternative is to just separate these cases in the yaml file as you mentioned; I'm fine with that solution as well.
Putting them into bad_arg is fine as well, I would just suggest renaming the bad_arg tests then since the name wouldn't really fit with what's happening in them.

clients/gtest/gbmv_gtest.cpp
daineAMD: Not necessarily. An alternative would be to split up the yaml categories into a sub-group for these edge cases. This is fine with me if we prefer that.

leekillough: I prefer separating out the YAML categories.

clients/include/testing_gbmv.hpp
```diff
@@ -167,22 +113,45 @@ void testing_gbmv(const Arguments& arg)
     // argument sanity check before allocating invalid memory
     if(M < 0 || N < 0 || lda < KL + KU + 1 || !incx || !incy || KL < 0 || KU < 0)
     {
```
TorreZuk: See the symm hemm which sets an invalid_size bool then you can just invalid_size || !M || !N and use the bool again for status check.

PR #1063, opened 2020-03-16
initial arg_check documentation

amcamd:


docs/source/argument_check.rst
```diff
@@ -0,0 +1,607 @@
...
+
+rocBLAS
+=======
+rocBLAS has 4 types of argument checking:
+
+1. ``rocblas_status_invalid_handle`` if the handle is a NULL pointer
+
+2. ``rocblas_status_invalid_pointer`` for NULL argument pointers
+
+3. ``rocblas_status_invalid_size`` for invalid size, increment or leading dimension argument
```
TorreZuk: Now have invalid_value which should be described:  E.g.  if a rocblas enum which isn't supported is passed.

docs/source/argument_check.rst
```diff
@@ -0,0 +1,607 @@
...
+        auto layer_mode = handle->layer_mode;
+ ------snip---logging---code---does-not-dereference---NULL---pointer--------------------------------------------------
+
+  //      if(!x)
+  //          return rocblas_status_invalid_pointer;
+
+  //      if(batch_count < 0)
+  //          return rocblas_status_invalid_size;
+
+        if(n <= 0 || incx <= 0 || batch_count <= 0)
```
TorreZuk: Batch count < 0 invalid_size is what I think is best, only quick success on == 0

docs/source/argument_check.rst
```diff
@@ -0,0 +1,607 @@
...
+
+     -Return rocblas_status_success
+
+6. Check for invalid sizes. Return rocblas_status_invalid_size if size arguments are invalid.
+
+7. Check for NULL pointer arguments not already covered by #3. Return rocblas_status_invalid_pointer if argument pointers are NULL.
+
+8. (Optional.) Allocate device memory, returning rocblas_status_memory_error if the allocation fails.
+
+9. Return rocblas_status_success, assuming no errors in HIP calls or other errors in the calculation.
```
TorreZuk: Would be good if this section can be reordered to match the general sequence of precedence on checks (if possible?).  i.e. first invalid_size and invalid_value are checked, .. quick return, ... null pointer checks only on data that must be de-referenced based on other values.   Whether the rules can be generalized across all functions we can also summarize.

docs/source/argument_check.rst
```diff
@@ -0,0 +1,607 @@
...
+
+        // leading dimensions must be valid
+        if(num_rows_a > lda || num_rows_b > ldb || num_rows_c > ldc || num_rows_d > ldd)
+            return rocblas_status_invalid_size;
+
+        // quick return m,n,k equal to 0 is valid in BLAS
+        if(!m || !n || !k || !batch_count)
+            return rocblas_status_success;
+
+       // --- on host if rocblas_pointer_mode_host, on device if rocblas_pointer_mode_device ---
```
TorreZuk: Only host pointer data can be queried for values that allow early success for identity operations.   Think the next lines should be alpha not !alpha ?

docs/source/argument_check.rst
amcamd: Thanks,  corrected.

docs/source/argument_check.rst
amcamd: This is a L1 function. There are no incorrect argument checks in L1, only quick return.

docs/source/argument_check.rst
amcamd: Reordered list

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+3. If the function requires temporary device memory, and ``handle->is_device_memory_size_query()`` is ``true``, then validate any pointers and arguments required to determine the optimal size of temporary device memory, returning ``rocblas_status_invalid_pointer`` or ``rocblas_status_invalid_size`` if the arguments are invalid, and otherwise ``return handle->set_optimal_device_memory_size(size...);``, where ``size...`` is a list of one or more sizes of temporary buffers, which are allocated with ``handle->device_malloc(size...)`` later.
+
+4. Perform logging if enabled, taking care not to dereference ``nullptr`` arguments.
+
+5. Check for invalid sizes. Return ``rocblas_status_invalid_size`` if size arguments are invalid.
+
+3. Check for unsupported enum value. Return ``rocblas_status_invalid_value`` if enum value is invalid.
+
+6. Return ``rocblas_status_invalid_pointer`` if any pointers used to determine quick return conditions are NULL.
+
```
TorreZuk: I think it is not just quick return condition testing that are tested.   Could it be said that any pointer required for calculation is NULL. (i.e. a pointer that is not ignored based on other arguments, for example N == 0 means ignore input pointers)

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+     - Return ``rocblas_status_success``
+
+   - If there is a return value
+
+     - If the return value pointer argument is nullptr, return ``rocblas_status_invalid_pointer``
+
+     - Else, return ``rocblas_status_success``
+
+9. Check for NULL pointer arguments not already covered by #3. Return ``rocblas_status_invalid_pointer`` if argument pointers are NULL.
+
```
TorreZuk: Reminder that renumbering was done.

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+10. (Optional.) Allocate device memory, returning ``rocblas_status_memory_error`` if the allocation fails.
+
+11. After completing the calculation, return ``rocblas_status_success``, assuming no errors in HIP calls or other errors in the calculation.
+
```
TorreZuk: Not after completing, "After launching the calculation".   Only know that the calculations were successfully launched asynchronously.

docs/source/argument_check.rst
amcamd: Thanks for catching this, corrected numbering.

docs/source/argument_check.rst
amcamd: I think after correcting the numbering is should be more clear.

AFAIK the Fortran philosophy is that if you can quick_return_success because N == 0 you do not need to check if pointers are NULL pointers. 

In the axpy example, there is a rocblas_status_success return if  alpha !=0 && *alpha == 0 before the check to return rocblas_status_invalid_pointer  if !x  || !y

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+To accommodate the additions
+============================
+
+- See Logging, below.
+
+- For batched and strided_batched L2 and L3 functions there is a quick-return-success for ``batch_count == 0``, and an invalid size error for ``batch_count < 0``. 
+
+- For batched and strided_batched L1 functions there is a quick-return-success for ``batch_count <= 0``
+
+- When ``rocblas_pointer_mode == rocblas_pointer_mode_device`` do not copy alpha and or beta from device to host for quick-return-success checks. In this case, ommit the quick-return-success checks for alpha and or beta.
```
daineAMD: Typo - "omit"

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+
+1. If ``handle == nullptr``, return ``rocblas_status_invalid_handle``.
+
+2. If the function does not require temporary device memory, call the macro ``RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);``.
+
+3. If the function requires temporary device memory, and ``handle->is_device_memory_size_query()`` is ``true``, then validate any pointers and arguments required to determine the optimal size of temporary device memory, returning ``rocblas_status_invalid_pointer`` or ``rocblas_status_invalid_size`` if the arguments are invalid, and otherwise ``return handle->set_optimal_device_memory_size(size...);``, where ``size...`` is a list of one or more sizes of temporary buffers, which are allocated with ``handle->device_malloc(size...)`` later.
+
+4. Perform logging if enabled, taking care not to dereference ``nullptr`` arguments.
+
+5. Check for invalid sizes. Return ``rocblas_status_invalid_size`` if size arguments are invalid.
```
daineAMD: 5 and 6 might be backwards, I think in LAPACK they check enums (rather, characters, for them) first then invalid_sizes.

docs/source/argument_check.rst
```diff
@@ -0,0 +1,636 @@
...
+- Scalar values in device memory are logged as their addresses. Scalar values in host memory are logged as their values, with a ``nullptr`` logged as ``NaN`` (``std::numeric_limits<T>::quiet_NaN()``).
+
+rocBLAS control flow:
+=====================
+
+1. If ``handle == nullptr``, return ``rocblas_status_invalid_handle``.
+
+2. If the function does not require temporary device memory, call the macro ``RETURN_ZERO_DEVICE_MEMORY_SIZE_IF_QUERIED(handle);``.
+
+3. If the function requires temporary device memory, and ``handle->is_device_memory_size_query()`` is ``true``, then validate any pointers and arguments required to determine the optimal size of temporary device memory, returning ``rocblas_status_invalid_pointer`` or ``rocblas_status_invalid_size`` if the arguments are invalid, and otherwise ``return handle->set_optimal_device_memory_size(size...);``, where ``size...`` is a list of one or more sizes of temporary buffers, which are allocated with ``handle->device_malloc(size...)`` later.
```
daineAMD: I'm ok with this, but I think it's inconsistent with the current implementation. In trsv, for example, we handle the query at the same place as the quick returns (below the logging and some error checks). So we'll have to go through and change those.

PR #1069, opened 2020-03-20
FIxing tbmv testing.

daineAMD:
I noticed while going through to reduce tests that tbmv didn't have any tests for N > lda > k, which is a valid use case. The tests were broken for this case, so I added tests and fixed the testing suite for this.

Will be reducing the number of tests here in a future commit (along with many other functions).

### DIFF
--- a/clients/include/testing_tbmv_strided_batched.hpp
+++ b/clients/include/testing_tbmv_strided_batched.hpp
@@ -103,53 +103,50 @@ void testing_tbmv_strided_batched(const Arguments& arg)
     rocblas_local_handle handle;
 
     // argument sanity check before allocating invalid memory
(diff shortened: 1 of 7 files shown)


clients/include/testing_tbmv_strided_batched.hpp
```diff
@@ -103,53 +103,50 @@ void testing_tbmv_strided_batched(const Arguments& arg)
     rocblas_local_handle handle;
 
     // argument sanity check before allocating invalid memory
-    if(M < 0 || K < 0 || lda < M || lda < 1 || !incx || K >= lda || batch_count <= 0)
+    bool invalidSize = M < 0 || K < 0 || lda < K + 1 || !incx || batch_count < 0;
+    if(invalidSize || !M || !batch_count)
     {
```
amcamd: can you use snake_case here rather than camelCase, so invalid_size rather than invalidSize

PR #1071, opened 2020-03-22
geam complex

amcamd:

### DIFF
--- a/clients/gtest/geam_gtest.cpp
+++ b/clients/gtest/geam_gtest.cpp
@@ -23,11 +23,10 @@ namespace
     // When the condition in the second argument is satisfied, the type combination
     // is valid. When the condition is false, this specialization does not apply.
     template <typename T>
-    struct geam_testing<
-        T,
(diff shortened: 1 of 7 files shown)


clients/gtest/geam_gtest.cpp
```diff
@@ -23,7 +23,11 @@ namespace
     // When the condition in the second argument is satisfied, the type combination
     // is valid. When the condition is false, this specialization does not apply.
     template <typename T>
-    struct geam_testing<T, std::enable_if_t<std::is_same<T, float>{} || std::is_same<T, double>{}>>
+    struct geam_testing<
+        T,
+        typename std::enable_if<std::is_same<T, float>{} || std::is_same<T, double>{}
+                                || std::is_same<T, rocblas_float_complex>{}
+                                || std::is_same<T, rocblas_double_complex>{}>::type>
```
TorreZuk: Not a bit deal but you could go back to the enable_if_t style and drop the ::type as it had it before

amcamd: Thanks for catching this, revert to enable_if_t style

PR #1073, opened 2020-03-23
Workaround script permissions

leekillough:


leekillough:
> 
> 
> So if the python rules changed are the old releases going to be broken?

We still need to figure out the root cause; this is just a workaround to allow the latest Tensile and rocBLAS to build, to unblock others.

If previous releases start failing too, then the Tensile workaround can be backported, or it might give more clues as to the root cause.

I don't know whether Python "rules" have changed, e.g. how file permissions are handled when copying files. I suspect it's something other than changes in Python behavior.

But being able to run a script without depending on execute permissions should not be a breaking change.

PR #1077, opened 2020-03-24
Use the venv for virtualenv for python 3

pfultz2:
The [`venv`](https://docs.python.org/3/library/venv.html) module from python 3 is the recommended virtualenv tool. This does require installing the `python3-venv` package.

This also fixes an error with `TensileCreateLibrary` where the local script can't be ran.


TorreZuk:
Paul may be busy so rebranched off develop and cherry picked along with adding the dependencies in install.sh https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/1078  (expect this one can be closed)

leekillough:
See #1078

PR #1078, opened 2020-03-24
Virtual env fix

TorreZuk:
This is based off develop branch not 3.0 like the earlier one, just adds dependency to Paul's simplification to use python3 venv.

### DIFF
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -6,56 +6,12 @@ if(NOT VIRTUALENV_PYTHON_EXE)
     find_program(VIRTUALENV_PYTHON_EXE python)
 endif()
 
-set(VIRTUALENV_SOURCE_DIR ${CMAKE_BINARY_DIR}/virtualenv-source CACHE PATH "Path to virtualenv source")
 set(VIRTUALENV_HOME_DIR ${CMAKE_BINARY_DIR}/virtualenv CACHE PATH "Path to virtual environment")
 
-set(VIRTUALENV_VERSION 15.1.0)
 function(virtualenv_create)
-    if (DETECT_LOCAL_VIRTUALENV)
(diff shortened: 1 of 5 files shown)


amdkila:
I haven't had the chance to test the last commit I made on any other distro except Ubuntu...I'm hoping CI cooperates. I will test it later in the afternoon on CentOS and SLES. The wildcard is used in case if the user has a different python version on their system- unfortunately CentOS and RHEL tie up all Python packages to each Python point release :(

amdkila:
I think we're home free now. Sorry- didn't get around to testing this until about 45 minutes ago. All distros have passed the docker stage and have created the virtual environment

TorreZuk:
No worries.   Not sure but I would have thought the python3-virtualenv is actually the old python module on those platforms that we should now be ignoring in this PR as switching to the venv module as both could exist, quite a messy naming situation.

amdkila:
Yeah, that was my assumption too, but I searched for all pages with venv in their name and found nothing. It appears to be the new one- [this page](https://software.opensuse.org/package/python3-virtualenv) on the openSUSE site gave me a clue that this appears to be the latest python3 virtualenv

TorreZuk:
virtualenv lives on a standalone module and venv should be included as a standard package in many python3 since 3.3, unfortunately venv is separated out in some distributions.   The virtualenv package shouldn't be needed but won't hurt either as Paul's code explicitly invokes venv.   The distros that don't have python3*-venv probably don't need anything beyond python3* unless they are really trying to annoy us which might be the case :-)

TorreZuk:
Still need a review @mahmoodw ?  Will wait for some more tests.   @amdkila  are you done with package adjustments for now?

amdkila:
> Still need a review @mahmoodw ? Will wait for some more tests. @amdkila are you done with package adjustments for now?

Yes I am

PR #1080, opened 2020-03-25
Use _Float16 only if clang >= 10 

amdkila:
- This is due to clang++ 9.0.0 incompatibility, which causes problems for hipBLAS.
- Will need to propagated to master as well


amdkila:
That's right Lee. When hcc was clang 9 based, there was _Float16 support, which is why this check was at clang 9 in the first place. But looks like some older clang 9.x.x do not have the support.

PR #1085, opened 2020-03-27
allow use of pre-installed packages as before

TorreZuk:
use pyyaml from site-install as handled outside of virtualenv

### DIFF
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -10,7 +10,7 @@ set(VIRTUALENV_HOME_DIR ${CMAKE_BINARY_DIR}/virtualenv CACHE PATH "Path to virtu
 
 function(virtualenv_create)
     execute_process(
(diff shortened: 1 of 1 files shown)


leekillough:
> `ensurepip` isn't installed by default on Debian/Ubuntu either, so `pip` cannot be found sometimes.
> 
> This allows it to be found in the host environment.

Reason for the Ubuntu errors (same errors I reported the other day and then temporarily backed off):

https://askubuntu.com/questions/879437/ensurepip-is-disabled-in-debian-ubuntu-for-the-system-python

https://stackoverflow.com/questions/39539110/pyvenv-not-working-because-ensurepip-is-not-available

https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1290847

Possible fix #1093

TorreZuk:
Change included in PR #1093 so closing

PR #1093, opened 2020-03-28
Cleanup source, fix venv

leekillough:
Remove non-fatal errors from BLIS installation in `install.sh` (such as copying a directory or creating a symlink which already exists, or popping off a directory stack early), fix `venv` invocation esp. w.r.t. `pip`, and remove unnecessary whitespace and unicode characters in `argument_check.rst`


leekillough:
I have verified that it works on Ubuntu (Debian) with respect to PIP, for both the old and new Tensile clients.

The change from:
```
 execute_process(
        COMMAND ${VIRTUALENV_HOME_DIR}/bin/python ${VIRTUALENV_HOME_DIR}/bin/pip install ${ARGN}
    )
```
to:
```
 execute_process( COMMAND ${VIRTUALENV_HOME_DIR}/bin/pip install ${ARGN} )
```

seems to help, by calling the virtualenv installed `pip` directly rather than through `python`.

PR #1098, opened 2020-03-31
Ger opt

TorreZuk:
* caches vector for reuse
* makes benchmark output consistent


TorreZuk:
The numbers aren't clear on the performance script plots.  I will make some python scripts to to plot these clearly based on the results from the performance master script.   Random variation is significant so I will likely try to plot that as well, I expect these memory bound functions have larger fluctuations.   This is a draft as I wanted an early check on the code as may not get the plots until later in the week.

TorreZuk:
Closing until ready

PR #1100, opened 2020-04-03
Add alarm to tests to detect hangs; reduce memory leaks in tests; improve rocblas_abort implementation; make sigsetjmp code conformant; add comments

leekillough:
Adds a 10-minute alarm for every test case (the maximum right now is around 2 minutes on GFX906). If a test takes more than 10 minutes, then all testing is aborted, and a message explaining why is displayed. This should prevent the hangs in the CI when an aborted test is recovered from and everything hangs afterwards, probably because of a deadlock from a still-locked mutex.

Make `catch_signals_and_exceptions_as_failures` nestable, and put it around rocBLAS calls as close as possible, to reduce memory leaks. If a signal occurs in a rocBLAS call, it will be caught and translated into a Google Test failure, but the surrounding vector allocations will still be able to be destructed.

Clean up the implementation to use strict standard-conforming `sigsetjmp` usage, and catch `std::terminate`.

Fix a couple of bugs where Boolean values in macros were defined like:
```
#define NAME false
```
It should be:
```
#define NAME 0
```
... so that it can be used in preprocessor expressions, e.g. `#if NAME`.

Add `rocblas_abort` to the public API,  and fix its implementation to be thread-safe.

Add comments explaining why some of the symbols are "poisoned" in `rocblas-test` and `rocblas-bench`.

Some reformatting caused by `clang-format`.

### DIFF
--- a/clients/gtest/rocblas_gtest_main.cpp
+++ b/clients/gtest/rocblas_gtest_main.cpp
@@ -4,14 +4,9 @@
 
 #include "rocblas_data.hpp"
 #include "rocblas_parse_data.hpp"
+#include "rocblas_test.hpp"
 #include "test_cleanup.hpp"
 #include "utility.hpp"
-#include <atomic>
-#include <cerrno>
-#include <csetjmp>
-#include <csignal>
-#include <cstdlib>
-#include <gtest/gtest.h>
 
 using namespace testing;
 
@@ -154,98 +149,6 @@ static void rocblas_set_listener()
     }
 
     listeners.Append(listener);
-}
-
-/*********************************************
- * Signal-handling for detecting test faults *
- *********************************************/
-// Id of the thread which is catching signals
-static volatile pthread_t rocblas_test_sighandler_tid;
-
-// sigjmp_buf describing stack frame to go back to
-static sigjmp_buf rocblas_sigjmp_buf;
-
-// Whether rocblas_sigjmp_buf is set and catching signals is enabled
-static volatile sig_atomic_t rocblas_sighandler_enabled = false;
-
-// Signal handler
-extern "C" void rocblas_test_signal_handler(int sig)
-{
-    int e = errno; // Save errno
-
-    // If the signal handler is disabled, restore this signal's disposition
-    // to default, and reraise the signal
-    if(!rocblas_sighandler_enabled)
-    {
-        signal(sig, SIG_DFL);
-        raise(sig);
-        errno = e; // Restore errno
-        return;
-    }
-
-    // If the thread receiving the signal is different from the thread
-    // catching signals, send the signal to the thread catching signals.
-    if(!pthread_equal(pthread_self(), rocblas_test_sighandler_tid))
-    {
-        pthread_kill(rocblas_test_sighandler_tid, sig);
-        sleep(1);
-        errno = e; // Restore errno
-        return;
-    }
-
-    // Jump back to the handler code
-    // Note: This bypasses stack unwinding, and may lead to memory leaks, but
-    // it is better than crashing. Throwing exceptions from signal handlers is
-    // poorly supported, and may result in recursive calls to std::terminate.
-    errno = e; // Restore errno
-    siglongjmp(rocblas_sigjmp_buf, sig);
-}
-
-// Set up signal handlers
-static void rocblas_test_sigaction()
-{
-    struct sigaction act;
-    sigfillset(&act.sa_mask);
-    act.sa_flags   = 0;
-    act.sa_handler = rocblas_test_signal_handler;
-
-    for(int sig : {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGPIPE, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
-    {
-        sigaction(sig, &act, nullptr);
-    }
-}
-
-// Lambda wrapper which detects signals and exceptions in an invokable function
-void catch_signals_and_exceptions_as_failures(const std::function<void()>& test)
-{
-    // Save this thread's id, to detect signals in different threads
-    rocblas_test_sighandler_tid = pthread_self();
-
-    // Set up the return point
-    int sig = sigsetjmp(rocblas_sigjmp_buf, true);
-
-    // If this is a return, indicate the signal received
-    if(sig)
-    {
-        rocblas_sighandler_enabled = false;
-        FAIL() << "Received " << strsignal(sig) << " signal";
-    }
-    else
-    {
-        // Run the test function, catching signals and exceptions
-        // Disable the signal handler after running the test
-        rocblas_sighandler_enabled = true;
-        try
-        {
-            test();
-            rocblas_sighandler_enabled = false;
-        }
-        catch(...)
-        {
-            rocblas_sighandler_enabled = false;
-            FAIL() << "Received unhandled exception";
-        }
-    }
 }
 
 // Print Version
--- a/clients/gtest/rocblas_test.cpp
+++ b/clients/gtest/rocblas_test.cpp
@@ -1,0 +1,149 @@
+#include "rocblas_test.hpp"
+#include <cerrno>
+#include <csetjmp>
+#include <csignal>
+#include <cstdlib>
+#include <exception>
+#include <pthread.h>
+#include <unistd.h>
+
+/*********************************************
+ * Signal-handling for detecting test faults *
+ *********************************************/
+
+// State of the signal handler
+static struct
+{
+    // Whether sigjmp_buf is set and catching signals is enabled
+    volatile sig_atomic_t enabled;
+
+    // Id of the thread which is catching signals
+    volatile pthread_t tid;
+
+    // sigjmp_buf describing stack frame to go back to
+    sigjmp_buf sigjmp_buf;
+
+    // The signal which was received
+    volatile sig_atomic_t signal;
+} handler;
+
+// Signal handler (must have external "C" linkage)
+extern "C" void rocblas_test_signal_handler(int sig)
+{
+    int saved_errno = errno; // Save errno
+
+    // If the signal handler is disabled, because we're not in the middle of
+    // running a rocBLAS test, restore this signal's disposition to default,
+    // and reraise the signal
+    if(!handler.enabled)
+    {
+        signal(sig, SIG_DFL);
+        errno = saved_errno;
+        raise(sig);
+        return;
+    }
+
+    // If the thread receiving the signal is different from the thread
+    // catching signals, send the signal to the thread catching signals.
+    if(!pthread_equal(pthread_self(), handler.tid))
+    {
+        pthread_kill(handler.tid, sig);
+        errno = saved_errno;
+        return;
+    }
+
+    // If this is an alarm timeout, we abort
+    if(sig == SIGALRM)
+    {
+        static constexpr char msg[]
+            = "\nAborting tests due to an alarm timeout.\n\n"
+              "This could be due to a deadlock caused by mutexes being left locked\n"
+              "after a previous test's signal was caught and partially recovered from.\n";
+        // We must use write() because it's async-signal-safe and other IO might be blocked
+        write(STDERR_FILENO, msg, sizeof(msg) - 1);
+        rocblas_abort();
+    }
+
+    // Jump back to the handler code after setting handler.signal
+    // Note: This bypasses stack unwinding, and may lead to memory leaks, but
+    // it is better than crashing.
+    handler.signal = sig;
+    errno          = saved_errno;
+    siglongjmp(handler.sigjmp_buf, true);
+}
+
+// Set up signal handlers
+void rocblas_test_sigaction()
+{
+    // Set up a separate stack in case stack overflows occur
+    alignas(64) static char stack_memory[SIGSTKSZ];
+    stack_t                 stack = {stack_memory, 0, sizeof(stack_memory)};
+
+    struct sigaction act;
+    sigfillset(&act.sa_mask);
+
+    // Signal handler above
+    act.sa_handler = rocblas_test_signal_handler;
+
+    // We use the stack if sigaltstack() returns success
+    act.sa_flags = !sigaltstack(&stack, nullptr) ? SA_ONSTACK : 0;
+
+    for(int sig : {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGALRM, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
+        sigaction(sig, &act, nullptr);
+}
+
+// Number of seconds each test is allowed to take before all testing is killed.
+constexpr unsigned TEST_TIMEOUT = 600;
+
+// Lambda wrapper which detects signals and exceptions in an invokable function
+void catch_signals_and_exceptions_as_failures(std::function<void()> test, bool set_alarm)
+{
+    // Save the current handler (to allow nested calls to this function)
+    auto old_handler = handler;
+
+    // Set up the return point, and handle siglongjmp returning back to here
+    if(sigsetjmp(handler.sigjmp_buf, true))
+    {
+        FAIL() << "Received " << strsignal(handler.signal) << " signal";
+    }
+    else
+    {
+        // Test timeout can be overriden by the ROCBLAS_TEST_TIMEOUT environment variable
+        static const unsigned test_timeout = []() {
+            unsigned    timeout;
+            const char* env = getenv("ROCBLAS_TEST_TIMEOUT");
+            return env && sscanf(env, "%u", &timeout) == 1 ? timeout : TEST_TIMEOUT;
+        }();
+
+        // Alarm to detect deadlocks or hangs
+        if(set_alarm)
+            alarm(test_timeout);
+
+        // Save this thread's id, to detect signals in different threads
+        handler.tid = pthread_self();
+
+        // Enable the signal handler
+        handler.enabled = true;
+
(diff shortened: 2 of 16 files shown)


clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -188,17 +209,17 @@ extern "C" void rocblas_test_signal_handler(int sig)
...
+    errno = saved_errno;
+
+    // Jump back to the handler code after setting signal_received
     // Note: This bypasses stack unwinding, and may lead to memory leaks, but
-    // it is better than crashing. Throwing exceptions from signal handlers is
-    // poorly supported, and may result in recursive calls to std::terminate.
-    errno = e; // Restore errno
-    siglongjmp(rocblas_sigjmp_buf, sig);
+    // it is better than crashing.
+    signal_received = sig;
```
leekillough, line 221: We would have to list every signal in a `switch` statement if we send the signal through `siglongjmp`, so we set a global variable instead.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,10 +230,15 @@ static void rocblas_test_sigaction()
     act.sa_flags   = 0;
     act.sa_handler = rocblas_test_signal_handler;
 
-    for(int sig : {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGPIPE, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
-    {
+    for(int sig : {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGALRM, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
```
leekillough, line 233: `SIGPIPE` is was replaced with `SIGALRM` because `SIGPIPE` is asynchronous and we don't need it, while we need `SIGALRM`

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -221,31 +247,50 @@ void catch_signals_and_exceptions_as_failures(const std::function<void()>& test)
     // Save this thread's id, to detect signals in different threads
     rocblas_test_sighandler_tid = pthread_self();
 
-    // Set up the return point
-    int sig = sigsetjmp(rocblas_sigjmp_buf, true);
+    // Catch std::terminate
+    // old_terminate must be volatile to survive across sigsetjmp
+    volatile auto old_terminate = std::set_terminate(terminate_handler);
```
leekillough, line 252: On the chance that thrown exceptions create `std::terminate`, we can recover from it here.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -221,31 +247,50 @@ void catch_signals_and_exceptions_as_failures(const std::function<void()>& test)
...
 
-    // If this is a return, indicate the signal received
-    if(sig)
+    // Set up the return point, and handle siglongjmp returns back to here
+    switch(sigsetjmp(rocblas_sigjmp_buf, true))
```
leekillough, line 255: We cannot store the return value of `sigsetjmp` into a variable and be 100% standards compliant -- its return value can only be used as the controlling expression of a control statement, or discarded.

clients/include/testing_gemm_batched_ex.hpp
```diff
@@ -15,7 +15,7 @@
 #include "unit.hpp"
 #include "utility.hpp"
 
-#define DEBUG_PRINT false
+#define DEBUG_PRINT 0
```
leekillough, line 18: If `DEBUG_PRINT` is used in statements like `#if DEBUG_PRINT`, then it needs to be a numerical value rather than `true` or `false` .

clients/include/utility.hpp
```diff
@@ -30,13 +30,31 @@
...
+// puts, putchar, fputs, printf, fprintf, vprintf, vfprintf: Use rocblas_cout or rocblas_cerr
+// sprintf, vsprintf: Possible buffer overflows; us snprintf or vsnprintf instead
+// strerror: Thread-unsafe; use snprintf / dprintf with %m or strerror_* alternatives
+// strtok: Thread-unsafe; use strtok_r
+// gmtime, ctime, asctime, localtime: Thread-unsafe
+// tmpnam: Thread-unsafe; use mkstemp or related functions instead
+// putenv: Use setenv instead
+// clearenv, fcloseall, ecvt, fcvt: Miscellaneous thread-unsafe functions
+// sleep: Might interact with signals by using alarm(); use nanosleep() instead
+// abort: Does not abort as cleanly as rocblas_abort, and can be caught by a signal handler
```
leekillough, line 50: This explains why all the symbols are poisoned for `rocblas-test` and `rocblas-bench`, to catch common programming errors. The poisoning is not applied to the library or samples, because we cannot control outside users.

library/include/rocblas-functions.h
```diff
@@ -12476,18 +12476,18 @@ ROCBLAS_EXPORT rocblas_status rocblas_zsyrkx_strided_batched(rocblas_handle
...
-           A is not referenced.
+        When uplo == rocblas_fill_lower the  leading  k by k
+        lower triangular part of the array  A must contain the lower
+        triangular matrix  and the strictly upper triangular part of
+        A is not referenced.
 
-           Note that when  diag == rocblas_diagonal_unit  the diagonal elements of
-           A  are not referenced either,  but are assumed to be  unity.
+        Note that when  diag == rocblas_diagonal_unit  the diagonal elements of
+        A  are not referenced either,  but are assumed to be  unity.
```
leekillough, line 12490: `clang-format` sometimes makes changes in comments' indentation.

library/include/rocblas-functions.h
```diff
@@ -15350,7 +15350,7 @@ ROCBLAS_EXPORT const char* rocblas_status_to_string(rocblas_status status);
 /* \brief Initialize rocBLAS, to avoid costly startup time at the first call.
 */
 
-ROCBLAS_EXPORT void rocblas_init();
+ROCBLAS_EXPORT void rocblas_init(void);
```
leekillough, line 15353: For functions declared int `rocblas-functions.h` or any other public headers which must compile with both C and C++, we should use `(void)` instead of `()` to indicate no arguments, because `(void)` is the same on C and C++, while `()` is old K&R syntax (before prototypes) in C.

library/include/rocblas-functions.h
```diff
@@ -15436,6 +15436,12 @@ ROCBLAS_EXPORT rocblas_status rocblas_set_device_memory_size(rocblas_handle hand
  ******************************************************************************/
 ROCBLAS_EXPORT bool rocblas_is_managing_device_memory(rocblas_handle handle);
 
+/*! \brief
+    \details
+    Abort function which safely flushes all IO
+ ******************************************************************************/
+ROCBLAS_EXPORT void rocblas_abort(void) __attribute__((__noreturn__));
```
leekillough, line 15443: `rocblas_abort` should be added as public function, for people who need to abort, but in a way that's clean to rocBLAS.

clients/include/rocblas_test.hpp
```diff
@@ -54,7 +54,19 @@
...
+#define EXPECT_ROCBLAS_STATUS(STATUS, EXPECT)                               \
+    do                                                                      \
+    {                                                                       \
+        rocblas_status stat = rocblas_status(-1);                           \
+        catch_signals_and_exceptions_as_failures([&] { stat = (STATUS); }); \
+        if(stat == rocblas_status(-1))                                      \
+            return;                                                         \
+        else                                                                \
+            ASSERT_EQ(stat, (EXPECT));                                      \
+    } while(0)
```
leekillough, line 69: This change should remove most memory leaks. It puts another `catch_signals_and_exceptions_as_failures` wrapper around rocBLAS calls.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,43 +233,71 @@ static void rocblas_test_sigaction()
...
 
-    // Set up the return point
-    int sig = sigsetjmp(rocblas_sigjmp_buf, true);
+// Lambda wrapper which detects signals and exceptions in an invokable function
+void catch_signals_and_exceptions_as_failures(const std::function<void()>& test, bool set_alarm)
+{
+    // Save the current state (to allow nested calls to this function)
+    // Variables must be volatile to survive across sigsetjmp
+    volatile auto old_state     = state;
+    volatile auto old_terminate = std::set_terminate(terminate_handler); // Catch std::terminate
```
leekillough, line 254: We allow nested calls to `catch_signals_and_exceptions_as_failures`, catching signals as close as possible to their cause, and thus decreasing the chances of memory leaks.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,43 +233,71 @@ static void rocblas_test_sigaction()
...
```
leekillough, line 253: We allow nested calls to `catch_signals_and_exceptions_as_failures`, catching signals as close as possible to their cause, and thus decreasing the chances of memory leaks.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,43 +233,71 @@ static void rocblas_test_sigaction()
...
+        break;
     }
+
+    // Cancel the alarm if it was set
+    if(set_alarm)
+        alarm(0);
+
+    // Restore the previous state
+    std::set_terminate(old_terminate);
+    state = const_cast<const decltype(state)&>(old_state);
```
leekillough, line 300: This copies the old state back, but requires weird casts.

clients/include/rocblas_test.hpp
```diff
@@ -118,11 +130,12 @@ bool match_test_category(const Arguments& arg, const char* category);
     INSTANTIATE_TEST_CATEGORY(testclass, known_bug)
 
 // Function to catch signals and exceptions as failures
-void catch_signals_and_exceptions_as_failures(const std::function<void()>& test);
+void catch_signals_and_exceptions_as_failures(const std::function<void()>& test,
+                                              bool                         set_alarm = false);
```
leekillough, line 134: By default we do not set an alarm timer.

clients/include/rocblas_test.hpp
```diff
@@ -118,11 +130,12 @@ bool match_test_category(const Arguments& arg, const char* category);
...
 
 // Macro to call catch_signals_and_exceptions_as_failures() with a lambda expression
 #define CATCH_SIGNALS_AND_EXCEPTIONS_AS_FAILURES(test) \
-    catch_signals_and_exceptions_as_failures([&] { test; })
+    catch_signals_and_exceptions_as_failures([&] { test; }, true)
```
leekillough, line 138: But on the outer level of the tests, we set an alarm.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -159,46 +159,70 @@ static void rocblas_set_listener()
...
+{
+    // Whether rocblas_sigjmp_buf is set and catching signals is enabled
+    volatile sig_atomic_t rocblas_sighandler_enabled;
+
+    // Id of the thread which is catching signals
+    volatile pthread_t rocblas_test_sighandler_tid;
+
+    // sigjmp_buf describing stack frame to go back to
+    sigjmp_buf rocblas_sigjmp_buf;
+} state;
```
leekillough, line 174: The structure containing the state of the signal handler. It is pushed/popped when nested.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -159,46 +159,70 @@ static void rocblas_set_listener()
...
+    // If this is an alarm timeout, we abort
+    if(sig == SIGALRM)
+    {
+        static constexpr char msg[]
+            = "\nAborting tests due to an alarm timeout.\n\n"
+              "This could be due to a deadlock caused by mutexes being left locked\n"
+              "after a previous test's signal was caught and partially recovered from.\n";
+        write(STDERR_FILENO, msg, sizeof(msg) - 1);
+        rocblas_abort();
+    }
```
leekillough, line 198: When more than 10 minutes have elapsed, all tests are aborted, and this message is displayed.

Hangs may occur if a signal or exception occurs during a test and a mutex is left locked. This alarm catches that and prevents hanging tests.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,43 +233,71 @@ static void rocblas_test_sigaction()
...
+        FAIL() << "Received " << strsignal(signal_received) << " signal";
+        break;
+
+    case -1: // terminate_handler()
+        FAIL() << "Received std::terminate";
+        break;
+
+    default: // Unknown return value
+        FAIL() << "Internal error in rocblas_gtest_main.cpp";
+        break;
```
leekillough, line 291: `sigsetjmp()` can return `0` after a normal setup call, `1` if `siglongjmp` was used with an argument of `1` to return, `-1` if `terminate_handler` was used. Other than those values, it's an internal error.

clients/gtest/rocblas_gtest_main.cpp
```diff
@@ -209,43 +233,71 @@ static void rocblas_test_sigaction()
...
     }
+
+    // Cancel the alarm if it was set
+    if(set_alarm)
+        alarm(0);
+
+    // Restore the previous state
+    std::set_terminate(old_terminate);
+    state = const_cast<decltype(state)&>(old_state);
```
leekillough, line 300: To support nesting, the old state and `terminate`  handler are restored.

leekillough: > To support nesting, the old state and `terminate` handler are restored.

See [rocblas_test.cpp](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/32eb05b9b88a7b375e7c39baa01b63187f863b01/clients/gtest/rocblas_test.cpp), where all of the signal- and exception-handling code was moved to.  Simpler and more descriptive names and structures have been used.

clients/include/rocblas_test.hpp
```diff
@@ -54,7 +54,19 @@
         }                                         \
     } while(0)
 
-#define EXPECT_ROCBLAS_STATUS ASSERT_EQ
+// This wraps the rocBLAS call with catch_signals_and_exceptions_as_failures().
+// By placing it at the rocBLAS call site, memory resources are less likely to
+// be leaked in the event of a caught signal.
+#define EXPECT_ROCBLAS_STATUS(STATUS, EXPECT)                               \
+    do                                                                      \
```
TorreZuk, line 61: CHECK_ROCBLAS_ERROR is defined to this too, which is probably alright.   We seem to have some inconsistent checks to status and error in some of the benchmark timing loops which we will likely want to remove.

clients/include/rocblas_test.hpp
TorreZuk, line 138: See the alarm going off in CI for gemm_ex so not sure this is expected.  If the process is blocked the alarm doesn't take this into account I don't think.

clients/gtest/rocblas_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    for(int sig : {SIGABRT, SIGBUS, SIGFPE, SIGILL, SIGALRM, SIGSEGV, SIGSYS, SIGUSR1, SIGUSR2})
+        sigaction(sig, &act, nullptr);
+}
+
+// Number of seconds each test is allowed to take before all testing is killed.
+// If this time limit is exceeded for any particular test, it is assumed that
+// there is a permanent deadlock, and testing is terminated.
+constexpr int TEST_TIMEOUT = 600;
+
```
TorreZuk, line 107: Often CI sets testing timeouts so not sure if we want to expose to them as environment variable.  E.g. simulator would require huge timeout.

clients/include/rocblas_test.hpp
leekillough: > 
> 
> CHECK_ROCBLAS_ERROR is defined to this too, which is probably alright. We seem to have some inconsistent checks to status and error in some of the benchmark timing loops which we will likely want to remove.

`#ifdef GOOGLE_TEST` is used to tell `rocblas-test` apart from `rocblas-bench`. The same macros are used by both, but they have different definitions for `rocblas-test` and `rocblas-bench`. This change only affects `rocblas-test`, and so it only modifies the macros which are defined when `GOOGLE_TEST` is defined. We do not want them to be consistent except maybe in their API, because they have different requirements between unit testing and benchmarking.

clients/include/rocblas_test.hpp
leekillough: > See the alarm going off in CI for gemm_ex so not sure this is expected. 

If a signal is raised because of a fatal HIP memory error, the alarm might go off on the next test because of deadlocks.

> If the process is blocked the alarm doesn't take this into account I don't think.

The alarm is not activated until a single `gemm_ex` test in `rocblas-test` starts. If the test executable is delayed from starting in the CI, or if there is a long delay between two individual `rocblas-test` tests, then the alarm won't do anything. The alarm is only active between the beginning and ending of a test case. It's possible but rare that a test starts, and then it's suspended by the CI, and then when it wakes up, the alarm fires, but that is unlikely. The alternative of hangs seems much worse.

clients/gtest/rocblas_test.cpp
leekillough: The `ROCBLAS_TEST_TIMEOUT` environment variable has been added to allow you to override the default of `600` seconds.

leekillough: If you set `ROCBLAS_TEST_TIMEOUT=0`, you will disable the alarm entirely.

PR #1107, opened 2020-04-06
Ger opt

TorreZuk:
* Fixed benchmarking inconsistencies and optimized

### DIFF
--- a/clients/include/testing_ger.hpp
+++ b/clients/include/testing_ger.hpp
@@ -2,6 +2,7 @@
  * Copyright 2018-2020 Advanced Micro Devices, Inc.
  * ************************************************************************ */
 
+#include "bytes.hpp"
 #include "cblas_interface.hpp"
 #include "flops.hpp"
 #include "near.hpp"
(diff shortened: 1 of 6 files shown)


clients/include/testing_ger.hpp
```diff
@@ -194,20 +195,21 @@ void testing_ger(const Arguments& arg)
             rocblas_ger<T, CONJ>(handle, M, N, &h_alpha, dx, incx, dy, incy, dA_1, lda);
         }
 
-        gpu_time_used     = (get_time_us() - gpu_time_used) / number_hot_calls;
+        gpu_time_used = (get_time_us() - gpu_time_used) / number_hot_calls;
+
         rocblas_gflops    = ger_gflop_count<T, CONJ>(M, N) / gpu_time_used * 1e6;
-        rocblas_bandwidth = (2.0 * M * N) * sizeof(T) / gpu_time_used / 1e3;
+        rocblas_bandwidth = ger_gbyte_count<T>(M, N) / gpu_time_used * 1e6;
```
daineAMD, line 201: Is this already updated in ger_batched and ger_strided_batched?

TorreZuk: Good catch will update them now

PR #1110, opened 2020-04-07
Fix trivial off-by-1 error (1 too small)

leekillough:
-

### DIFF
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -101,7 +101,7 @@ rocblas_int query_device_property()
             char buf[320];
             snprintf(
                 buf,
-                sizeof(buf) - 1,
+                sizeof(buf),
                 "Device ID %d : %s\n"
                 "with %3.1f GB memory, clock rate %d MHz @ computing capability %d.%d \n"
                 "maxGridDimX %d, sharedMemPerBlock %3.1f KB, maxThreadsPerBlock %d, warpSize %d\n",


leekillough:
> 
> 
> Implementations of snprintf vary across platforms. If you don't check the return code (which is a pain) it is safer to keep the -1 size but you must guaranteed null termination by setting buf[319] = 0; before the call.

`snprintf` is guaranteed to 0-terminate the string, even if it is truncated. The second argument is the length of the entire array, which will always be 0-terminated, regardless of how much of the string is written.

`snprintf` is used here not for error-checking, but to avoid buffer overflows. It would be a lot of extra unnecessary work to tell whether the string was truncated, but even if it is truncated, we guarantee that it's still 0-terminated and that it won't overwrite the array.

The original length was 1 shorter than it could have been.

leekillough:
```
int snprintf( char *restrict buffer, size_t bufsz, const char *restrict format, ... );
```
> Writes the results to a character string buffer. At most `bufsz - 1` characters are written. The resulting character string will be terminated with a null character, unless `bufsz` is zero. If `bufsz` is zero, nothing is written and buffer may be a null pointer, however the return value (number of bytes that would be written not including the null terminator) is still calculated and returned.

That's the language of the standard since C99 and C++11. If some compilers "vary", they are non-compliant.

TorreZuk:
Sorry going off recollections, sounds like it should be consistently implemented now.  Used to be a problem on windows, solaris, and likely others so seeing the -1 that reminded me of the workaround.

PR #1115, opened 2020-04-08
Batched memcheck fixes

daineAMD:
Should resolve [SWDEV-220984](http://ontrack-internal.amd.com/browse/SWDEV-220984). Apparently I missed updating a bunch of batched tests to the new memory checking scheme. This should use `CHECK_DEVICE_ALLOCATION` instead of `CHECK_HIP_ERROR` when calling `memcheck` on all device vectors now.

Also updates gemm_batched_ex test to use new batch_vectors as these tests were segfaulting apparently.

### DIFF
--- a/clients/include/norm.hpp
+++ b/clients/include/norm.hpp
@@ -239,14 +239,15 @@ double norm_check_general(char           norm_type,
 }
 
 /* ============== Norm Check for batched case ============= */
+
 template <typename T, typename T_hpa>
-double norm_check_general(char               norm_type,
-                          rocblas_int        M,
-                          rocblas_int        N,
-                          rocblas_int        lda,
-                          host_vector<T_hpa> hCPU[],
-                          host_vector<T>     hGPU[],
-                          rocblas_int        batch_count)
+double norm_check_general(char                      norm_type,
+                          rocblas_int               M,
+                          rocblas_int               N,
+                          rocblas_int               lda,
+                          host_batch_vector<T_hpa>& hCPU,
+                          host_batch_vector<T>&     hGPU,
+                          rocblas_int               batch_count)
 {
     // norm type can be O', 'I', 'F', 'o', 'i', 'f' for one, infinity or Frobenius norm
     // one norm is max column sum
--- a/clients/include/testing_gemm_batched_ex.hpp
+++ b/clients/include/testing_gemm_batched_ex.hpp
@@ -374,7 +374,7 @@ void testing_gemm_batched_ex(const Arguments& arg)
     hD_2.copy_from(hD_1);
     for(int b = 0; b < batch_count; b++)
     {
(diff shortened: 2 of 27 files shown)


clients/include/norm.hpp
```diff
@@ -239,14 +239,15 @@ double norm_check_general(char           norm_type,
...
 
 /* ============== Norm Check for batched case ============= */
+
 template <typename T, typename T_hpa>
-double norm_check_general(char               norm_type,
-                          rocblas_int        M,
-                          rocblas_int        N,
-                          rocblas_int        lda,
-                          host_vector<T_hpa> hCPU[],
-                          host_vector<T>     hGPU[],
```
TorreZuk, line 248: Are you sure you replaced all the old style host_vector arrays?  I guess CI testing will tell.

clients/include/testing_gemm_batched_ex.hpp
```diff
@@ -391,8 +369,15 @@ void testing_gemm_batched_ex(const Arguments& arg)
...
             rocblas_init<To>(hC[b], M, N, ldc);
         rocblas_init<To>(hD_1[b], M, N, ldd);
-        hD_2[b]    = hD_1[b];
-        hD_gold[b] = hD_1[b];
+    }
+
+    hD_2.copy_from(hD_1);
+    for(int b = 0; b < batch_count; b++)
+    {
+        for(int i = 0; i < size_d; i++)
```
TorreZuk, line 377: size_t i

leekillough:
General comment: We should not use `hipError_t` in our rocBLAS APIs, such as arguments or return values of our functions, unless we are calling a function or macro which specifically converts or tests `hipError_t`.

It's apples and oranges for a rocBLAS routine to return `hipError_t`  values for its functions, and it limits rocBLAS's expressibility to only being able to express `hipError_t` values.

I mean, it would be like writing your own library, but being restricted to passing and returning `errno` values. "What `errno` value do I use for `<this>` error?"

It would be better for the `hipError_t` to be encapsulated, and for a rocBLAS class to have a test for success, like `explicit operator bool() const`, which is how the allocation functions were initially implemented. You allocate, and then you test `if (allocated_memory)` to check if it was successful. The details of `hipError_t` are hidden under the hood.

Also: `CHECK_DEVICE_ALLOCATION` was recently modified (and simplified), [as per this (#25)](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.github/CONTRIBUTING.md)

clients/include/testing_gemm_batched_ex.hpp
leekillough: > 
> 
> size_t i

I have an upcoming PR on this issue, a source-wide change to use integer types better.

PR #1120, opened 2020-04-09
adds export include for rocblas_complex

TorreZuk:
* adds include for file to be standalone

### DIFF
--- a/library/include/rocblas-complex-types.h
+++ b/library/include/rocblas-complex-types.h
@@ -35,6 +35,7 @@ typedef struct
 #include <hip/hip_runtime.h>
 #include <math.h>
 #include <ostream>
+#include <rocblas-export.h>
 #include <type_traits>
 
 /*! \brief rocblas_complex_num is a structure which represents a complex number


library/include/rocblas-complex-types.h
```diff
@@ -35,6 +35,7 @@ typedef struct
 #include <hip/hip_runtime.h>
 #include <math.h>
 #include <ostream>
+#include <rocblas-export.h>
```
leekillough, line 38: Do you need this for C as well, or just C++?

TorreZuk: This is for C++
This should have been included as it uses the ROCBLAS_EXPORT define.

leekillough: Should it be `<rocblas-export.h>` or `"rocblas-export.h"`?

`rocblas.h` already #includes `rocblas-export.h`.
If this is a rocBLAS client, it should include `rocblas.h` or should include `rocblas-export.h` individually.

There's no reason for `rocblas-complex-types.h` to `#include` another header file. That should be the responsibility of the `#include`r.

rocSOLVER should use rocBLAS, not intrude upon it.

Be prepared for the consequences.

TorreZuk: I guess it could just as well be " " included but won't bother changing for now.    This was actually for some experiments in rocBLAS-Examples and hip-clang I noticed it wasn't self contained.  

True enough the export of complex is for rocSOLVER and will have to be carefully managed.  Will try to avoid exporting any other helper types that were put in the APIs to avoid complications.

leekillough: It's more because of the mutual dependencies which might be created when header files include each other.

`rocblas.h`: depends on `rocblas-complex-types.h` and `rocblas-export.h`

`rocblas-complex-types-h`: depends on `rocblas-export.h`

(There's no cycle, yet.)

I would have `#include`d `rocblas-export.h` separately in rocSOLVER, rather than adding it to `rocblas-complex-types.h`. Some projects like SPARSE might use our `rocblas-complex-types.h` for their own source, so it would be nice if it's a "leaf" header and doesn't include others (except HIP and system headers).

The problem most like likely occur in these cases:
1) A cycle is created, and one of the headers does not have a `#pragma once` or `#ifndef` include guard.
2. One header file gets included through another, and then a third file or header doesn't get its dependencies satisfied, because the header files were included out of the order expected. Usually this can be avoided by each header `#include`ing everything it depends on, and that's what `header_compilation_tests.sh` was designed to ensure, but I've seen some cases where it still doesn't work. (`header_compilation_tests.s` is currently disabled for `hip-clang`, so someone will need to port it unless they are willing to leave it disabled, and thus not get the benefit of checking header files for problems.)

3. The poisoned identifiers (only applies to `rocblas-test` and `rocblas-bench`, not the library) are poisoned before a required header is `#include`d, and then when it is `#include`d, it gets poisoned errors. When you `#include` a header before an identifier is poisoned, it is not poisoned for that header. So if you ever see poisoned identifiers from a system header, it needs to be `#include`d in `clients/include/utility.hpp` _before_ the poisoning occurs.

PR #1128, opened 2020-04-13
Rot fixes

daineAMD:
Changes to rot, rotm, rotg, and rotmg.

- Fixes error-checking and quick-return order
- Allows negative incx and incy for rot and rotm
- Does not test for incx == 0 || incy == 0 for rot or rotm as both x and y are outputs.

### DIFF
--- a/clients/gtest/blas1_gtest.yaml
+++ b/clients/gtest/blas1_gtest.yaml
@@ -22,6 +22,9 @@ Definitions:
     - { incx:  1, incy: -1 }
     - { incx: -1, incy:  1 }
     - { incx: -1, incy: -1 }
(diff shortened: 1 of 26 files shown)


daineAMD:
The .yaml file shouldn't need any changes, they already test negative increments for the appropriate rot* tests (currently for pre-checkin), they just previously handled the cases differently. I'm not checking incx == 0 || incy == 0 as the results are indeterminate.

daineAMD:
CI is failing an unrelated gemm_ex bug (HSA_STATUS_ERROR_OUT_OF_RESOURCES). I will merge this soon since it should be safe.

daineAMD:
(accidentally hit closed rather than comment... oops!)

PR #1129, opened 2020-04-14
Make rocblas_cout / rocblas_cerr's rocblas_ostream thread-local

leekillough:
Make `rocblas_cout` and `rocblas_cerr` have thread-local `rocblas_ostream`

### DIFF
--- a/library/src/include/rocblas_ostream.hpp
+++ b/library/src/include/rocblas_ostream.hpp
@@ -221,14 +221,14 @@ public:
     // Implemented as singleton to avoid the static initialization order fiasco
     static rocblas_ostream& cout()
     {
-        static rocblas_ostream cout{STDOUT_FILENO};
+        thread_local rocblas_ostream cout{STDOUT_FILENO};
         return cout;
     }
 
     // Implemented as singleton to avoid the static initialization order fiasco
     static rocblas_ostream& cerr()
     {
-        static rocblas_ostream cerr{STDERR_FILENO};
+        thread_local rocblas_ostream cerr{STDERR_FILENO};
         return cerr;
     }
 


leekillough:
It hasn't been stated explicitly yet, but a single `rocblas_ostream` object should only be used by one thread at a time. 

Each `rocblas_ostream` object contains a `std::ostringstream` object, which will be corrupted if written to by two threads concurrently.

However, it is okay for two `rocblas_ostream` objects to refer to the same file, and in fact, that's what the recent refactoring enabled.

We need to make `rocblas_cout` and `rocblas_cerr` have `thread_local` `rocblas_ostream`, even if they are referring to the same file, because different threads need to be able to buffer output without interference or mutexes with other threads.

The destructor will be called when the thread exits, which decreases the reference count to the worker, which will be destroyed if it's the last worker.

evetsso:
> It hasn't been stated explicitly yet, but a single rocblas_ostream object should only be used by one thread at a time.

Should this be stated explicitly now in the doc files?

TorreZuk:
If it will be thread local is there still a need for a worker thread and locks?

leekillough:
> 
> If it will be thread local is there still a need for a worker thread and locks?

Yes, that's an orthogonal issue, having to do with multiple threads (or multiple ostreams in the same thread) writing to the same files. You have to detect when multiple writers are writing to the same file, and serialize their output at "flush points" so that they are not interleaved.

The only locks are when searching for and possibly creating the worker, and when pushing task data structures onto the worker's queue. Locks are **_never_** held during IO. We do not put mutexes around `rocblas_cout << ...` or its implementation, for example. The buffers are assumed to be owned by thread, and no locks are used to access them. The buffers are independent of the files and workers.

This is a separate issue. The `rocblas_ostream` class must be `thread_local` so that each thread has its own independent **_buffer_**. The buffers have to be thread-local. The file that gets written to, however, may be shared across threads, and that's where the worker comes in. Each `rocblas_ostream` needs to be thread-specific for its own buffer space, and it won't block on IO until that buffer is flushed by either writing `std::endl` or `std::flush`, or by calling `rocblas_ostream::flush()`. However, it only blocks on `promise`s, never by holding mutexes which will block other threads.

Note that we are considering plans to move this outside of rocBLAS as a common project to be used by rocBLAS, rocSOLVER, rocFFT, and possibly others.

Another PR will be submitted soon which will make logging handle-local, replacing the legacy code which made logging a static member of `rocblas_handle`.

PR #1131, opened 2020-04-14
Make logging handle-specific

leekillough:
This makes the logging handle-specific, which, by rocBLAS rules, is thread-specific. It initializes the handle's logging during handle creation, and uses the `ROCBLAS_LAYER` environment variable's value at the time of handle creation, to determine the logging for that handle. It removes the legacy code which made logging a static member of `rocblas_handle`.

The recent change #922 had mostly to do with making the IO thread-safe in the sense of not interleaving output from multiple threads writing to the same file/device, while remaining as lock-free as possible. This change, and #1129 _make the actual data structures thread-local_, **which is an orthogonal and much simpler issue** than #922.

### DIFF
--- a/clients/include/testing_logging.hpp
+++ b/clients/include/testing_logging.hpp
@@ -80,8 +80,6 @@ void testing_logging()
     ASSERT_EQ(setenv_status, 0);
 #endif
 
-    rocblas::reinit_logs(); // reinitialize logging with newly set environment
-
     //
     // call rocBLAS functions with log_trace and log_bench to output log_trace and log_bench files
     //
@@ -329,8 +327,6 @@ void testing_logging()
 #ifdef GOOGLE_TEST
     ASSERT_EQ(setenv_status, 0);
 #endif
-
-    rocblas::reinit_logs(); // reinitialize logging, flushing old data to files
 
     //
     // write "golden file"
--- a/library/src/handle.cpp
+++ b/library/src/handle.cpp
@@ -66,6 +66,9 @@ _rocblas_handle::_rocblas_handle()
 
     // Allocate device memory
     THROW_IF_HIP_ERROR((hipMalloc)(&device_memory, device_memory_size));
+
+    // Initialize logging
+    init_logging();
 }
 
 /*******************************************************************************
@@ -82,6 +85,10 @@ _rocblas_handle::~_rocblas_handle()
     }
     if(device_memory)
         (hipFree)(device_memory);
+
+    delete log_trace_os;
+    delete log_bench_os;
+    delete log_profile_os;
 }
 
 /*******************************************************************************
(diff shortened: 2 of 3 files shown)


leekillough:
> Is there any particular reason why we're using new/delete instead of std::unique_ptr?

It would be more overhead, not any clearer to understand, and we're exception- and memory-safe as it stands now.

There are three logging pointers to `rocblas_ostream`, each initialized to `nullptr`, and later possibly changed with two different ways of calling `new`. Then they stay the same after that, throughout the lifetime of the handle.

If we used `unique_ptr`, there are many logging functions which would have to be changed to use `handle->ptr.get()` instead of just `handle->ptr`.

It's simpler and clearer the way it is now, and it's not risking memory leaks.

leekillough:
> It would be more overhead, not any clearer to understand, and we're exception- and memory-safe as it stands now.

Correction: If the constructor fails because `new` throws `std::bad_alloc`, then the previously-set pointers might leak.   We can catch and fix that. I'd rather do that than using `unique_ptr`, which would have the very same complications in a constructor unless we used initialization instead of assignment, which would be more cumbersome.

evetsso:
> Correction: If the constructor fails because `new` throws `std::bad_alloc`, then the previously-set pointers might leak. We can catch and fix that. I'd rather do that than using `unique_ptr`, which would have the very same complications in a constructor unless we used initialization instead of assignment, which would be more cumbersome.

Are you sure about that?  My observation has been that it's *really* hard to get std::unique_ptr to leak, even in this kind of situation.  I wrote a quick test program to check this and it doesn't appear to leak:

```c++
#include <stdio.h>
#include <memory>

struct A {
  A() {
    printf("ctor %p\n", this);
  }
  ~A() {
    printf("dtor %p\n", this);
  }
};

struct thrower {
  thrower() {
    second = std::make_unique<A>();
    throw std::exception();
    first = std::make_unique<A>();
  }
  std::unique_ptr<A> first;
  std::unique_ptr<A> second;
};

int main() {
  try {
    thrower t;
  }
  catch(...) {
    puts("got exception");
  }
  return 0;
}
```

leekillough:
> Are you sure about that? My observation has been that it's _really_ hard to get std::unique_ptr to leak, even in this kind of situation. I wrote a quick test program to check this and it doesn't appear to leak:

If an exception is thrown in the constructor body, then unwinding will destroy all of the non-static class members, which are assumed initialized when entering the constructor body, so if an exception occurs during an assignment (before the assignment finished), the `std::unique_ptr` should still be in a defined state, and it and all other `std::unique_ptr` members will be destroyed during unwinding.

I think what I have now (a single `try-catch` block around a bunch of allocations) is good enough, and amortizes the overhead across all 3 allocations, without requiring we go so far as to use `std::unique_ptr`, although I'm open to that.

Note that rocBLAS is a C API library, so we have to catch exceptions at the outermost C API level, and turn them into error return codes. This `rocblas_handle` constructor is an internal C++ implementation, and can't be called from C.

leekillough:
I have changed it to use `std::unique_ptr`. It looks simpler; the efficiency difference is probably a tossup.

leekillough:
CI is broken. Ran tests by hand, and they all passed:
```
[----------] Global test environment tear-down
[==========] 1071413 tests from 483 test cases ran. (19321914 ms total)
[  PASSED  ] 1071413 tests.
rocBLAS version: 2.21.0.2266-571c7559 (new Tensile client)
```

PR #1148, opened 2020-05-13
Only use _Float16 when ROCM_USE_FLOAT16 is defined, such as during rocBLAS builds

leekillough:
Resolves [SWDEV-235577](http://ontrack-internal.amd.com/browse/SWDEV-235577)

**This is just a draft; the changes still need to be tested with rocSOLVER, hipBLAS, and the new vs. old Tensile client.**


clients/benchmarks/CMakeLists.txt
```diff
@@ -173,3 +173,5 @@ set_target_properties( rocblas-bench PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJ
 add_dependencies( rocblas-bench rocblas-common )
 add_subdirectory ( ./perf_script )
 target_compile_options( rocblas-bench PRIVATE -DROCBLAS_BENCH )
+
+add_definitions( -DROCM_USE_FLOAT16 )
```
pfultz2, line 177: Shouldn't this be `target_compile_definitions(rocblas-bench PRIVATE -DROCM_USE_FLOAT16)`?

pfultz2:
It would be better to add the flags to each target that needs it with `target_compile_definitions(<cmake-target> PRIVATE -DROCM_USE_FLOAT16)` instead of using directory level `add_definitions`. It makes clearer to see which targets explicitly are using that flag.

PR #1150, opened 2020-05-23
Remove all instances of hcc in rocBLAS

amdkila:
- Removes hcc in pre-commit and header_compilation tests
- Removes hcc from install script
- Sets default TENSILE_CODE_OBJECT_VERSION to V3 and TENSILE_COMPILER to hipcc
- Removes all paths with hcc in the CMake files
- Only will be done in master-rocm-3.5 branch, thus the direct PR to main rocBLAS repo

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -8,7 +8,7 @@ export PATH=/usr/bin:/bin
 
 set -x
 
-format=/opt/rocm/hcc/bin/clang-format
+format=/opt/rocm/llvm/bin/clang-format
 
 # Redirect stdout to stderr.
 exec >&2
--- a/header_compilation_tests.sh
+++ b/header_compilation_tests.sh
@@ -44,11 +44,11 @@ if ! [ -z ${ROCM_PATH+x} ]; then
     rocm_path=${ROCM_PATH}
 fi
 
-HCC=${rocm_path}/hcc/bin/hcc
+HIPCC=${rocm_path}/hip/bin/hipcc
 
-HCC_OPTS="-Werror -DBUILD_WITH_TENSILE=1 -DTensile_RUNTIME_LANGUAGE_HIP=1 -DTensile_RUNTIME_LANGUAGE_OCL=0 -Drocblas_EXPORTS -I$(realpath library/include) -I$(realpath library/src/include) -I$(realpath $BUILD_DIR/include) -I$(realpath $SOURCE_DIR/library/src/blas3/Tensile) -isystem ${rocm_path}/hip/include -isystem ${rocm_path}/hsa/include -isystem ${rocm_path}/hcc/include -isystem ${rocm_path}/include -I$(realpath $BUILD_DIR/Tensile) -O3 -DNDEBUG -fPIC"
+HIPCC_OPTS="-Werror -DBUILD_WITH_TENSILE=1 -DTensile_RUNTIME_LANGUAGE_HIP=1 -DTensile_RUNTIME_LANGUAGE_OCL=0 -Drocblas_EXPORTS -I$(realpath library/include) -I$(realpath library/src/include) -I$(realpath $BUILD_DIR/include) -I$(realpath $SOURCE_DIR/library/src/blas3/Tensile) -isystem ${rocm_path}/hip/include -isystem ${rocm_path}/include -I$(realpath $BUILD_DIR/Tensile) -O3 -DNDEBUG -fPIC"
 
-GPU_OPTS="-Wno-unused-command-line-argument -fvisibility=hidden -fvisibility-inlines-hidden -hc -fno-gpu-rdc --amdgpu-target=gfx803 --amdgpu-target=gfx900 --amdgpu-target=gfx906 -Werror"
+GPU_OPTS="-Wno-unused-command-line-argument -fvisibility=hidden -fvisibility-inlines-hidden -fno-gpu-rdc --amdgpu-target=gfx803 --amdgpu-target=gfx900 --amdgpu-target=gfx906 -Werror"
 
 CLANG=${rocm_path}/llvm/bin/clang
 CLANG_OPTS="-xc-header -std=c99"  # auto set in hip_common.h -D__HIP_PLATFORM_HCC__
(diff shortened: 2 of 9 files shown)


amdkila:
> This is bigger than what I suggested. More reviewers needed.
Yep. No rush to merge this one in. I added Lee specifically because of the compilation_tests being changed. It works though...at least when I tried it locally it did.

library/src/CMakeLists.txt
```diff
@@ -80,20 +80,11 @@ if( BUILD_WITH_TENSILE )
     target_link_libraries( Tensile PRIVATE hip::device )
   endif()
 
-  if( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
-    # Remove following when hcc is fixed; hcc emits following spurious warning ROCm v1.6.1
-    # "clang-5.0: warning: argument unused during compilation: '-isystem /opt/rocm/include'"
-    # target_compile_options( Tensile PRIVATE -Wno-unused-command-line-argument -fno-gpu-rdc -Wno-deprecated-declarations)
+  if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
```
TorreZuk, line 83: This was a breaking change alone so expect you may break 3.5 on some gpus with this.

amdkila: That wouldn't surprise me. I really need CI to run this. Haven't even tested this on other distros

saadrahim:
My request was just install.sh. That change is harmless and any user can go back with one line change. Let's get that in only.

I prefer no hcc option in install.sh but I accept rocblas owners may want to keep it is an option.

TorreZuk:
Sure a simple option hide is safe late in the game.   The full PR could just re- target develop but not sure if we'll try and use the new find_package OpenMP in 3.6 which could cause conflicts so may want to wait until after 3.6 code freeze.   Anyway it's not my call.

amdkila:
I'm operating under the assumption that this sort of change would not make it into develop at the moment. It may be someone that should only be in the release branch, at least until it's clear that hcc is never to be used. The timing isn't great. I can always make the original PR as Saad suggested

leekillough:
> > This is bigger than what I suggested. More reviewers needed.
> 
> Yep. No rush to merge this one in. I added Lee specifically because of the compilation_tests being changed. It works though...at least when I tried it locally it did. 

`header_compilation_tests.sh` is disabled or greatly handicapped for hip-clang. I have not ported it to hip-clang because it hasn't stabilized enough. It would have caught the bug which required [PR#85](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/85)

leekillough:
> I'm operating under the assumption that this sort of change would not make it into develop at the moment. It may be someone that should only be in the release branch, at least until it's clear that hcc is never to be used. The timing isn't great. I can always make the original PR as Saad suggested

Why not just make hip-clang the default in `install.sh`, which should have already been the default?

leekillough:
> > > This is bigger than what I suggested. More reviewers needed.
> > 
> > 
> > Yep. No rush to merge this one in. I added Lee specifically because of the compilation_tests being changed. It works though...at least when I tried it locally it did.
> 
> `header_compilation_tests.sh` is disabled or greatly handicapped for hip-clang. I have not ported it to hip-clang because it hasn't stabilized enough. It would have caught the bug which required [PR#85](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/85)

In other words, you shouldn't need to remove HCC from it until you fully port it to work with hip-clang, because it shouldn't be calling HCC when hip-clang is built, such as by not being called at all (or an early exit at the top of the script).

Unless you port it to work 100% with hip-clang, I would leave it alone, since it is disabled or handicapped with hip-clang.

.githooks/pre-commit
```diff
@@ -8,7 +8,7 @@ export PATH=/usr/bin:/bin
 
 set -x
 
-format=/opt/rocm/hcc/bin/clang-format
```
leekillough, line 11: This change will be needed in hipBLAS and Tensile too.

install.sh
```diff
@@ -22,7 +22,7 @@ rocBLAS build & installation helper script
       -o | --cov                 Set Tensile code_object_version (V2 or V3)
       -t | --test_local_path     Use a local path for Tensile instead of remote GIT repo
            --cpu_ref_lib         Specify library to use for CPU reference code in testing (blis or lapack)
-           --[no-]hip-clang      Whether to build library for amdgpu backend using hip-clang
+           --hip-clang           Whether to build library for amdgpu backend using hip-clang
```
leekillough, line 25: I would remove `--hip-clang` entirely from the flags and just accept and ignore or give error for old options. Removing the `--no-hip-clang` option while leaving `--hip-clang` makes no sense.

leekillough:
The`header_compilation_tests.sh` is disabled for hip-clang here:

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/0155bcd67cd4587f78ca1774509f387de50cc7e9/install.sh#L558

It should not be modified until it is ported to hip-clang. The HCC commands have contextual information which must be preserved in the port, and shouldn't be removed from the script now.

install.sh
saadrahim: Devops requested that we support the option temporarily to give them time to upgrade without failed builds.

saadrahim: It should be cleaned up after they transition.

leekillough: Support both options until you remove both entirely. Don't remove `--no-hip-clang` while leaving `--hip-clang`. 

Change the default all you wish, but allow the user to specify one or the other explicitly until you remove the options entirely.

You can remove the options from the documentation if you remove both the positive and negative sense options, even if you still support them as hidden options. But listing one without the other is misleading.

header_compilation_tests.sh
```diff
@@ -44,21 +44,21 @@ if ! [ -z ${ROCM_PATH+x} ]; then
     rocm_path=${ROCM_PATH}
```
leekillough, line 44: This file is [not used in hip-clang builds](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/0155bcd67cd4587f78ca1774509f387de50cc7e9/install.sh#L556), and should not be modified until it is completely ported to hip-clang, which is non-trivial. The existing commands have contextual information which must be translated into hip-clang equivalents. HCC should not be removed, but should modified to hip-clang. I reject the changes in this file.

PR #1154, opened 2020-06-19
use python3 

TorreZuk:
* python3 as Centos8 may not have python without alternatives


TorreZuk:
It is run as python3 for tensile and we have python3 as a dependency.

TorreZuk:
I said they could just install python2 as python using alternatives but Aditya chose this.

pruthvistony:
The changes look good to me.

adilad:
Will this work for Centos7 ?

leekillough:
On modern Pythons versions, `python` should be a symlink to `python3`. 

I don't see any problems with this change, unless certain platforms require `python` to be used instead of `python3`.

TorreZuk:
>This works on older OSes?
It is run as python3 for tensile and we have python3 as a dependency.
(implied it should work on older ones if tensile does)
> Will this work for Centos7 ?
Passed on our Centos7.

PR #1159, opened 2020-08-07
SWDEV-246520-hide-symbols

TorreZuk:
* hides llvm symbols used by embedded rocblas-tensile library


sunway513:
@deven-amd has ran the tests in TF and confirmed this PR would fix the issues we were facing in SWDEV-246520. 
Thanks @bragadeesh @TorreZuk !

PR #1161, opened 2020-08-11
Fix broken kernel that causes assertion failure in SWDEV-245842 

amdkila:


amdkila:
Heard from QA that it is not necessary to include in rocm-3.7.x, so I will close this PR

PR #1164, opened 2020-08-18
support clang-12, update clang header path

scchan:
update clang header path in cmake


scchan:
@amcamd  done, moved this PR to the internal repo.

PR #1165, opened 2020-09-15
Updates to debian package name

pruthvistony:
- According to Debian standards, package name and package
  version should be separated by '_'.

### DIFF
--- a/docker/dockerfile-install-ubuntu
+++ b/docker/dockerfile-install-ubuntu
@@ -9,7 +9,7 @@ COPY *.deb /tmp/
 
 # Install the debian package, and print out contents of expected changed locations
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends --allow-unauthenticated -y \
-    /tmp/rocblas-*.deb \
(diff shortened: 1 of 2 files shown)


saadrahim:
I am not sure how this is changing the package name. The package name is not generated in these files modified.

pruthvistony:
Needs to be raised on rocBLAS-internal, closing this.

cgmb:
Just to expand on Saad's comment, the packages are built with CPack. The use of dashes is [the default CPack behaviour](https://cmake.org/cmake/help/v3.18/module/CPack.html#variable:CPACK_PACKAGE_FILE_NAME). This change could be done globally for all ROCm packages in [rocm-cmake](https://github.com/RadeonOpenCompute/rocm-cmake/blob/master/share/rocm/cmake/ROCMCreatePackage.cmake).

As far as know, the Debian standards are mostly about whether the package can be included in the Debian repositories, and I'm not sure CPack can be used for that purpose anyway. If you know of any existing packages in the Debian main, contrib, or non-free repositories that use CPack, I would love to learn about them.

PR #1171, opened 2020-10-22
[Hotfix] SWDEV-254253

leekillough:
Fix for [SWDEV-254253](http://ontrack-internal.amd.com/browse/SWDEV-254253)

Changes:
- Removes synchronization from `rocblas_set_stream()`
- Documents that it's the user's responsibility to call `hipStreamSynchronize()` before calling `rocblas_set_stream()` and before calling `hipStreamDestroy()` when switching between non-default streams
- Code quality improvements for stream (use accessors and friend function instead of public data member)


amcamd (approved):
LGTM

docs/source/handle_stream_device.rst
```diff
@@ -51,6 +51,31 @@ you have completed using it with:
...
+If you change the stream from one non-default stream to another non-default stream, it
+is your responsibility to synchronize the old stream before setting the new stream, and
+before optionally destroying the old stream:
+
+::
+
+    // Synchronize the old stream
+    if(hipStreamSynchronize(stream) != hipSuccess) return EXIT_FAILURE;
+
+    // Destroy the old stream
```
bragadeesh, line 63: There are a sequence of 4 steps mentioned here and 2 of that are left to user control. Please clarify saying the destruction of old stream and creation of new stream are left to user control. The only requirement is user synchronizes old stream and creates new stream before the call to rocblas_set_stream

Please note that user application might have multiple threads and may manage those things outside of 1 thread. We certainly are not suggesting users to follow the 4-step sequence in 1 thread as mentioned here.

leekillough: > There are a sequence of 4 steps mentioned here and 2 of that are left to user control. 

All 4 steps are left to user control.

bragadeesh (approved):

PR #1178, opened 2020-12-09
Increasing test timeout to 6 hours

saadrahim:


TorreZuk (approved):
Test reduction PR will be made today, but may as well bump it up.

PR #1181, opened 2020-12-18
adding patch to strip out xnack/ecc from gcnArch name

yoichiyoshida:
adding patch to maintain previous behavior following hip gcnArchName change


zaliu:
@yoichiyoshida  Can you please target the internal repo?

PR #1188, opened 2021-04-08
remove division by number_hot_calls when using ArgumentModel

amcamd:
- resolves SWDEV-280617
- The function ArgumentModel divides gpu_time by number_hot_calls so the calling code should not do this division.
- This fix is for the three functions: syr_strided_batched, gemm_strided_batched, rot_strided_batched


TorreZuk (approved):
LGTM

PR #1189, opened 2021-05-03
Use CMAKE_CXX_COMPILER_ID to detect Clang

scchan:
We are adding the vendor string "AMD" to ROCm's hip-clang as a consequence, the regex to detect clang from its version string would be broken.  Instead of doing that, check whether the cmake's built-in CMAKE_CXX_COMPILER_ID variable is defined as Clang.


amcamd (changes requested):
Is this a hot-fix for ROCm 4.3? If this is true, it is submitted to rocBLAS/master, else it should be submitted to rocBLAS-internal/develop.

scchan:
replaced by https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/1190

PR #1203, opened 2021-08-31
Update clients.rst for external rocBLAS github link

paklui:
externally facing doc, so change the link from:
https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/docs/source/install.rst
to:
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/docs/source/install.rst

resolves #___

Summary of proposed changes:
-
-
-

### DIFF
--- a/docs/source/clients.rst
+++ b/docs/source/clients.rst
@@ -8,7 +8,7 @@ There are two client executables that can be used with rocBLAS. They are,
 
 2. rocblas-test
 
-These two clients can be built by following the instructions at `Building and Installing rocBLAS github page <https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/docs/source/install.rst>`_. After building the rocBLAS clients, they can be found in the directory ``rocBLAS/build/release/clients/staging``.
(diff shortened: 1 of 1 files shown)


amcamd (approved):

zaliu (commented):
Wouldn't this change break our sync flow from rocBLAS-internal to rocBLAS?

amcamd:
It would be simpler it it went to rocBLAS-internal/master, but I think if we merge rocBLAS/master into rocBLAS-internal/master it will not break the flow, because rocBLAS-internal/master will now be ahead of rocBLAS/master. We will also need to merge or cherry pick rocBLAS-internal/master back into rocBLAS-internal/develop.

paklui:
> It would be simpler it it went to rocBLAS-internal/master, but I think if we merge rocBLAS/master into rocBLAS-internal/master it will not break the flow, because rocBLAS-internal/master will now be ahead of rocBLAS/master. We will also need to merge or cherry pick rocBLAS-internal/master back into rocBLAS-internal/develop.

Ok, maybe we don't need this change here. Please feel free to make the change in rocBLAS-internal/master. I can close this then

PR #1205, opened 2021-09-13
only add targets if building coverage

TorreZuk:
* fixes building with --clients-only


TorreZuk:
Closed as can be fixed with a patch file.

PR #1206, opened 2021-09-13
Remove unsafe emacs script.

malcolmroberts:


amcamd:
Can you please close this PR and submit it to rocBLAS-internal. Can you also please add an explanation on why the emacs script is unsafe.

PR #1236, opened 2022-03-11
update CHANGELOG.md for ROCm 5.1

amcamd:


TorreZuk (approved):
IMO this is an awkward procedure they should have created an automated system for it.

PR #1239, opened 2022-04-01
Fix typo and specify repo

ckime-amd:
Add link to repo and suggest formatting change for branches.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -6,10 +6,10 @@ Pull-request guidelines
 =======================
 
 
-Our code contriubtion guidelines closely follows the model of `GitHub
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
Looks good to me.  Thanks.

TorreZuk:
Note this will actually either have to be redirected to internal repo rocBLAS-internal so not sure if you can do that @ckime-amd .   @amcamd will advise.  But if we close this don't be surprised and it will be replicated internally.

amcamd (approved):
- Thanks for these corrections. 
- Can you go ahead and merge this PR. 
- For PRs from AMD employees that have access to https://github.com/ROCmSoftwarePlatform/rocBLAS-internal it is better to direct PRs to rocBLAS-internal.

PR #1268, opened 2022-09-26
Hotfix to hide hipMallocAsync/hipFreeAsync on ROCm 5.2 and earlier.

rkamd:
Hotfix to hide hipMallocAsync/hipFreeAsync on ROCm 5.2 and earlier.


TorreZuk (approved):
looks the same to me as first hotfix, so fine

PR #1279, opened 2022-10-31
updating i8ii_NN/TN lib logics

babakpst:
After my latest lib logic update for I8II_NN/TN, the performance of some sizes dropped, because these sizes are using the new kernels which are not optimized for these sizes. 
This PR adds new kernels to the lib logic to improve the performance of those sizes.


benjaminulmer (approved):
Logic files look good

TorreZuk (approved):
LGTM but you need hotfix approval from release management

PR #1284, opened 2022-12-19
aldebaran BBS tuning

benjaminulmer:
Tuning for SWDEV-372453

All sizes are new and there are some new kernels as well.


babakpst (approved):
I assume you are using the re-tuning script for this PR. Ideally, we should not see any new kernels in the diff (I need to update merge.py to not adds duplicates). The performances of new sizes are low. Please merge only if this PR solves the issue.

TorreZuk (approved):
same precheckin HMM failures as baseline, LGTM if extended and staging testing pass

benjaminulmer:
> I assume you are using the re-tuning script for this PR

It's a combination a retuning and regular tuning. All the kernels added are true new kernels

babakpst:
This PR is ready to merge. The failed tests are unrelated to these changes.

nielenventer:
This patch causes a regression for large GEMMs (https://ontrack-internal.amd.com/browse/SWDEV-375718), the reason seems to be that there are no tuning points for larger M values (this one stops at 16), and so some large GEMMs are picking the kernels for the M=16 case (v small tile size, perf regression). I'm doing some extra tuning for larger M and will update this commit.

nielenventer:
I added some new exact sizes, with large M, following the pattern of the other tunings in this commit. I confirmed it fixes the regression.

Only the Retune tool was used, but unfortunately new kernels are added due to known issue with merge script.

PR #1290, opened 2023-02-02
Hotfix : Update CMakeLists.txt to link libpthread

rkamd:
Update CMakeLists.txt to link libpthread for all rocblas samples programs to avoid build error.

### DIFF
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -69,7 +69,7 @@ foreach( exe ${sample_list_fortran} )
 endforeach( )
 
 foreach( exe ${sample_list_all} )
-  target_link_libraries( ${exe} PRIVATE roc::rocblas )
+  target_link_libraries( ${exe} PRIVATE Threads::Threads roc::rocblas )
 
   set_target_properties( ${exe} PROPERTIES
     CXX_STANDARD 14
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -69,7 +69,7 @@ foreach( exe ${sample_list_fortran} )
 endforeach( )
 
 foreach( exe ${sample_list_all} )
-  target_link_libraries( ${exe} PRIVATE Threads::Threads roc::rocblas )
+  target_link_libraries( ${exe} PRIVATE roc::rocblas )
 
   set_target_properties( ${exe} PROPERTIES
     CXX_STANDARD 14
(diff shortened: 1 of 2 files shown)


amcamd (approved):

TorreZuk (approved):
Hopefully the guard isn't required as Threads::Threads target didn't exist with some old OS.  Fine for now but would also tend to put more general library last in the list, i.e. rocblas before Threads::Threads.

cgmb (changes requested):
This doesn't feel right to me. The error that prompted this was:
```
ld.lld: error: undefined reference due to --no-allow-shlib-undefined: pthread_create
>>> referenced by ../../library/src/librocblas.so.0.1.50500
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [clients/samples/CMakeFiles/rocblas-example-c-dgeam.dir/build.make:100: clients/staging/rocblas-example-c-dgeam] Error 1
```
The error is saying that librocblas depends on pthreads. Yes, linking all the samples against pthreads will prevent the samples from failing to link, but my interpretation of this error is that other programs written by rocblas users will be seeing the same problem as the samples are. This is a problem in the rocblas library, not a problem in the samples.

If you look at [library/src/CMakeLists.txt](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/9d6f87a9e25daa688a020ca030ed4c6b991b2ef1/library/src/CMakeLists.txt#L44-L45), you can see that the rocBLAS library build code does:
```cmake
set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)
```
but it never actually uses the package that it finds. I believe that if the rocblas library itself was linked to pthreads, the usage requirement would be propagated to the samples automatically.

TorreZuk:
Ruban can you try what @cgmb suggested, as it may have gotten lost sometime in the past as we were getting the -lpthreads for free from hipcc.     Just add it to the library PRIVATE Threads::Threads as we use it for logging so it may as well be added now.   Not sure what you saw from ldd for it in your test but the library must have linked against it from some dependency?

rkamd:
@cgmb , Linked rocblas library itself to pthreads library.

cgmb (approved):
Thanks. Assuming you've tested this and verified that it prevents the build failure, I think that is a better solution.

rkamd:
> Thanks. Assuming you've tested this and verified that it prevents the build failure, I think that is a better solution.

Yes, this resolves build error.

PR #1292, opened 2023-02-07
Hotfix: Remove hipStreamSynchronize from rotg

rkamd:
- Remove hipStreamSynchronize from rotg functions as they are not needed. 
- Resolves HIP graph capture pre_checkin test failures.


TorreZuk (approved):
LGTM

PR #1335, opened 2023-06-30
SWDEV-398093 - Exclude Tensile library files from ASAN package (#1924)

raramakr:
ASAN package requires only address sanitized shared libraries and license files. Tensile library files are not shared libraries, so excluding them from ASAN package

resolves #___

Summary of proposed changes:
-
-
-


TorreZuk (approved):
Looks to match what I approved for develop, but should get @amcamd to approve all hot-fixes as per our process (and wait for tests to pass).

amcamd (approved):
As mentioned by @TorreZuk , please wait for CI tests to pass before merging.

PR #1351, opened 2023-08-09
Fix rmake for windows

SzymonOzog:
It previously crashed with:
UnboundLocalError: local variable 'all_args' referenced before assignment

### DIFF
--- a/rmake.py
+++ b/rmake.py
@@ -208,7 +208,7 @@ def fatal(msg, code=1):
 def deps_cmd():
     if os.name == "nt":
         exe = f"python3 rdeps.py"
-        cmdline_args = ""
+        all_args = ""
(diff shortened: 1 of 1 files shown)


TorreZuk:
@SzymonOzog thanks for contributing this fix.    This is fixed already in branch release/rocm-rel-5.6 and for now we can advise others on release/rocm-rel-5.5 branch to install dependencies on windows directly with python rdeps.py and bypass the rmake.py -d flag.   For the time being we have frozen this branch so we may close the PR and reopen it when we can update the branch again.

TorreZuk:
Closing for now as rocm-rel-5.5 branch updates are not currently an option.   May reopen this PR later.

PR #1383, opened 2024-01-06
Update readme for Fedora

trixirt:
I package rocBLAS on Fedora, here is how you install it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,10 @@ from the system package manager. For example, on Ubuntu you can use the followin
 sudo apt-get update
 sudo apt-get install rocblas
 ```
+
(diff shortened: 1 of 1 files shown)


amcamd (changes requested):
Could you please move this information to the file: https://github.com/ROCm/rocBLAS/blob/develop/docs/Linux_Install_Guide.rst 

The information in README.md is duplicate information and I will remove it. The correct place for this information is in the above file.

amcamd:
@trixirt Thank you for the instructions to install Fedora packages. I merged your commit then moved your instructions from README.md to docs/Linux_Install_Guide.rst.

PR #1387, opened 2024-01-25
fix thread-safety on tensile path detection (#2291)

TorreZuk:
* cherry pick fix from develop for random detection failures of the tensile library location (multi-thread init only)
* random failures can also be avoided with environment variable ROCBLAS_TENSILE_LIBPATH


amcamd (approved):

TorreZuk:
All failures are isNan on host for which alternative reference code is in develop.

PR #1403, opened 2024-03-08
Hotfix: tune aquavanjaram942X BBS+HHS NN datatype

amgddm:
resolves #LWPTENSILE-891

Summary of proposed changes:
- MatrixInstruction: 32x32x8
- LRVW8, GRVW8
- ClusterLocalRead: [True]
- VgprForLocalReadPacking: [True]


babakpst (approved):

nakajee:
Please do not use DoWhile.
It is useless for asm kernel and it causes issues without specifying assertion.

PR #1407, opened 2024-03-15
Hotfix reference lib 6.1

TorreZuk:
* avoids reference code rare and random host ONLY NaNs on some gemms resulting in test failures:
Value of: rocblas_isnan(hGPU[i + idx])
  Actual: false
Expected: true
* no change to library


amcamd (approved):
LGTM

TorreZuk:
No reports of need so closing PR for now as in 6.2 branches

PR #1411, opened 2024-03-25
Hotfix: Remove passing the device set to lazy load API

rkamd:
Remove passing the list of devices to Lazy Loading API, due to following reasons,

 - Increases the initilaization time.
 - Resolves [SWDEV-438929](https://ontrack-internal.amd.com/browse/SWDEV-438929) - hipModuleNotFound error ( this is a workaround and root cause is still unknown)

Reverting this change might affect heterogeneous GPU support.


mahmoodw (approved):
LGTM - tested and working as a workaround to the ticket

PR #1431, opened 2024-05-10
update version patch field for ROCm 6.1.2

amcamd:


TorreZuk (approved):
Sure, I would have gone with 4.1.2 to match the expected tag but fine as that.

vamovsik:
Fix for https://ontrack-internal.amd.com/browse/SWDEV-461882

PR #1432, opened 2024-05-14
Hotfix: rocBLAS failure during stream capture

rkamd:
resolves #SWDEV-454629

Summary of proposed changes:
Fix error during graph capture. A recent change in HIP runtime to not support hipStreamQuery() during stream capture ( matches CUDA behavior) causes rocblas_set_stream() to return error 'hipErrorStreamCaptureUnsupported' during stream capture.

- This PR checks if the current stream is in capture mode, and skips the use of hipStreamQuery().
- Fixed return of hipStreamQuery()

### DIFF
--- a/library/src/rocblas_auxiliary.cpp
+++ b/library/src/rocblas_auxiliary.cpp
@@ -246,9 +246,14 @@ try
     if(stream == handle->stream)
         return rocblas_status_success;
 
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
LGTM

amcamd (approved):

PR #1434, opened 2024-05-30
Update proper install component for tensile data of static builds

arvindcheru:
resolves #___

Summary of proposed changes:
- Update install component for Tensile data in static builds 
- the component install component updated to clients-common.
-

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -802,7 +802,7 @@ if( BUILD_WITH_TENSILE )
     if( BUILD_SHARED_LIBS )
       set( TENSILE_DATA_COMPONENT_NAME ${CMAKE_INSTALL_DEFAULT_COMPONENT_NAME} )
     else()
-      set( TENSILE_DATA_COMPONENT_NAME clients-common )
+      set( TENSILE_DATA_COMPONENT_NAME devel )
     endif()
     rocm_install(
       DIRECTORY ${CMAKE_BINARY_DIR}/Tensile/library
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -799,10 +799,15 @@ if( BUILD_WITH_TENSILE )
   endif()
   # For ASAN package, Tensile library files(which are not shared libraries) are not required
   if( NOT ENABLE_ASAN_PACKAGING )
+    if( BUILD_SHARED_LIBS )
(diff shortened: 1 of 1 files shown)


library/src/CMakeLists.txt
```diff
@@ -799,10 +799,15 @@ if( BUILD_WITH_TENSILE )
   endif()
   # For ASAN package, Tensile library files(which are not shared libraries) are not required
   if( NOT ENABLE_ASAN_PACKAGING )
+    if( BUILD_SHARED_LIBS )
+      set( TENSILE_DATA_COMPONENT_NAME ${CMAKE_INSTALL_DEFAULT_COMPONENT_NAME} )
+    else()
+      set( TENSILE_DATA_COMPONENT_NAME clients-common )
+    endif()
```
TorreZuk: So I am not sure of this convention.  Clients are users of the library.   The library runtime depends on TensileData.  Downstream consumers of the library don't depend on clients-common ?

TorreZuk: @evetsso and @lawruble13 please chime in.

lawruble13: This had been my initial recommendation, but on discussion with you folks I now agree that this isn't the way to approach it. We need a separate package to contain the Tensile files in a static context.

TorreZuk:
Okay but as this is not required for development it is only required for runtime, this also seems worse than the current CMAKE_INSTALL_DEFAULT_COMPONENT_NAME (repeated as github still bad for commit comments)

library/src/CMakeLists.txt
evetsso: Does this component name get turned into a package name when we start making packages?

TorreZuk (approved):
Fine until we can think of something better, maybe cherry pick to rel staging 6.2 to get CI to run as that is where it needs to land.  I can cherry pick back to develop for you after all that. Get another rocblas approval too please.

amcamd:
As Torre requested, please make sure to cherry-pick this to to rocBLAS:release-staging/rocm-rel-6.2 where it will be tested by other groups.

amcamd (approved):

TorreZuk:
@lawruble13 and @evetsso do you accept the devel package location for these runtime files?

evetsso:
I don't have a better suggestion at this time.

PR #1438, opened 2024-06-04
Hotfix: Update rocblas_set_stream() to check the capture status of new stream 

rkamd:
Hotfix: Update rocblas_set_stream() to check the capture status of new stream. If the new stream is in capture mode, then skip calling hipStreamQuery()


TorreZuk (approved):
LGTM

PR #1442, opened 2024-06-07
Remove the apostrophe in package description (#2559)

raramakr:
* Remove the apostrophe in package description
* While creating wheel package, the rpm tags are read from the rpm package.  The apostrophe in the package description is causing syntax error while parsing the description tag.

resolves #___

Summary of proposed changes:
-
-
-


TorreZuk (approved):
LGTM

PR #1447, opened 2024-06-20
Hotfix: Use hipStreamQuery only if the stream capture status is `hipStreamCaptureStatusNone`  

rkamd:
resolves # [SWDEV-466097](https://ontrack-internal.amd.com/browse/SWDEV-466097) & [SWDEV-464655](https://ontrack-internal.amd.com/browse/SWDEV-464655)

Checking whether Stream capture is active ( hipStreamCaptureStatusActive ) might not be enough, as there could be a status hipStreamCaptureStatusInvalidated which indicates the stream capture is invalidated but not terminated. This PR addresses this issue by calling hipStreamQuery() only if the stream capture status is `hipStreamCaptureStatusNone`

### DIFF
--- a/library/src/include/handle.hpp
+++ b/library/src/include/handle.hpp
@@ -366,19 +366,18 @@ public:
         //default stream will not be in capture mode
         if(stream != 0)
         {
-            bool status = hipStreamIsCapturing(stream, &capture_status) == hipSuccess;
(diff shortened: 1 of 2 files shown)


TorreZuk (approved):
LGTM

PR #1475, opened 2024-09-03
Reuse device memory in testing_gemm_ex

pghysels:
Reuse device memory allocation in testing_gemm_ex. 
Based on code originally by @bethune-bryant 

This drastically speeds up testing of many consecutive calls to gemm_ex.
For a run with 50K different gemms, I observed a ~23x speedup.


TorreZuk:
Requested move to internal.

PR #1482, opened 2024-09-17
Update tensile_tag.txt for venv fix

TorreZuk:
* tensile venv activation fix


davidd-amd (approved):
Confirmed that this is the correct commit on the analogous branch in Tensile. LGTM.

amcamd (approved):

PR #1510, opened 2024-10-25
[DO NOT SUBMIT/TESTING ONLY] Change tensile tag to topic/drop-hipcc-perl

kzhuravl:
resolves #___

Summary of proposed changes:
-
-
-


TorreZuk:
I think you can close this PR, tip commit of develop now uses tensile tag without perl hipcc.

TorreZuk:
Test PR closing as it should not be required now even for test.  Perl based hipcc should be bypassed in Tensile and included by develop rocblas tag.

PR #1513, opened 2024-11-01
cherry pick commit to remove gfx940 and gfx941 from fat binary for ROCm 6.3 

amcamd:
remove gfx940 and gfx941 from fat binary for ROCm 6.3


TorreZuk (approved):
Looks good

PR #1520, opened 2024-11-18
improved heuristic due to stress subset testing (#2769)

TorreZuk:
* only remove stress tests when either no filter or only removing known_bugs

(cherry picked from commit 47f9eb9b73ec92d1551f86c3cd6032bbb9e986ef)


daineAMD (approved):
I approved the original PR so may want another review to be sure. But here's my approval again anyway.

PR #1537, opened 2025-01-08
clients: when blas is not avialble at hardcoded paths, try pkg-config on linux to find a reference blas implementation

IMbackK:
Allows clients to build on linux systems where a different blas implementation is used or blis is not installed in one of the hardcoded places and pkg-config is available.


TorreZuk (commented):
Changes look fine to me but we may have to bring changes internally before they would reappear in develop.  We will discuss.

amcamd:
@IMbackK ,  Thanks for your contribution, PR is merged.

PR #1552, opened 2025-02-25
Cherry pick tensile tag with compression

bstefanuk:
Updates tensile tag for ROCm 6.4 such that code object compression is enabled.


rkamd (approved):

TorreZuk:
Looks like it fails to build on windows

PR #1570, opened 2025-04-07
Fallback to rocBLAS GEMM for complex types when others fail.

rkamd:
resolves # LWPMLSE-844

Summary of proposed changes:
- Fallback to rocBLAS GEMM for complex types when other libraries (hipBLASLt or Tensile) fail.

### DIFF
--- a/library/src/blas3/Tensile/gemm_templates.cpp
+++ b/library/src/blas3/Tensile/gemm_templates.cpp
@@ -135,8 +135,7 @@ rocblas_status rocblas_internal_gemm(rocblas_handle    handle,
     // Return the current status if a solution was not found
     // by other libraries(hipBLASLt or Tensile) or if the data is not both complex.
     // Otherwise, fall through to rocblas source GEMM.
-    if(status != rocblas_status_not_implemented
-       && !(rocblas_is_complex<TConstPtr> && rocblas_is_complex<TPtr>))
+    if(status != rocblas_status_not_implemented && !rocblas_is_complex<TScal>)
     {
         return status;
     }
--- a/library/src/blas3/Tensile/gemm_templates.cpp
+++ b/library/src/blas3/Tensile/gemm_templates.cpp
@@ -135,7 +135,7 @@ rocblas_status rocblas_internal_gemm(rocblas_handle    handle,
     // Return the current status if a solution was not found
     // by other libraries(hipBLASLt or Tensile) or if the data is not both complex.
     // Otherwise, fall through to rocblas source GEMM.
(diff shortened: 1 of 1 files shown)


library/src/blas3/Tensile/gemm_templates.cpp
```diff
@@ -76,61 +77,71 @@ rocblas_status rocblas_internal_gemm(rocblas_handle    handle,
...
+                                      batch_count);
+    }
+
+    // Return the current status if a solution was not found
+    // by other libraries(hipBLASLt or Tensile) or if the data is not both complex.
+    // Otherwise, fall through to rocblas source GEMM.
+    if(status != rocblas_status_not_implemented
+       && !(rocblas_is_complex<TConstPtr> && rocblas_is_complex<TPtr>))
+    {
+        return status;
```
TorreZuk: Are you seeing complex source backend with the LAYER 8?    I would have thought these pointers need base type extraction?   If these aren't mixed real/complex scalar/data API then TScal maybe would suffice?

rkamd: Latest commit should have the change to check TScal type. Also added a debug code to fall through for cgemm /zgemm & LAYER8 confirms source backend.

TorreZuk (approved):
I guess source fallback for non complex as well could be option.  @amcamd ?

library/src/blas3/Tensile/gemm_templates.cpp
```diff
@@ -76,61 +77,70 @@ rocblas_status rocblas_internal_gemm(rocblas_handle    handle,
...
+                                      k,
+                                      batch_count);
+    }
+
+    // Return the current status if a solution was not found
+    // by other libraries(hipBLASLt or Tensile) or if the data is not both complex.
+    // Otherwise, fall through to rocblas source GEMM.
+    if(status != rocblas_status_not_implemented || !rocblas_is_complex<TScal>)
+    {
+        return status;
```
amcamd, line 140: Is the second logic condition  `!rocblas_is_complex<TScal>)` required? What is `rocblas_status` return from `rocblas_call_tensile` if it is called with `rocblas_is_complex<TScal>`?

amcamd (changes requested):

library/src/blas3/Tensile/gemm_templates.cpp
amcamd: I recall the comment, I see the fall-through is only for complex.

amcamd (approved):

PR #1575, opened 2025-04-09
stress tests fix gemm bad strides and misc cleanup

TorreZuk:
* stress related test cleanup

### DIFF
--- a/clients/gtest/gemm_strided_batched_gtest.yaml
+++ b/clients/gtest/gemm_strided_batched_gtest.yaml
@@ -938,7 +938,6 @@ Tests:
     - { M:  63,  N:  512 }
   pointer_mode_host: false
   batch_count: 1
-  stride_scale: [ 1 ]
 
 - name: gemm_ex_large_batch
   category: stress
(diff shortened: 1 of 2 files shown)


clients/gtest/gemm_strided_batched_gtest.yaml
```diff
@@ -975,7 +977,7 @@ Tests:
     - gemm_strided_batched_ex: *nonint8_real_precisions
     - gemm_strided_batched_ex: *int8_precision
   arguments:
-    - { M: 4, N: 4, K: 4, lda: 4, ldb: 4, ldc: 4, stride_a: 4, stride_b: 4, stride_c: 4, stride_d: 4 }
```
daineAMD, line 978: I think you'll have to change rocblas_gentest.py to include gemm_strided_batched for these to be calculated correctly. A quick search didn't find that function included there.

TorreZuk: > I think you'll have to change rocblas_gentest.py to include gemm_strided_batched for these to be calculated correctly. A quick search didn't find that function included there.

Okay now I recall we didn't want to slow down gemms with stride_scale system, probably should abort when not processed.  Tests were running so will debug what value got used....

daineAMD (approved):

PR #1590, opened 2025-04-29
Fix compilation error due to missing rocTX library

rkamd:
Fix compilation error due to missing rocTX library

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -82,21 +82,50 @@ function( rocblas_library_settings lib_target_ )
     target_link_libraries( ${lib_target_} PRIVATE hip::device )
   else()
     target_link_libraries( ${lib_target_} PRIVATE hip::device -lstdc++fs --rtlib=compiler-rt --unwindlib=libgcc)
-    if (BUILD_SHARED_LIBS)
-        set(ROCTX_PATH "" CACHE STRING "Path to the roctx library (directory containing libroctx64.so)")
-        find_library(ROCTX_LIBRARY
+
+    if(BUILD_SHARED_LIBS)
+      set(ROCTX_PATH "" CACHE STRING "Path to the roctx library (directory containing libroctx64.so)")
+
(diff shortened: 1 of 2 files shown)


afzpatel:
Could we also add the include directory for rocTracer? I'm encountering an error in Spack while attempting to build rocblas for ROCm 6.4.0: `fatal error: 'roctracer/roctx.h' file not found`. In Spack, all of the ROCm packages are installed in separate paths, rather than in a single /opt/rocm directory. The issue can be resolved with two simple lines of code:
```
find_path(ROCTRACER_INCLUDE_DIR "roctracer/roctx.h")  
target_include_directories(${lib_target_} PRIVATE ${ROCTRACER_INCLUDE_DIR}) 
```

TorreZuk:
I am fine with the ifndef DISABLE form but are you adding @afzpatel 's request?   Best to treat it like a package.

rkamd:
> I am fine with the ifndef DISABLE form but are you adding @afzpatel 's request? Best to treat it like a package.

yes, the latest commit should have that change

library/CMakeLists.txt
```diff
@@ -82,8 +82,52 @@ function( rocblas_library_settings lib_target_ )
     target_link_libraries( ${lib_target_} PRIVATE hip::device )
   else()
     target_link_libraries( ${lib_target_} PRIVATE hip::device -lstdc++fs --rtlib=compiler-rt --unwindlib=libgcc)
-    if (BUILD_SHARED_LIBS)
-      target_link_libraries(${lib_target_} PRIVATE -lroctx64)
+
+    if(BUILD_SHARED_LIBS)
```
TorreZuk, line 86: Looks fine, in future maybe WIN32 and static can just use the DISABLE_ROCTX define.

TorreZuk (approved):
LGTM if all pipelines succeed.

afzpatel (approved):
LGTM, Thanks for adding the requested changes.

PR #1595, opened 2025-05-01
Adding new Arch to default list

mahmoodw:


TorreZuk (approved):
LGTM if tests pass but then add extended label.  May need manual trigger but can ask Eiden.

TorreZuk (commented):
Should remove the convertScalarForHipblasLT functions as well

amcamd (approved):

PR #1597, opened 2025-05-06
[swdev-523501] Add almalinux for the gfortran package dependency

srekolam:
resolves #___

Summary of proposed changes:
-
- Add Almalinux to the os-version for the gfortran package dependency.
-


TorreZuk (approved):
Looks good

PR #1584, opened 2025-04-21
Stream Order Allocation as default

rkamd:
resolves # [LWPMLSE-374](https://ontrack-internal.amd.com/browse/LWPMLSE-374)

Use hipMallocAsync() and hipFreeAsync() for workspace allocation
Remove user_managed mode in device memory management of rocBLAS
Clean up client test code
Remove deprecated funtions

PR #1598, opened 2025-05-06
Revert "Stream Order Allocation as default (#1584)"

rkamd:
This reverts commit 9160b10882448ce3d32ac8ecefce6f72775124c6.


TorreZuk (approved):
This is good, we will need to bring it back next week.  The hipBLAS fix PR I will make today.

amcamd (approved):

PR #1599, opened 2025-05-07
Fail if BLIS is not found but `LINK_BLIS` is set

marbre:
Fail if `LINK_BLIS` is set (defaults to `ON`) but the library is not found. Instead of issuing only a warning, return with a fatal error as the configuration phase otherwise passes but building rocBLAS fails when trying to link.


amcamd (approved):

TorreZuk:
Windows failure not caused by this change

TorreZuk (approved):

PR #1600, opened 2025-05-07
Find BLAS with `find_package()`

marbre:
Switches to `find_package()` instead of assuming that the reference BLAS is available without any check. This also allows to use OpenBLAS.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -185,7 +185,7 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     set( BLIS_CPP ../common/blis_interface.cpp )
   endif()
 
-  message(STATUS "Linking BLAS LIB: ${BLAS_LIBRARY}")
+  message(STATUS "Linking reference BLAS LIB: ${BLAS_LIBRARY}")
 
   if ( WARN_NOT_ILP64_PREFERRED )
     message( WARNING "Using ${BLAS_LIBRARY} as reference library, 64-bit tests may fail. Test suite should be run with --gtest_filter=-*stress*")
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -157,7 +157,8 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
         endif()
       endif()
     else()
-      set( BLAS_LIBRARY "blas" )
+      find_package( BLAS REQUIRED )
+      set( BLAS_LIBRARY "${BLAS_LIBRARIES}" )
     endif()
   else() # WIN32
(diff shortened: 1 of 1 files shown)


marbre:
We could keep `reference`in the log but from my understanding that wording refers to the Netlib implementation and with `find_package()`other BLAS providers can used as well.

TorreZuk:
> We could keep `reference`in the log but from my understanding that wording refers to the Netlib implementation and with `find_package()`other BLAS providers can used as well.

This simply means it is a CPU reference it does not imply specific implementation

marbre:
> > We could keep `reference`in the log but from my understanding that wording refers to the Netlib implementation and with `find_package()`other BLAS providers can used as well.
> 
> This simply means it is a CPU reference it does not imply specific implementation

Ah okay, I read it differently. Thanks for the explanation.

cgmb:
For reference, [0001-use-generic-blas-for-reference.patch](https://salsa.debian.org/rocm-team/rocblas/-/blob/debian/5.5.1+dfsg-7/debian/patches/0001-use-generic-blas-for-reference.patch?ref_type=tags) is used by Debian and Ubuntu use to allow rocblas-test to use any blas library chosen through [update-alternatives](https://wiki.debian.org/DebianScience/LinearAlgebraLibraries). I'm not sure if it will be possible to do that on the latest versions of rocBLAS given the need for 64-bit BLAS interfaces, but I mention it just because it's related and in-use.

cgmb:
Since rocBLAS likes using multi-threaded AOCL BLIS, I added support for AOCL to [FindBLAS.cmake](https://cmake.org/cmake/help/latest/module/FindBLAS.html). It landed in CMake 3.27, and can be selected by building with `-DBLA_VENDOR=AOCL_mt`.

marbre:
> Since rocBLAS likes using multi-threaded AOCL BLIS, I added support for AOCL to [FindBLAS.cmake](https://cmake.org/cmake/help/latest/module/FindBLAS.html). It landed in CMake 3.27, and can be selected by building with `-DBLA_VENDOR=AOCL_mt`.

That's good to know, thanks! I think it should be still okay to merge the patch as is to enable other BLAS providers but in that case the code to find BLIS could be further simplified to make use of `find_package()` for that case as well.

TorreZuk (approved):
Yes good to know @cgmb when we drop earlier cmake support might remove the specific precidence rules, but when "all vendors are considered" what order does it look?   This PR should be fine to require a package on the last fallback vs. just linking.

TorreZuk:
@marbre are you ready to merge now?

marbre:
> @marbre are you ready to merge now?

Oh I don't think I have permissions to merge @TorreZuk.

TorreZuk:
@marbre I can merge for you when you are fine for it to go in, just say it is ready for merge.

marbre:
> @marbre I can merge for you when you are fine for it to go in, just say it is ready for merge.

Yes, this is fine to merge as is for now. Please merge on my behalf.

PR #1591, opened 2025-04-29
Compatibility fixes for building on Windows.

ScottTodd:
resolves https://github.com/ROCm/TheRock/issues/482

Summary of proposed changes:
- Fix casing of `OpenBLAS` package.
- Fix ordering of `#include <windows.h>`.

PR #1601, opened 2025-05-07
Fix casing on OpenBLAS package for Windows compatibility

marbre:
rocBLAS configure is failing to find OPENBLAS on Windows with -DBUILD_TESTING=ON, which is fixed with this patch. See https://github.com/ROCm/TheRock/issues/482 for further details.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -172,7 +172,7 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
                     NO_DEFAULT_PATH
                   )
       if (NOT BLAS_LIBRARY)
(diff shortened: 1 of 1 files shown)


TorreZuk:
@marbre aren't you working with @ScottTodd ?   Did you discuss why they didn't request a merge on https://github.com/ROCm/rocBLAS/pull/1591  as this overlaps with his changes.  Are both required or only the case change?

marbre:
> @marbre aren't you working with @ScottTodd ? Did you discuss why they didn't request a merge on #1591 as this overlaps with his changes. Are both required or only the case change?

Both are required but I missed that Scott already sent a PR. Closing this in favor of #1591. Sorry for the noise.

PR #1602, opened 2025-05-07
Fix CMAKE_INSTALL_PREFIX init behavior on Win32.

marbre:
Per the guidance at https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html, CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT can be "used by project code to change the default without overriding a user-provided value".

This allows users and integration projects like https://github.com/ROCm/TheRock to set an install path of their choosing while still defaulting to this project's preferred value if the value is not already set. For more context, see https://github.com/ROCm/TheRock/pull/343.


TorreZuk (approved):
Windows test failures are unrelated.  Looks fine.

PR #1603, opened 2025-05-07
Update 'packaging' in pip venv to stabilize setuptools.

lucbruni-amd:
When using setuptools>=71.0.0 and packaging<22.0, setuptools may fail with `TypeError: canonicalize_version() got an unexpected keyword argument 'strip_trailing_zero'`.

### DIFF
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -59,6 +59,15 @@ function(virtualenv_install)
         endif()
     endif()
 
+    message("${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install --upgrade packaging")
+    execute_process(
+        RESULT_VARIABLE rc
+        COMMAND ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install --upgrade packaging
+    )
+    if(rc)
+        message(FATAL_ERROR ${rc})
+    endif()
+
     message("${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install ${ARGN}")
     execute_process(
       RESULT_VARIABLE rc
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -57,15 +57,15 @@ function(virtualenv_install)
         if(rc)
             message(FATAL_ERROR ${rc})
         endif()
-    endif()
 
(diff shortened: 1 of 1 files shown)


cmake/virtualenv.cmake
```diff
@@ -59,6 +59,15 @@ function(virtualenv_install)
...
 
+    message("${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install --upgrade packaging")
+    execute_process(
+        RESULT_VARIABLE rc
+        COMMAND ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install --upgrade packaging
+    )
+    if(rc)
+        message(FATAL_ERROR ${rc})
+    endif()
+
```
TorreZuk: Could this be guarded by TENSILE_VENV_UPGRADE_PIP as well?  It is closely related.  Are you defaulting to pip upgrade for this baseline python install as well?

davidd-amd:
@lucbruni-amd the 942 tests timed out and we need to update to develop.

cmake/virtualenv.cmake
TorreZuk: This still hasn't been addressed.

TorreZuk (changes requested):
Needs fixes to be merged

cmake/virtualenv.cmake
lucbruni-amd: Thanks, acknowledged - code owners, please don't merge until this is addressed.

lucbruni-amd: @ScottTodd - tagging here.

ScottTodd: We do not currently set `TENSILE_VENV_UPGRADE_PIP` in TheRock but we could. I'd be okay with moving this new code up 2 lines into that block. @lucbruni-amd can you make that change here? Then @marbre when we pull in changes to rocBLAS we can add that additional CMake argument here: https://github.com/ROCm/TheRock/blob/195a3e395fdf56337450f42360249414c0cde0ae/math-libs/BLAS/CMakeLists.txt#L167-L183.

ScottTodd (approved):

TorreZuk (approved):
Yes that looks fine, as by default needs to maintain a hermetic build so this is good now.  Can you update the documentation in the changelog and rmake.py to mention this double upgrade in this PR

ammallya:
Imported to ROCm/rocm-libraries

marbre:
> Yes that looks fine, as by default needs to maintain a hermetic build so this is good now. Can you update the documentation in the changelog and rmake.py to mention this double upgrade in this PR

@lucbruni-amd can you address those and add it to the PR in the monorepo?

PR #1608, opened 2025-05-12
Stream Order Allocation as default 

rkamd:
resolves #[LWPMLSE-374](https://ontrack-internal.amd.com/browse/LWPMLSE-374)

Use hipMallocAsync() and hipFreeAsync() for workspace allocation
Remove user_managed mode in device memory management of rocBLAS
Clean up client test code
Remove deprecated funtions

PR #1621, opened 2025-05-21
Revert "Stream Order Allocation as default  (#1608)"

TorreZuk:

* testing only for now


TorreZuk:
@rkamd please review as it looks like it passed the get_solutions tests already.  Need to analyze

TorreZuk:
Further analysis to be done by @rkamd on how stream order allocation might affect hipblaslt but not proceeding with this PR.

PR #1637, opened 2025-06-07
Update tensile tag

davidd-amd:
Updates tag to tip of develop to include tensile msgpack fix.


TorreZuk (approved):
LGTM if all relevant tests pass

PR #1642, opened 2025-06-11
Revert "Stream Order Allocation as default  (#1608)"

rkamd:

resolves # SWDEV-537363

Revert the stream order allocation as default.

### DIFF
--- a/library/include/internal/rocblas-functions.h
+++ b/library/include/internal/rocblas-functions.h
@@ -25024,7 +25024,25 @@ ROCBLAS_EXPORT rocblas_status rocblas_get_device_memory_size(rocblas_handle hand
 
 /*! \brief
     \details
-    Allows user to set the device memory for the handle to use as a workspace (user-owned scheme).
+    Changes the size of allocated device memory at runtime.
+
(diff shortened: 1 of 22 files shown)


library/include/internal/rocblas-functions.h
```diff
@@ -25047,6 +25065,16 @@ ROCBLAS_EXPORT rocblas_status rocblas_set_workspace(rocblas_handle handle, void*
...
 
+/*! \brief
+    \details
+    Returns true when device memory in handle is managed by the user
+    @param[in]
+    handle          rocblas handle
+ ******************************************************************************/
+ROCBLAS_DEPRECATED_MSG("rocblas_is_user_managing_device_memory will be removed in a future release "
+                       "and supported modes will be rocblas_managed and user_owned")
+ROCBLAS_EXPORT bool rocblas_is_user_managing_device_memory(rocblas_handle handle);
```
TorreZuk, line 25076: Fine for clean revert but need to see what might be added back quick, like is this API required to avoid the issue?

TorreZuk (approved):
I just reproduced perf drop with the original commit on 942, so more supporting data.

PR #1644, opened 2025-06-11
Revert "Stream Order Allocation as default  (#1608)" (#1642)

rkamd:

(cherry picked from commit 1d3ef0eb0ec21d4f6657697cc3637132f41e755d)


TorreZuk (approved):
This compile issue is missing artifacts but passed on develop I think so fine when other tests pass IMO

PR #1645, opened 2025-06-11
SWDEV-537656 - Add include for assert in example_user_driven_tuning

iassiour:
resolves https://ontrack-internal.amd.com/browse/SWDEV-537656

Summary of proposed changes:
As part of HIP runtime's 7.0 breaking change, compile time constants _AMDGCN_WAVEFRONT_SIZE & warpSize has been removed via :
https://github.com/AMD-ROCm-Internal/hip/pull/82 
https://github.com/AMD-ROCm-Internal/clr/pull/330 

This is causing  "warpSize cannot be used in a constant expression error".


TorreZuk:
@iassiour I did some original changes related to this when only constexpr was proposed as being dropped.  Not sure what happened in your rebase, but I can help if required.  If the even the symbol warpSize is now being removed this makes sense.

iassiour:
@TorreZuk after rebase I realize that the "warpSize cant be used as a constexpr" issue is now resolved in the develop branch. I think that leaves only the addition of an extra include for assert from my original PR.

amcamd (approved):

TorreZuk:
I will do the cherry pick to release-staging/rocm-rel-7.0 @iassiour

PR #1648, opened 2025-06-12
Zero imag portion of diagonal in syrk/herk

daineAMD:
Needs to be cherry-picked back to releases. Running tests and need to ensure that the new tests fail for the broken kernel as well.

### DIFF
--- a/library/src/blas3/herk_syrk_device.hpp
+++ b/library/src/blas3/herk_syrk_device.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (C) 2020-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2020-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


library/src/blas3/herk_syrk_device.hpp
```diff
@@ -1379,6 +1379,11 @@ rocblas_copy_triangular_excluding_diagonal_kernel(rocblas_int    n,
             }
         }
 
+        // When copying back to C, we need to zero-out diagonal
+        if constexpr(!copy_from_C_to_W_C && rocblas_is_complex<T>)
+            if(row == col)
+                C[row + row * int64_t(ldc)] = std::real(C[row + row * int64_t(ldc)]);
```
TorreZuk: comment questions.     zero-out diagonal imaginary ?  
kernel name is now a misnomer "rocblas_copy_triangular_excluding_diagonal_kernel"

TorreZuk (approved):
Existing tests failed with unit check without your fix, correct?
Should we make debt ticket to use near for some hpl tests?

TorreZuk:
Forgot to say great work on quick fix!

daineAMD:
This is going to have a few more changes tomorrow. We aren't testing this correctly and I think another kernel may be broken so turning off CI for now.
Edit: I was wrong, and decided to do it this evening.

- I reverted the changes to the syrk/herk testing. Thought that the near-check code must exist for a reason so no use changing that for a cherry-pick. Can change it in another PR into develop
- Changed herk testing initialization to general matrix rather than hermitian matrix. By initializing it to HERM, we can be hiding accesses of the opposite triangle, plus it initializes the imag-diagonal to zero. Further, I feel like we should remove initializing symmetric/triangular/hermitian matrices differently where possible because of this reason. Can discuss tomorrow.
- With this testing change, the code before this fails as expected, and now passes with the new code. Still want to wait for CI as didn't run stress or on other archs.

library/src/blas3/herk_syrk_device.hpp
daineAMD: Do you like rocblas_copy_triangular_herk_syrk_kernel?

daineAMD:
For future work regarding re-evaluating testing: https://ontrack-internal.amd.com/browse/LWPMLSE-860

PR #1658, opened 2025-06-25
fix warnings

TorreZuk:
* fix explicit specialization cannot have a storage class
* clang still requires repeated static to avoid duplicate symbols but inline should avoid these too in a more C++17 manner


daineAMD (approved):
no space left for static-analysis, probably would be good to get that pipeline to run before merge.

ScottTodd:
I suspect this may have broken Windows PyTorch builds when compiling using clang-cl. More details at https://github.com/ROCm/TheRock/issues/1123.

(and yes, we're working towards getting that entire build process running on all PRs continuously, so it doesn't take 1 month to detect breaks...)

PR #1660, opened 2025-06-26
Add hipblaslt_include_dirs to clients include path

daineAMD:


TorreZuk (approved):
Thanks, yes just version info.  I thought we should really also print their runtime string, as if the build version doesn't match runtime we really should know that too.
Doesn't have to be added here, but that version print was 7.1 addition.

PR #1670, opened 2025-08-11
disable dependabot.yml

samjwu:
disable dependabot.yml

Summary of proposed changes:
- stops dependabot prs for docs
- dependabot.yml is separate from the settings/advanced security config, but still has to be valid. example: https://github.com/ROCm/rocBLAS/pull/1669 was still opened despite having invalid dependabot.yml

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,1 +1,12 @@
-
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/docs/sphinx" # Location of package manifests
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
LGTM

TorreZuk (commented):
Not sure if develop should also be target branch?

TorreZuk:
> Not sure if develop should also be target branch?

Or only develop if this is locked.

samjwu:
my understanding is the bot config for only the default branch is used (ie: develop_deprecated in this case)

